PR #59, opened 2019-09-25
Initial implementation of a 'client' subsystem.

Whathecode:
The central access point into the client API is `ClientManager`. Through this you can currently add studies using `addStudy`, which will initialize a `StudyRuntime` which will become the central object through which study lifetime is controlled (accessible through `studies` in `StudyRuntime`). A 'snapshot' system (for persistence) is available on `ClientManager`.

For now, a study will be instantly deployed when calling `addStudy`, if possible. In case the study cannot be deployed yet since other devices need to be registered first, `isDeployed` will be false, and you can call `tryDeployment` once the other device has been registered.

Once `isDeployed` is true, `deploymentInformation` in `StudyRuntime` will be set, which is a `MasterDeviceDeployment` containing all the information needed to run the study. Normally, you would not have to consume this object yourself but this would be used to further initialize the client domain model. Exceptions would be thrown in case something fails upon deployment. However, for now this is made public to work towards an initial integration of the client and server domain model.


carp.client.core/src/commonTest/kotlin/dk/cachet/carp/client/application/ClientManagerTest.kt
```diff
@@ -0,0 +1,78 @@
...
+ * Tests for [ClientManager].
+ */
+class ClientManagerTest
+{
+    @Test
+    fun add_study_succeeds()
+    {
+        // Create deployment and client manager.
+        val ( deploymentManager, deploymentStatus) = createStudyDeployment( createSmartphoneStudy() )
+        val clientManager = SmartphoneManager( smartphone.createRegistration(), deploymentManager )
```
patrickusiewicz: Wouldn't it be better to get the `smartphone` from the `deploymentStatus`?
Then who ever are going to consume the client subsystem, will now how to initialize the `SmartphoneManager` based on information in the `deploymentStatus`.

Whathecode: In a normal use case you would retrieve it through `StudyDeploymentStatus`. However, for the unit tests here I use a hardcoded object to reduce the LOC of each test (this is quite repetitive).

Whathecode: Oh, I am mistaken. I simply used `smartphone.createRegistration` here to simply creating the matching `DeviceRegistration`. I think what would be used in practice is `DefaultDeviceRegistrationBuilder().build()`, for which perhaps a cleaner API needs to be provided.

patrickusiewicz: I do get you want to reduce the LOC for each test.
But couldn't you create a shared function to retreive the smartphone from the `deploymentStatus`?

PR #68, opened 2019-10-02
Convert DeploymentService functions to suspending functions

patrickusiewicz:
The `DeploymentService` interface defines blocking methods. This is not a suitable API since implementations might rely on slow operations, such as REST calls.

A straightforward solution would be to rely on Kotlin's built-in language support for asynchronous operations: [coroutines](https://kotlinlang.org/docs/reference/coroutines-overview.html). There is multiplatform support for this. However, a potential downside is that it might be unclear how to consume such APIs natively. More research is needed to determine exactly how this would work.

Furthermore, we have had a discussion on the following two related points:

- Should the same interface be used to _implement_ the service as well as to consume it?
- Is the need for an asynchronous implementation of this interface an implementation detail or not? I.e., is this an infrastructural concern or not?

The current pull request includes the required changes to be able to implement `DeploymentService` using coroutines. Note: every function that implements or consumes the suspending functions in `DeploymentService` also had to be converted into suspending functions. Helper functions are created in the `carp.test` module to consume the new suspending functions by blocking.

### DIFF
--- a/carp.client.core/src/commonTest/kotlin/dk/cachet/carp/client/domain/StudyRuntimeTest.kt
+++ b/carp.client.core/src/commonTest/kotlin/dk/cachet/carp/client/domain/StudyRuntimeTest.kt
@@ -1,6 +1,7 @@
 package dk.cachet.carp.client.domain
 
 import dk.cachet.carp.client.infrastructure.*
+import dk.cachet.carp.test.runBlockingTest
 import kotlin.test.*
 
 
@@ -10,8 +11,7 @@ import kotlin.test.*
 class StudyRuntimeTest
 {
     @Test
-    fun initialize_matches_requested_runtime()
-    {
+    fun initialize_matches_requested_runtime() = runBlockingTest {
         // Create a deployment manager which contains a 'smartphone study'.
         val ( deploymentManager, deploymentStatus) = createStudyDeployment( createSmartphoneStudy() )
 
@@ -23,8 +23,7 @@ class StudyRuntimeTest
     }
 
     @Test
-    fun initialize_deploys_when_possible()
-    {
+    fun initialize_deploys_when_possible() = runBlockingTest {
         // Create a deployment manager which contains a 'smartphone study'.
         val ( deploymentManager, deploymentStatus) = createStudyDeployment( createSmartphoneStudy() )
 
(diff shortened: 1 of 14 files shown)


patrickusiewicz:
The `runBlockingTest` functions are based on workarounds from [kotlin issue](https://youtrack.jetbrains.com/issue/KT-22228).

carp.test/src/commonMain/kotlin/dk/cachet/carp/test/runBlockingTest.kt
```diff
@@ -0,0 +1,3 @@
+package dk.cachet.carp.test
+
+expect fun runBlockingTest(block: suspend () -> Unit)
```
Whathecode: Very cool solution. Doesn't seem to cause too many problems this way. Minor note: carp Kotlin code takes spaces in parenthesis.

carp.client.core/src/commonTest/kotlin/dk/cachet/carp/client/domain/StudyRuntimeTest.kt
```diff
@@ -10,8 +11,7 @@ import kotlin.test.*
 class StudyRuntimeTest
 {
     @Test
-    fun initialize_matches_requested_runtime()
-    {
+    fun initialize_matches_requested_runtime() = runBlockingTest {
         // Create a deployment manager which contains a 'smartphone study'.
         val ( deploymentManager, deploymentStatus) = createStudyDeployment( createSmartphoneStudy() )
```
Whathecode: While you are at at, maybe also rename these `deploymentManager` to `deploymentService`. I seemingly forgot to rename them in `StudyRuntimeTest`, or this was more merge-foo.

patrickusiewicz: Dont you have permissions to push changes to this PR? If you do, feel free to make does changes.

carp.test/src/commonMain/kotlin/dk/cachet/carp/test/runBlockingTest.kt
patrickusiewicz: Missed the spaces in parenthesis style. Do you have a style guide written somewhere? I dont see one in your repo. Maybe an automatically apply style guide to PRs could be added?

Whathecode: It's in my head. =)

PR #101, opened 2020-02-19
Test extending types serialization

Whathecode:
Serialization of all extending types in `carp.protocols` is tested.

There were no tests for this previously, which is why we did not notice `AltBeaconDeviceRegistration` could not be serialized on the JS runtime.

A temporary workaround for what is very likely a `kotlinx.serialization` bug is included in this PR for `AltBeaconDeviceRegistration`, making sure that the new tests pass.

`ConcreteTypesSerializationTest` can be used to test extending types in extending libraries.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -77,6 +77,7 @@ configure( subprojects - detektModule ) {
 
     // Specify platforms and test frameworks to use.
     apply plugin: 'kotlin-multiplatform'
+    apply plugin: 'kotlinx-serialization'
     kotlin {
         jvm {
             compilations.main.kotlinOptions.jvmTarget = versions.jvmTarget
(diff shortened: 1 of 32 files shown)


build.gradle
```diff
@@ -67,6 +67,7 @@ if (publishPropertiesFile.exists())
 }
 
 // Configure all subprojects as testable, publishable, Kotlin multiplatform projects.
+// A `kotlinx.serialization` dependency is added to serialize domain models.
 configure( subprojects - detektModule ) {
     version = globalVersion
```
patrickusiewicz: Because you have added kotlinx.serialization dependencies, you should apply kotlinx.serialization plugin in this scope.

patrickusiewicz:
Shouldn't we link to your bug report to kotlinx.serialization?

build.gradle
Whathecode: Well-spotted!

PR #107, opened 2020-03-06
Property level updates on repos

ltj:
When implementing the repositories with complex data-stores (PostgreSQL, Mongo etc) I find that I often cannot leverage the power of these system due to the all-or-nothing approach that comes with the snapshot approach.

By targeting the smallest data scope possible in the repository methods you trade more methods and a little added complexity for great flexibility in how you design for the chosen datastore, especially since these are tuned for partial updates, be it document or rdbms based. 

This is not a fully thought out suggestion but rather something we can experiment a bit with. As a start I suggest we see how it plays out in the study domain.


ltj:
...aaand fuck this

PR #108, opened 2020-03-10
Add base class for aggregate roots

Whathecode:
To simplify implementing repository interfaces using a relational database, more detailed information about changes to aggregate roots need to be tracked. This PR includes a base class which supports tracking events within aggregate roots, which can be consumed once these changes need to be persisted into a repository.

Alternatively, in case a document store is needed, this base `AggregateRoot` class also provides generic access to convert aggregate roots into snapshots and vice versa (`AggregateRoot.getSnapshot` and `Snapshot.toObject`). The idea is this might potentially be used to have a singular implementation to store/update any aggregate root in a document store.

Lastly, snapshots and aggregate root events should really be `Immutable`. To enforce a correct immutable implementation, the `Immutable` base class is applied to them.

Right now, the aggregate root base class has only been applied to `Study`. In case this approach seems workable, I can apply them to all other aggregate roots in the project (e.g., `StudyDeployment`, `StudyProtocol`, ...).

### DIFF
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/Study.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/Study.kt
@@ -1,7 +1,9 @@
 package dk.cachet.carp.studies.domain
 
 import dk.cachet.carp.common.DateTime
+import dk.cachet.carp.common.Immutable
 import dk.cachet.carp.common.UUID
+import dk.cachet.carp.common.ddd.AggregateRoot
 import dk.cachet.carp.deployment.domain.users.Participation
 import dk.cachet.carp.deployment.domain.users.StudyInvitation
 import dk.cachet.carp.protocols.domain.InvalidConfigurationError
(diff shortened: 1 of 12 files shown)


carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/Study.kt
```diff
@@ -23,14 +25,23 @@ class Study(
...
+    name: String,
     /**
      * A description of the study, shared with participants once they are invited to the study.
      */
     val invitation: StudyInvitation = StudyInvitation.empty(),
     val id: UUID = UUID.randomUUID()
-)
+) : AggregateRoot<Study, StudySnapshot, Study.Event>()
 {
+    sealed class Event : Immutable()
```
ltj: @Whathecode Wonder if there should be a `StudyAdded` event as well that simply references itself or is empty. This way we can leave out the repo `add` method as a separate thing and the initial creation of the entity becomes part of the event log (when present in the infrastructure)

Whathecode: Isn't that implied? If we remove `add` and only have `store` (or our old `update`), if you then pass an object which does not exist, that you have to add it? But, I'm guessing you are already looking towards full fledged event sourcing, perhaps, in which case we'd need an `added` event either way?

Whathecode: My main question would be, who decides when that event should be added? The aggregate root takes care of all of its internals, but not itself. Whether it is initialized through the constructor, or snapshot, there should be no real difference. Then it is the application service that knows when an AR is new. But, opening up raising events breaks encapsulation.I think it is cleaner to have creation separated from

ltj: hmm...but it's more than the semantics of non-existence. It's also the semantics of dispatching an empty event queue on `store`. We can argue this means it has to be created and error if it exists already. Ok. Let's leave it implicit for now, and then add an explicit event if we see reason to do so.

PR #112, opened 2020-03-15
Add information on available actions to `StudyStatus`

Whathecode:


carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/StudyStatus.kt
```diff
@@ -9,22 +9,53 @@ import kotlinx.serialization.Serializable
...
+     */
+    @Serializable
+    data class Live(
+        override val studyId: UUID,
+        override val name: String,
+        override val creationDate: DateTime,
+        override val canDeployToParticipants: Boolean,
+        override val canSetStudyProtocol: Boolean
+    ) : StudyStatus()
+}
```
ltj: I think this was actually more or less what I wanted :) I'm still no fan of top-level polymorphic serialization as i think it imposes a very stack specific interpretation of messages passed between systems.  But let's try this for now

PR #113, opened 2020-03-19
Additional `StudyService` endpoints to get and set study details

Whathecode:
Some endpoints are missing for Luca to manipulate studies. This PR is currently incomplete, but since I generally work with cohesive commits, you can already review the current additions and comment as I go along.

- `getStudyDetails`: returns `StudyDetails`, including the study protocol, which Luca needed in order to show a summary of the protocol after it was set. This also includes the study description.
- `updateInternalDescription`: update the study name and description.
- `setStudyInvitation`: to change the invitation which is sent out to participants. A 'description' is also added to `StudyInvitation`. This can only be set as long as the study has not gone live, which is now also reflected as a boolean in `StudyStatus`.

In addition, I also updated dependency versions, including an update to Kotlin 1.3.71.


build.gradle
```diff
@@ -13,22 +13,22 @@ buildscript {
 
         versions = [
             // Kotlin multiplatform versions.
-            kotlin:'1.3.70',
+            kotlin:'1.3.71',
```
ltj: Good call. That should ease transition into 1.4

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyService.kt
```diff
@@ -13,17 +14,37 @@ import dk.cachet.carp.studies.domain.users.Participant
 /**
  * Application service which allows creating and managing studies.
  */
+@Suppress( "TooManyFunctions" ) // TODO: Perhaps split up participation management from main interface.
```
ltj: We should not do this to satisfy detekt defaults, but rather because it could make it a bit clearer what belongs where and that you are operating on different entities altogether

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyServiceHost.kt
```diff
@@ -31,20 +33,63 @@ class StudyServiceHost(
      * Create a new study for the specified [owner].
      *
      * @param name A descriptive name for the study, assigned by, and only visible to, the [owner].
+     * @param description An optional description of the study, assigned by, and only visible to, the [owner].
      * @param invitation
      *  An optional description of the study, shared with participants once they are invited.
      *  In case no description is specified, [name] is used as the name in [invitation].
      */
-    override suspend fun createStudy( owner: StudyOwner, name: String, invitation: StudyInvitation? ): StudyStatus
+    override suspend fun createStudy( owner: StudyOwner, name: String, description: String, invitation: StudyInvitation? ): StudyStatus
```
ltj: Please keep lines shorter, otherwise it makes reviews in the GH UI cumbersome 👌

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyService.kt
```diff
@@ -13,17 +14,37 @@ import dk.cachet.carp.studies.domain.users.Participant
...
      * Create a new study for the specified [owner].
      *
      * @param name A descriptive name for the study, assigned by, and only visible to, the [owner].
+     * @param description An optional description of the study, assigned by, and only visible to, the [owner].
      * @param invitation
      *  An optional description of the study, shared with participants once they are invited.
      *  In case no description is specified, [name] is used as the name in [invitation].
      */
-    suspend fun createStudy( owner: StudyOwner, name: String, invitation: StudyInvitation? = null ): StudyStatus
+    suspend fun createStudy( owner: StudyOwner, name: String, description: String = "", invitation: StudyInvitation? = null ): StudyStatus
```
ltj: break params on multiple lines

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/StudyServiceRequest.kt
```diff
@@ -21,9 +22,19 @@ import kotlinx.serialization.Serializable
...
+
+    @Serializable
+    data class SetInternalDescription( val studyId: UUID, val name: String, val description: String ) :
+        StudyServiceRequest(),
+        ServiceInvoker<StudyService, StudyStatus> by createServiceInvoker( StudyService::setInternalDescription, studyId, name, description )
+
+    @Serializable
+    data class GetStudyDetails( val studyId: UUID ) :
+        StudyServiceRequest(),
+        ServiceInvoker<StudyService, StudyDetails> by createServiceInvoker( StudyService::getStudyDetails, studyId )
```
ltj: lot's of long lines :)

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyService.kt
Whathecode: We consider `Participant` a separate aggregate root (AR), which is only referred to from `Study` by ID through the list of participations. So, logically, it would make sense to split of management of distinct separate ARs into separate interfaces. I still think it is still useful to keep one 'study' endpoint, though, so think I can apply these changes without them having to impact you. It would simply involve interface delegation.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/StudyServiceRequest.kt
Whathecode: These I won't break apart since it's boiler plate code either way.

PR #119, opened 2020-04-08
Finish basic deployment status

Whathecode:
**Done:**
- All devices require registration on the back-end before a deployment can be considered 'ready' (https://github.com/cph-cachet/carp.core-kotlin/commit/eb4aa8862be1cb24c8272a6dce7c8004b96ee782). The previous `requiresRegistration` is now specific to individual devices, represented as `remainingDevicesToRegisterBeforeDeployment` in `DeviceDeploymentStatus` (https://github.com/cph-cachet/carp.core-kotlin/commit/b067518d733e0886ea86865541b6d197aaf8aa9a).
- Call `deploymentSuccesful` from `StudyRuntime` (https://github.com/cph-cachet/carp.core-kotlin/commit/6054d89c78a3d2730242030f251b186c358a23d6).
- A `StudyDeployment` can be stopped and has the following `StudyDeploymentStatus` states: `Invited`, `DeployingDevices`, `DeploymentReady`, and `Stopped` (https://github.com/cph-cachet/carp.core-kotlin/pull/119/commits/4bb823711f028a9b785f3c7eb6216b937428a84c).
- The study subsystem now has access to deployment statuses. This can be used to show an overview of all running study deployments (or participant groups, as they are now called in the study subsystem). Participant groups can later carry extra study-subsystem-specific information, but there is a one-to-one mapping to deployments. Added `StudyService.getParticipantGroupStatuses()` and `DeploymentService.getStudyDeploymentStatuses` which it redirects the call to (https://github.com/cph-cachet/carp.core-kotlin/pull/119/commits/4bb6e01408bc522e39ffe89ffd32ea4cf16384d4).
- Added endpoint to stop deployments to `DeploymentService` (https://github.com/cph-cachet/carp.core-kotlin/pull/119/commits/245d29d594407b03280c3fbfbfdda0ffdbbdbf2c) and to stop 'participant groups' as `StudyService.stopParticipantGroup` (https://github.com/cph-cachet/carp.core-kotlin/pull/119/commits/a1e10fe89254eb9368dd9a7430c15c23d89e8126).

**To discuss:**
- I added `DeploymentService.getStudyDeploymentStatuses( studyDeploymentIds )` to enable rendering the full overview of all deployments in the study manager. @ltj, I seem to remember that raised a concern for you in regards to authorization. Will implementing authorization for each individual `studyDeploymentId` be a problem here? Recall that the deployment subsystem is unaware of study ID.
- The study manager will later want to filter, e.g., show all running participant groups. Will it be the task of the study subsystem to cache stopped groups, or should such a filter be passed on to the deployment subsystem?

**Skipped for now**
- We talked about whether or not we should be able to see who stopped a study deployment. But, while trying to implement this I noticed this might depend on a better model of how different accounts participate in a deployment (participant or researcher?), and potentially even incorporating authorization in the deployment domain. This needs more analysis.
- I considered storing the time when a deployment was stopped, but then why not store when it started, when the first device was registered, when the deployment was ready, etc. I think we need to do some use case analysis to have a better understanding which timestamps should be stored and how they are accessed. If all this were needed, this would also advocate switching to event sourcing.
- I did not yet add the TypeScript declarations for devices in the protocol subsystem. This will hopefully be generated during compile time instead in Kotlin 1.4.

### DIFF
--- a/carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntime.kt
+++ b/carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntime.kt
@@ -29,6 +29,10 @@ class StudyRuntime private constructor(
          * Instantiate a [StudyRuntime] by registering the client device in the [deploymentService].
          * In case the device is immediately ready for deployment, also deploy.
          *
+         * @throws IllegalArgumentException when:
+         * - a deployment with [studyDeploymentId] does not exist
+         * - [deviceRoleName] is not present in the deployment or is already registered and a different [deviceRegistration] is specified than a previous request
+         * - [deviceRegistration] is invalid for the specified device or uses a device ID which has already been used as part of registration of a different device
          * @throws UnsupportedOperationException in case deployment failed since not all necessary plugins to execute the study are available.
          */
         internal suspend fun initialize(
@@ -52,7 +56,6 @@ class StudyRuntime private constructor(
         ): StudyRuntime
         {
             // Register the client device this study runs on for the given study deployment.
-            // TODO: What if registration succeeds, but deployment fails since not all required plugins are available? Registering again is not allowed right now.
             val deploymentStatus = deploymentService.registerDevice( studyDeploymentId, deviceRoleName, deviceRegistration )
 
             // Initialize runtime.
@@ -60,23 +63,16 @@ class StudyRuntime private constructor(
             val runtime = StudyRuntime( deploymentService, studyDeploymentId, clientDeviceStatus.device as AnyMasterDeviceDescriptor )
 
             // After registration, deployment information might immediately be available for this client device.
-            if ( clientDeviceStatus is DeviceDeploymentStatus.NotDeployed && clientDeviceStatus.isReadyForDeployment )
-            {
-                runtime.deploymentInformation = deploymentService.getDeviceDeploymentFor( studyDeploymentId, deviceRoleName )
-                runtime.deploy( runtime.deploymentInformation!! )
+            runtime.tryDeployment( clientDeviceStatus )
+
+            return runtime
+        }
+
+        internal fun fromSnapshot( snapshot: StudyRuntimeSnapshot, deploymentService: DeploymentService ): StudyRuntime =
+            StudyRuntime( deploymentService, snapshot.studyDeploymentId, snapshot.device ).apply {
+                isDeployed = snapshot.isDeployed
+                deploymentInformation = snapshot.deploymentInformation
             }
-
-            return runtime
-        }
-
-        internal fun fromSnapshot( snapshot: StudyRuntimeSnapshot, deploymentService: DeploymentService ): StudyRuntime
-        {
-            val runtime = StudyRuntime( deploymentService, snapshot.studyDeploymentId, snapshot.device )
-            runtime.isDeployed = snapshot.isDeployed
-            runtime.deploymentInformation = snapshot.deploymentInformation
-
-            return runtime
-        }
     }
 
 
@@ -87,7 +83,7 @@ class StudyRuntime private constructor(
         private set
 
     /**
-     * In case deployment succeeded ([isDeployed] is true), this contains all the information on the study the run.
+     * In case deployment succeeded ([isDeployed] is true), this contains all the information on the study to run.
      * TODO: This should be consumed within this domain model and not be public.
      *       Currently, it is in order to work towards a first MVP which includes server/client communication through the domain model.
      */
@@ -103,21 +99,30 @@ class StudyRuntime private constructor(
     suspend fun tryDeployment(): Boolean
     {
         val deploymentStatus = deploymentService.getStudyDeploymentStatus( studyDeploymentId )
-        val clientDeviceStatus = deploymentStatus.devicesStatus.first { it.device == device }
+        val clientDeviceStatus = deploymentStatus.getDeviceStatus( device )
 
-        if ( clientDeviceStatus is DeviceDeploymentStatus.NotDeployed && clientDeviceStatus.isReadyForDeployment )
-        {
-            deploymentInformation = deploymentService.getDeviceDeploymentFor( studyDeploymentId, device.roleName )
-            deploy( deploymentInformation!! )
-        }
-
-        return isDeployed
+        return tryDeployment( clientDeviceStatus )
     }
 
-    private fun deploy( deploymentInformation: MasterDeviceDeployment )
+    private suspend fun tryDeployment( deviceStatus: DeviceDeploymentStatus ): Boolean
     {
-        // TODO: Throw exception in case there are missing plugins to perform the operations (e.g., measurements) specified in the deployment.
+        // Early out in case state indicates the device is not yet ready to deploy.
+        if ( deviceStatus !is DeviceDeploymentStatus.NotDeployed || !deviceStatus.isReadyForDeployment ) return false
 
-        isDeployed = true
+        // TODO: Get deployment and throw exception in case there are missing plugins to perform the operations (e.g., measurements).
+        val deployment = deploymentService.getDeviceDeploymentFor( studyDeploymentId, device.roleName )
+        deploymentInformation = deployment
+
+        // Notify deployment service of successful deployment.
+        try
+        {
+            deploymentService.deploymentSuccessful( studyDeploymentId, device.roleName, deployment.getChecksum() )
+            isDeployed = true
+        }
+        // Handle race conditions with competing clients modifying device registrations, invalidating this deployment.
+        catch ( arg: IllegalArgumentException ) { }
+        catch ( arg: IllegalStateException ) { }
+
+        return isDeployed
     }
 }
--- a/carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntime.kt
+++ b/carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntime.kt
@@ -120,8 +120,8 @@ class StudyRuntime private constructor(
             isDeployed = true
         }
         // Handle race conditions with competing clients modifying device registrations, invalidating this deployment.
-        catch ( arg: IllegalArgumentException ) { }
-        catch ( arg: IllegalStateException ) { }
+        catch ( ignore: IllegalArgumentException ) { }
+        catch ( ignore: IllegalStateException ) { }
 
         return isDeployed
     }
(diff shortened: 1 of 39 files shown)


carp.deployment.core/src/commonMain/kotlin/dk/cachet/carp/deployment/domain/StudyDeploymentSnapshot.kt
```diff
@@ -20,6 +20,7 @@ data class StudyDeploymentSnapshot(
     val deviceRegistrationHistory: Map<String, List<@Serializable( DeviceRegistrationSerializer::class ) DeviceRegistration>>,
     val deployedDevices: Set<String>,
     val invalidatedDeployedDevices: Set<String>,
+    val hasStopped: Boolean,
```
ltj, line 23: `isStopped`

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyServiceHost.kt
```diff
@@ -216,18 +217,19 @@ class StudyServiceHost(
...
      * @throws IllegalStateException when the study is not yet ready for deployment.
      */
-    override suspend fun deployParticipantGroup( studyId: UUID, group: Set<AssignParticipantDevices> ): StudyStatus
+    override suspend fun deployParticipantGroup( studyId: UUID, group: Set<AssignParticipantDevices> ): ParticipantGroupStatus
     {
         require( group.isNotEmpty() ) { "No participants to deploy specified." }
 
         // Verify whether the study is ready for deployment.
         val study: Study? = repository.getById( studyId )
         require( study != null ) { "Study with the specified studyId is not found." }
```
patrickusiewicz, line 226: You can also use `requireNotNull`.

carp.deployment.core/src/commonMain/kotlin/dk/cachet/carp/deployment/infrastructure/DeploymentServiceRequest.kt
```diff
@@ -25,12 +28,17 @@ sealed class DeploymentServiceRequest
...
+        Invoker<StudyDeploymentStatus> by createServiceInvoker( Service::createStudyDeployment, protocol )
 
     @Serializable
     data class GetStudyDeploymentStatus( val studyDeploymentId: UUID ) :
         DeploymentServiceRequest(),
-        ServiceInvoker<DeploymentService, StudyDeploymentStatus> by createServiceInvoker( DeploymentService::getStudyDeploymentStatus, studyDeploymentId )
+        Invoker<StudyDeploymentStatus> by createServiceInvoker( Service::getStudyDeploymentStatus, studyDeploymentId )
+
+    @Serializable
+    data class GetStudyDeploymentStatuses( val studyDeploymentIds: Set<UUID> ) :
```
ltj, line 39: Rename as well to reflect this an action that returns a list

carp.deployment.core/src/commonMain/kotlin/dk/cachet/carp/deployment/application/DeploymentService.kt
```diff
@@ -30,22 +29,27 @@ interface DeploymentService
...
      * @throws IllegalArgumentException when a deployment with [studyDeploymentId] does not exist.
      */
     suspend fun getStudyDeploymentStatus( studyDeploymentId: UUID ): StudyDeploymentStatus
 
+    /**
+     * Get the statuses for a set of deployments with the specified [studyDeploymentIds].
+     *
+     * @throws IllegalArgumentException when [studyDeploymentIds] contains an ID for which no deployment exists.
+     */
+    suspend fun getStudyDeploymentStatuses( studyDeploymentIds: Set<UUID> ): List<StudyDeploymentStatus>
```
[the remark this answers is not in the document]
patrickusiewicz: I agree, it is difficult distinguish the differences between the function names.

carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntime.kt
```diff
@@ -29,6 +29,10 @@ class StudyRuntime private constructor(
          * Instantiate a [StudyRuntime] by registering the client device in the [deploymentService].
          * In case the device is immediately ready for deployment, also deploy.
          *
+         * @throws IllegalArgumentException when:
+         * - a deployment with [studyDeploymentId] does not exist
+         * - [deviceRoleName] is not present in the deployment or is already registered and a different [deviceRegistration] is specified than a previous request
+         * - [deviceRegistration] is invalid for the specified device or uses a device ID which has already been used as part of registration of a different device
          * @throws UnsupportedOperationException in case deployment failed since not all necessary plugins to execute the study are available.
          */
         internal suspend fun initialize(
```
patrickusiewicz, line 38: How will `unregister` come into play with the StudyRuntime? Do you have to unregister a device before you initialize a StudyRuntime with a new deviceRegistration?

carp.deployment.core/src/commonMain/kotlin/dk/cachet/carp/deployment/application/DeploymentService.kt
```diff
@@ -77,21 +82,31 @@ interface DeploymentService
      * - a deployment with [studyDeploymentId] does not exist
      * - [masterDeviceRoleName] is not present in the deployment
      * - the [deploymentChecksum] does not match the checksum of the expected deployment. The deployment might be outdated.
-     * @throws IllegalStateException when the deployment cannot be deployed yet.
+     * @throws IllegalStateException when the deployment cannot be deployed yet, or the deployment has stopped.
      */
     suspend fun deploymentSuccessful( studyDeploymentId: UUID, masterDeviceRoleName: String, deploymentChecksum: Int ): StudyDeploymentStatus
```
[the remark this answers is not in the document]
Whathecode: We don't really do command/query separation. Most of the requests always return status after doing the command, so the same comments goes for almost all requests. This is more of a convenience so you don't need two requests to have latest state (or infer it yourself).

Personally not clear to me how prepending with `set` would make that clearer. :/ That said, not entirely against using `set`. I think I didn't since that sounded like something you could set 'multiple times'. It is really more of an event.

carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntime.kt
Whathecode: Good question. Currently, seems like it, yes. But something to discuss what a good API would be during the meeting.

patrickusiewicz: Yes agree, it is not important for this PR. Still wanted to ask the question so we can agree it is something we will look at in the future.

carp.deployment.core/src/commonMain/kotlin/dk/cachet/carp/deployment/application/DeploymentService.kt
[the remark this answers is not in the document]
ltj: I'm not religious about this, so maybe just add to the docs what firing this implies

carp.deployment.core/src/commonMain/kotlin/dk/cachet/carp/deployment/domain/StudyDeployment.kt
```diff
@@ -34,6 +34,9 @@ class StudyDeployment( val protocolSnapshot: StudyProtocolSnapshot, val id: UUID
         data class DeviceUnregistered( val device: AnyDeviceDescriptor ) : Event()
         data class DeviceDeployed( val device: AnyMasterDeviceDescriptor ) : Event()
         data class DeploymentInvalidated( val device: AnyMasterDeviceDescriptor ) : Event()
+        // TODO: Immutable base class does not allow this to be defined as `object Stopped : Event()`.
+        //       It requires a data class, but that does not make sense since an object would still be immutable.
+        data class Stopped( val ignoreThis: Unit = Unit ) : Event()
```
Whathecode, line 39: This is ugly and what I intended to possibly fix in this PR. But, then again, to consume this could you would just typecheck for `Stopped`. Removing the `ignoreThis` won't impact the consumer, so I guess I can easily fix this in the next release.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyServiceHost.kt
```diff
@@ -257,6 +259,49 @@ class StudyServiceHost(
...
+     */
+    override suspend fun getParticipantGroupStatusList( studyId: UUID ): List<ParticipantGroupStatus>
+    {
+        val study: Study? = repository.getById( studyId )
+        requireNotNull( study ) { "Study with the specified studyId is not found." }
+
+        // Get study deployment statuses.
+        // TODO: Participations are more logically stored per `studyDeploymentId`, rather than having to filter like this.
+        val studyDeploymentIds = study.participations.map { it.participation.studyDeploymentId }.toSet()
+        val studyDeploymentStatuses = deploymentService.getStudyDeploymentStatusList( studyDeploymentIds )
```
Whathecode, line 279: Right now `DeanonymizedParticipation` (type of `study.participations`) stores within it `Participation`, which stores `studyDeploymentId`.  Wherever lists of `DeanonymizedParticipation`s are returned for the same deployment, there is thus a duplication of `studyDeploymentId`. Furthermore, as shown in this code, it makes much more sense to within study store participations in a map keyed on `studyDeploymentId`. I thus wanted to remove `Participation` from `DeanonymizedParticipation` and instead store `participationId`. This might impact the repo implementation, @ltj .

carp.deployment.core/src/commonMain/kotlin/dk/cachet/carp/deployment/domain/StudyDeployment.kt
patrickusiewicz: Maybe `ignoreThis` should be private? Then the consumer cant access it.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyServiceHost.kt
ltj: I remember this being a bit weird but also very aligned with how participations are stored across deployments and studies. No problem with the proposed solution however

carp.deployment.core/src/commonMain/kotlin/dk/cachet/carp/deployment/domain/StudyDeployment.kt
ltj: Remove the Immutable base class then. There's no point in having it here imho

Whathecode: It is defined in the main DDD service interface. It makes sense to enforce immutable events in this pattern. I considered removing it, but then I have to remove it everywhere so I didn't. I'll look at it after release.

ltj: But isn't the sealed class - data class pattern immutable in nature? I get why you use Immutable in the context of extendability and your serializer, but here it just seems...weird :)

PR #123, opened 2020-04-21
Add studyDeploymentId field to ParticipationAdded event

ltj:
We need to store the studyDeploymentId alongside study participations. This field was missing in `ParticipationAdded` event forcing a more complex repo implementation to search through the participations map in the study, rather than just getting it directly.


ltj:
Thanks @Whathecode . I know it's tiny, but can we turn it into an alpha-16 release? Maybe with some small goodies you have lying around...or not :)

Whathecode:
Can't you just use the snapshot release during development (on a new branch)? That's what it is there for.

Whathecode:
Can't you just use the snapshot release during development (on a new branch)? That's what it is there for.

ltj:
Yes! I forgot about the marvellous ci-cd setup

Issue #55, opened 2019-08-23
Use Kotlin `Duration` instead of custom `TimeSpan`


Whathecode:
[Kotlin 1.3.50 has been released.](https://blog.jetbrains.com/kotlin/2019/08/kotlin-1-3-50-released/)

This includes a _"new Duration and Time Measurement API"_. This means that [our custom `TimeSpan` class](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/TimeSpan.kt) can probably be phased out and be replaced with the Kotlin API.

Whathecode:
I tried using it, but currently it is still marked as experimental, and furthermore no default serializer is included. Let's consider this at a later point in time.

PR #124, opened 2020-05-02
Add support for sampling configurations

Whathecode:
Well, this feature got a bit out of hand. 😮 The original requirement was straightforward enough: _allow specifying an interval at which to measure stepcount._ But, it turned out specifying the interval at which to measure anything falls within the realm of what [early analysis](https://docs.google.com/document/d/1TX3htEBw8v2g3QqwKB5Z1JbZqfFPKIrNo6JsMrmDIw4/) indicated should be supported by **'sampling schemes'**:

> Specification of sampling settings and constraints for data streams available on a device (e.g., sampling frequency).

Since this was still underspecified, I started out by doing [requirements analysis on sampling schemes](https://docs.google.com/document/d/1XQksaIxvFh-AjWh8gIKwuugQO3eNGX0F0CXQBJiQKBQ/). As part of reviewing this PR, please review these requirements as well; any additional input is useful. Down the road, I will incorporate these in the RAD/SDD.

One of the main findings is a difference between:

- A **'sampling scheme'** which specifies the possible configuration options, defaults, and constraints.
- A **‘sampling configuration’** which contains the actual configuration on how to sample.

A sampling scheme relates to a `DataType`. But, the specific _defaults and constraints_ of a scheme (instance) can be specific to a `Measure` (e.g., constraints of stepcount on Smartphone represented as `PhoneSensorMeasure` vs. a future potential `FitbitMeasure`) and specific to device registrations (e.g., different vendors of smartphones, all represented by the `Smartphone` device descriptor).

A sampling configuration can be used to override defaults, either on a per-device level, or on a per-measure level (the latter is not yet implemented but would be easy to add as an additional member on `Measure`). As far as I can tell this will cover most use cases. Later, we can allow specifying triggers to change device defaults, e.g., to collect less data when battery is low.

This PR includes:

- `DeviceDescriptor` can store new _default_ sampling configurations for available sensors on the device, identified by `DataType`, which override the defaults for the `DeviceDescriptor`.
- A DSL for setting up `DeviceDescriptor`'s (exemplified by `Smartphone`). Check the [updated protocols code sample in the README](https://github.com/cph-cachet/carp.core-kotlin/pull/124/files#diff-04c6e90faac2675aa89e2176d2eec7d8) to see what this looks like.
- Introduction of `DataTypeSamplingScheme` which stores type information and sampling configuration options (and builders). This allowed considerable refactoring. I removed the `PhoneSensorMeasureFactory` interface since the implementation and interface were too interlinked.
- Moved access to sensor measures to `Smartphone.Sensors`, anticipating the need to add others later such as surveys.
- Other minor changes that don't really need a review (bugfix, detekt 1.8.0 update, additional `TimeSpan` methods).

Todo:

- [x] Allow `DeviceDescriptor`-specific defaults for `SamplingConfiguration`. Likely, instances of `DataTypeMetadata` should be used rather than `object`s.
- [x] Consider renaming `DataTypeMetadata` to `DataTypeSamplingScheme`, since it seems like that is what it might end up being.
- [x] Look at whether sampling configuration options can be composed. What if you want to be able to set both an interval and reuse that configuration, but add another? -> Won't do this for now. It is definitely not straightforward and we do not know enough about recurring configuration options so this might introduce unnecessary/premature abstractions.
- [x] Add unknown polymorphic serializer for `SamplingConfiguration`.
- [x] Add tests.


carp.common/src/commonMain/kotlin/dk/cachet/carp/common/TimeSpan.kt
```diff
@@ -31,15 +33,37 @@ data class TimeSpan(
          * The value parameter is converted to microseconds, and that number of microseconds is used to initialize the new [TimeSpan].
          * Therefore, value will only be considered accurate to the nearest microsecond.
          */
-        fun fromMilliseconds( ms: Double ): TimeSpan
-        {
-            val ticks: Long = (ms * MICROSECONDS_IN_MS).toLong()
-            return TimeSpan( ticks )
-        }
+        fun fromMilliseconds( ms: Double ): TimeSpan = TimeSpan( (ms * MICROSECONDS_IN_MS).toLong() )
```
ltj, line 36: Why are parameters doubles? They should be integers as well.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/TimeSpan.kt
```diff
@@ -18,6 +18,8 @@ data class TimeSpan(
     companion object
```
ltj, line 18: I'm really not fond of this class.
- It implements stuff that has been very well implemented before (e.g. java.time)
- Timespan is ambiguous. Duration (amount and unit) and Period (ISO cal) is not. It's really a duration.
- A duration should have a seconds part (long) and a nano second component (long). Best precision and temporal span. Then convert in and out using chrono units

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/TimeSpanTest.kt
```diff
@@ -16,6 +16,20 @@ class TimeSpanTest
         assertEquals( 10.0, ten.totalMilliseconds )
     }
 
+    @Test
+    fun fromSeconds_with_whole_numbers_succeeds()
```
ltj, line 20: You'll want property based testing here. Also a good way to find the max values

carp.protocols.core/src/commonMain/kotlin/dk/cachet/carp/protocols/domain/data/IntervalSamplingConfiguration.kt
```diff
@@ -0,0 +1,22 @@
...
+import dk.cachet.carp.common.TimeSpan
+import dk.cachet.carp.protocols.domain.devices.DeviceDescriptor
+import kotlinx.serialization.Serializable
+
+
+/**
+ * A sampling configuration which allows configuring the time [interval] in between subsequent measurements.
+ */
+@Serializable
+data class IntervalSamplingConfiguration( val interval: TimeSpan ) : SamplingConfiguration()
```
ltj, line 12: The period vs frequency discussion again (see comment in design doc). By using a frequency in Hz you'll also allow the client to do related time calculations however it see fit.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/TimeSpan.kt
Whathecode: I'm pretty much just cloning the C# API for this: https://docs.microsoft.com/en-us/dotnet/api/system.timespan.fromseconds?view=netcore-3.1 :)

They use doubles. It is a more flexible API, since you can also specify one minute and a half (`.fromMinutes( 1.5 )`). But more importantly, any output of a calculation doesn't lose precision, and doesn't force you to be converted to microseconds, to construct TimeSpan's for.

Whathecode: Maybe I should add overrides for Int's though. I am a bit annoyed by having to write `.o` in `.fromMinutes( 15.0 )`.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/TimeSpan.kt
Whathecode: Down the road this will be replaced with Kotlin's `Duration`. I looked into that prior to extending this class, but it is still experimental: https://github.com/cph-cachet/carp.core-kotlin/issues/55

Whathecode: And, premature optimization, _but_, also worth noting that CARP would never support ns precision (we pinned it at microsecond precision), so doubling the size to represent an otherwise fairly common construct feels unnecessary.

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/TimeSpanTest.kt
Whathecode: Yeah, I need to find a better testing framework. :( That'd be a major rewrite, though.

carp.protocols.core/src/commonMain/kotlin/dk/cachet/carp/protocols/domain/data/IntervalSamplingConfiguration.kt
Whathecode: Isn't interval <-> frequency a two-way calculation? I don't see that as an argument for or against using one or the other. What matters is for the data type what the most expressive metric is. I argue for stepcount it is definitely `interval` (or `duration`, `period`, ... whichever we agree on).

Issue #95, opened 2020-01-27
Should we use the Klock library for `DateTime`?


Whathecode:
Once more advanced `DateTime` behavior is needed, we should likely consider switching to the [korlibs/klock](https://github.com/korlibs/klock/) library.

Until then, I think it is easiest to stick to our own expect/actual `DateTime` reducing the dependencies we pull in until we use more of it (probably time-based triggers in the future might/will).

Whathecode:
Related: currently there is a `TimeOfDay` object, which seems to be the equivalent of [Java's `LocalTime`](https://docs.oracle.com/javase/8/docs/api/java/time/LocalTime.html). Is there an equivalent in the Klock library we can use?

For consistency, it might make sense to use the same naming as Java.

Whathecode:
Another potential library: https://github.com/Kotlin/kotlinx-datetime

PR #130, opened 2020-05-26
New triggers: `ElapsedTimeTrigger`, `ScheduledTrigger`, and `ManualTrigger

Whathecode:
More important than a review of the code, I would like feedback on the design choices described below.

Added three new triggers:

- `ElapsedTimeTrigger` replaces `StartOfStudyTrigger` and is more general purpose. A `ElapsedTimeTrigger` with `elapsedTime` set to 0 is the same as the former `StartOfStudyTrigger`.
- `ScheduledTrigger` uses the standard iCalender notation for recurring events.
- `ManualTrigger` can be initiated by users. These should be displayed somewhere in the user interface. Potential future options other than the current `label` and `description`: time intervals during which this trigger is enabled/disabled.

The first two triggers require the runtime to know about the 'start' of the study on the client, for which I added a 'start time' in `StudyDeployment`. Currently, the start time is determined by the time at which all required devices are deployed and all devices are registered. This ensures the same start time is used across all participating devices.

I chose to reuse the RRULE of the iCalendar format since it is a very common standard for which libraries on various platforms are available, rather than coming up with our own configuration options. Currently not all rules are supported, but the current options are a good MVP. The current `RecurrenceRule` itself is a potential candidate to be integrated into the Kotlin multiplatform Klock library: https://github.com/korlibs/klock/issues/115 In case we decide to start using Klock, we could contribute to this library by using their date time calculations to iterate over recurrence rules, thereby providing Kotlin multiplatform RRULE support.


Whathecode:
Modeling recurrent events is probably best done adopting iCal standards: https://stackoverflow.com/a/38359184/590790

To do so, I expect we will need to extend on TimeSpan and DateTime's compatabilities. Rather than doing so ourselves, this might be a good time to look into the Klock library: https://github.com/cph-cachet/carp.core-kotlin/issues/95

PR #141, opened 2020-07-01
Fix protocol description from snapshot

ltj:
The dangers of default values :) The protocol description was not set correctly when using  the `StudyProtocol.fromSnapshot()` factory.


Whathecode:
I'll add a unit test while at it. Let's see why it didn't catch it ...

ltj:
> I'll add a unit test while at it. Let's see why it didn't catch it ...

Pretty sure my addition to your unit test fixed that ;) Problem was that you didn't set a description in the test data and then we always compared empty string with empty string as the default value. If you remove the fix the tests will correctly fail now, so no need for changes there

Whathecode:
> > I'll add a unit test while at it. Let's see why it didn't catch it ...
> 
> Pretty sure my addition to your unit test fixed that ;) Problem was that you didn't set a description in the test data and then we always compared empty string with empty string as the default value. If you remove the fix the tests will correctly fail now, so no need for changes there

Exactly. The reason `assertEquals( protocol.description, fromSnapshot.description )` in `creating_protocol_fromSnapshot_obtained_by_getSnapshot_is_the_same()` passed was because the description was not set in the test object, so "" equaled "", but not for the right reasons. :)

Thanks!

PR #143, opened 2020-07-13
Add `creationDate` to `AggregateRoot`.

Whathecode:
I was about to add `creationDate` to `StudyDeployment`. It seems like `creationDate` can be expected to be a desirable property of any aggregate root domain object.

We currently had it in `Study` and `StudyProtocol`. Adding it to `StudyDeployment` would mean the only aggregate root that does not have it is `StudyRuntime`.

This PR also adds it to `StudyRuntime`; does this make sense or is it a premature generalization?

### DIFF
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/ddd/AggregateRoot.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/ddd/AggregateRoot.kt
@@ -1,5 +1,6 @@
 package dk.cachet.carp.common.ddd
 
+import dk.cachet.carp.common.DateTime
 import dk.cachet.carp.common.Immutable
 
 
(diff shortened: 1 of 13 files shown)


Whathecode:
@patrickusiewicz Since I'll do a release today, I do think it makes sense to pull in this change as well. Do you have time to have a quick look?

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/ddd/AggregateRoot.kt
```diff
@@ -9,6 +10,12 @@ import dk.cachet.carp.common.Immutable
  */
 abstract class AggregateRoot<TRoot, TSnapshot : Snapshot<TRoot>, TEvent : Immutable>
 {
+    /**
+     * The date when this object was created.
+     */
+    var creationDate: DateTime = DateTime.now()
```
patrickusiewicz, line 16: Isn't it possible to use `val` instead of `var` with `protected set`?

patrickusiewicz: I take my comment back. It is because `StudyRuntime` sets the date based on what comes from the `StudyRuntimeSnapshot`.

PR #158, opened 2020-09-10
Release 1.0.0-alpha.20

Whathecode:
- Use class discriminator `$type` instead of array polymorphism for JSON
- Stop using non-default `allowStructuredMapKeys` for JSON by turning `UUID` and `DataType` into primitive serializable types
- Upgrade to Kotlin 1.4.10
- Eliminate dependency on `kotlin.reflect` for main library (still used in testing).


patrickusiewicz (approved):
Looking forward to not have the kotlin.reflect lib take up space in our Android app!

PR #164, opened 2020-09-30
Add test to ensure correct `DataType` additions.

Whathecode:
@ltj Just a pull request so you can see where (and why) the CARP `DataType`s have been moved to.

To write a unit test to ensure correct implementations of future `DataType`s I had to be able to iterate them. This is where the `EnumObjectList<DataType>` comes in. I've used this before to list all available sampling configurations in `DeviceDescriptor`s.

### DIFF
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/data/CarpDataTypes.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/data/CarpDataTypes.kt
@@ -3,39 +3,48 @@
  */
 package dk.cachet.carp.common.data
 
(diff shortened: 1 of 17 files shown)


ltj (approved):
OK. I'm totally on board the concept I just need to think a bit more about the `EnumObjectList` approach as I feel it becomes a bit sugary :)

Whathecode:
@ltj It's the best solution I found so far. Full rationale is here: https://whatheco.de/2020/05/05/list-of-strongly-typed-objects-acting-like-enum-in-kotlin/

Of course always open to suggestions! I might be overlooking something.

Issue #13, opened 2018-07-03
Add data streams to device descriptors


Whathecode:
Devices should specify which data streams are available on them. The study protocol can verify whether `DataStreamMeasure`s specified as part of a `TaskDescriptor` are supported on the device they are triggered  for.

Important to keep in mind that 'secondary' data streams which process other data streams will later be added to `MasterDeviceDescriptor` as well.

Whathecode:
Data streams are not yet modeled as such, but in the following PR, supported data types on a device are modeled, which have a one-to-one relation to _device_ data streams (not 'master data streams' that are processed): https://github.com/cph-cachet/carp.core-kotlin/pull/124/

It should be easy to add a check to verify a `Measure` is supported on a device by adding the supported data types to the `DeviceDescriptor` interface.

PR #165, opened 2020-10-02
Data stream abstraction on client

Whathecode:
This models available data streams on devices and sets up a `DataCollector` dependency (to be implemented) to determine whether requested measures can be collected on the master device and connected devices as specified in the `MasterDeviceDeployment`.

This currently compiles, but some remaining tasks are:

- [x] Write unit tests to verify the exception which is thrown when `DataCollector` indicates requested measures cannot be collected.
- [x] Now that `DeviceDescriptor.supportedDataTypes` has been added, we can probably add deployment warnings in case measures are requested on a device for unsupported data types. For extensibility reasons we will likely want to support this, but a `DeploymentWarning` would still be in order.

After doing this, this PR probably closes: https://github.com/cph-cachet/carp.core-kotlin/issues/13


Whathecode:
@patrickusiewicz The main thing to look at would be:

- The current two methods in the [DataCollector interface](https://github.com/cph-cachet/carp.core-kotlin/pull/165/files#diff-13c7461d0613884e12a846276c7218af), and how it is DI'd into `StudyRuntime`.
- Warning which is triggered in the study protocol for unexpected data types: https://github.com/cph-cachet/carp.core-kotlin/pull/165/files#diff-392bab5b59e3cb07da0b38517cf89ec4
- Exception in `StudyRuntime` when a deployment contains a requested measure which is unsupported by `DataCollector`: https://github.com/cph-cachet/carp.core-kotlin/pull/165/files#diff-280a76766b834c5fd4f8bcb32954ab41R144

carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntime.kt
```diff
@@ -108,24 +117,51 @@ class StudyRuntime private constructor(
...
         if ( deviceStatus !is DeviceDeploymentStatus.NotDeployed || !deviceStatus.isReadyForDeployment ) return false
 
-        // TODO: Get deployment and throw exception in case there are missing plugins to perform the operations (e.g., measurements).
+        // Get deployment information.
         val deployment = deploymentService.getDeviceDeploymentFor( studyDeploymentId, device.roleName )
+        check( deployment.deviceDescriptor == device )
         deploymentInformation = deployment
 
+        // Verify whether data can be collected for all requested measures.
+        for ( deviceTasks in deployment.getTasksPerDevice() )
```
patrickusiewicz, line 145: Maybe you could wrap the for loop and its content into a function for easier readability?

carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntime.kt
```diff
@@ -108,24 +117,51 @@ class StudyRuntime private constructor(
...
+        {
+            val dataTypes = deviceTasks.tasks.flatMap { it.measures }.map { it.type }.distinct()
+            for ( type in dataTypes )
+            {
+                val canCollectData =
+                    if ( deviceTasks.isConnectedDevice )
```
patrickusiewicz, line 151: What does `isConnectedDevice` mean? Is it for external sensors?

carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntime.kt
```diff
@@ -108,24 +117,51 @@ class StudyRuntime private constructor(
...
+                    {
+                        dataCollector.canCollectDataForConnectedDevice(
+                            type,
+                            deviceTasks.device::class,
+                            deviceTasks.deviceRegistration )
+                    }
+                    else dataCollector.canCollectData( type )
+
+                check( canCollectData )
```
patrickusiewicz, line 160: Does `canCollectDataForConnectedDevice` try to connect to the external sensor? We cannot always guarantee connection to an external sensor at this point. The external sensor might be out of reach. If the external sensor is out of reach the caller of `tryDeployment` cannot determine which external sensor the client cannot collect data from. This could be vital information to communicate to the user of the client.

carp.protocols.core/src/commonTest/kotlin/dk/cachet/carp/protocols/domain/deployment/UnexpectedMeasuresWarningTest.kt
```diff
@@ -0,0 +1,50 @@
...
+class UnexpectedMeasuresWarningTest
+{
+    @Test
+    fun isIssuePresent_true_when_unexpected_measure_in_protocol()
+    {
+        val protocol = createEmptyProtocol()
+        val master = StubMasterDeviceDescriptor()
+        protocol.addMasterDevice( master )
+        val expectedMeasure = StubMeasure()
+        val unexpectedMeasure = StubMeasure( DataType( "namespace", "unexpected" ) )
```
patrickusiewicz, line 20: How come `unexpectedMeasure` is unexpected? Is it because it contains a `DataType`?

patrickusiewicz: Because 'StubMasterDeviceDescriptor.supportedDataTypes' is empty. Could you update the test to make it clearer to the reader what measures the `StubMasterDeviceDescriptor()` supports and doesn't? Maybe pass the `expectedMeasure` into `StubMasterDeviceDescriptor()`?

carp.protocols.core/src/commonMain/kotlin/dk/cachet/carp/protocols/domain/deployment/UnexpectedMeasuresWarning.kt
```diff
@@ -0,0 +1,40 @@
...
+     * Holds an unexpected [measure] which at some point in the [StudyProtocol] may be sent to [device].
+     */
+    data class UnexpectedMeasure( val device: AnyDeviceDescriptor, val measure: Measure )
+
+    override val description: String =
+        "The study protocol contains measures that are requested on a device for which the requested data type is not expected." +
+        "This is allowed, but requires the client implementation to have corresponding support to handle this unexpected data type."
+
+
+    override fun isIssuePresent( protocol: StudyProtocol ): Boolean = getUnexpectedMeasures( protocol ).any()
```
patrickusiewicz, line 26: How and when does the client determine which measures it supports?

carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntime.kt
Whathecode: I think that is a personal preference. I tend to not extract things out if there is no clear possibility for reuse. https://whatheco.de/2010/12/07/function-hell/

I gave this specific case some more thought and can't find a suitable place to extract the logic to. Not in `DataCollector` as it would introduce an unnecessarily strong link to `MasterDeviceDeployment`. It lives in the realm between `DataCollector` and `MasterDeviceDeployment`, which essentially is `StudyRuntime`. As a public function in `StudyRuntime` I think it leaks the abstraction it is trying to hide. As a private function in `StudyRuntime` it would only get called from one location.

I do agree this function is on the long side, but prefer to refactor/extract as it grows more and when it might make more sense.

carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntime.kt
Whathecode: I'll add documentation to `DeviceTasks`. But yes, it is external sensors, corresponding to `StudyProtocol.addConnectedDevice`.

carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntime.kt
Whathecode: No. It only determines whether the runtime has the necessary integrations loaded to connect to a sensor of that type with the given device registration.

carp.protocols.core/src/commonTest/kotlin/dk/cachet/carp/protocols/domain/deployment/UnexpectedMeasuresWarningTest.kt
Whathecode: The `DataType` within the measure is not listed in the supported data types of the device. `StubmasterDeviceDescriptor.supportedDataTypes` is not empty, it contains `STUB_DATA_TYPE`.

That is hidden away so I understand that is unclear. Think I will indeed allow caller to pass supported types to the stubs.

carp.protocols.core/src/commonMain/kotlin/dk/cachet/carp/protocols/domain/deployment/UnexpectedMeasuresWarning.kt
Whathecode: This is even before the study protocol (or rather, the `MasterDeviceDeployment` subset) reaches the client. These are warnings triggered during the creation of a study protocol, i.e., on the study manager.

carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntime.kt
Whathecode: `supportsDataCollection( DataType )` and `supportsDataCollectionForConnectedDevice`?

carp.protocols.core/src/commonTest/kotlin/dk/cachet/carp/protocols/domain/deployment/UnexpectedMeasuresWarningTest.kt
Whathecode: Fixed in https://github.com/cph-cachet/carp.core-kotlin/pull/165/commits/51bd4cb0212cad2a03c97485d3157fbd47da6e11

carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntime.kt
Whathecode: Fixed in https://github.com/cph-cachet/carp.core-kotlin/pull/165/commits/b84075b5841c10a502961e49857aa52a2bbfa458

carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntime.kt
Whathecode: Fixed in https://github.com/cph-cachet/carp.core-kotlin/pull/165/commits/088ff54924c56034bb573bb40e1d85204619e5a2

Whathecode:
I'll go ahead and merge this PR. I will continue on similar work to this, so if additional changes are needed to the fixes I applied I can still incorporate them in the next PR.

Thanks for the feedback!

PR #178, opened 2020-10-15
Release 1.0.0-alpha.23

Whathecode:
Updated dependencies:

- `kotlinx.serialization` to `1.0.0`
  - The collection initializers for JS which were previously used have been removed. The new TS declarations now require `ListSerializer`, `MapSerializer`, and `SetSerializer` initializer functions rather than the old `get_list`, `get_set`, and `get_map`. (https://github.com/cph-cachet/carp.core-kotlin/pull/178/commits/bea4dd6c7dcae555bd50171dce268f1597a535ac)

Protocols:

- Added new device registration: `BLESerialNumberDeviceRegistration` and `MACAddressDeviceRegistration`.
- Added new device descriptor: `BLEHeartRateSensor` (using `MACAddressDeviceRegistration`). (https://github.com/cph-cachet/carp.core-kotlin/pull/166)

Common:

- Added new data types: `Acceleration`, `RRInterval`, and `SensorSkinContact`, used by `BLEHeartRateSensor`.
- Bugfix: `UUID.stringRepresentation` was case sensitive. Made it case sensitive (lower case needed) and added a `parse` method to make initialization less strict (allowing upper case).


carp.common/src/commonMain/kotlin/dk/cachet/carp/common/MACAddress.kt
```diff
@@ -0,0 +1,49 @@
...
+            { "Invalid MAC address string representation. Expected six groups of two upper case hexadecimal digits, separated by hyphens (-)." }
+    }
+}
+
+
+/**
+ * Regular expression to verify whether the string representation of a [MACAddress] matches the IEEE 802 standard notation.
+ */
+val MACAddressRegex = Regex( "([0-9A-F]{2}-){5}([0-9A-F]{2})" )
+
```
ltj, line 37: May I suggest: `^([0-9A-Fa-f]{2}-){5}([0-9A-Fa-f]{2})$` IEEE actually states that `:` is also a valid delimiter but I'm fine with hyphen only. It should however be case insensitive

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/MACAddressTest.kt
```diff
@@ -0,0 +1,39 @@
...
+    {
+        assertFailsWith<IllegalArgumentException> { MACAddress( "Invalid" ) }
+        // Not long enough.
+        assertFailsWith<IllegalArgumentException> { MACAddress( "00-11-22-33-44" ) }
+        // Too long.
+        assertFailsWith<IllegalArgumentException> { MACAddress( "00-00-00-00-00-00-FF" ) }
+        // Invalid character.
+        assertFailsWith<IllegalArgumentException> { MACAddress( "G0-00-00-00-00-00" ) }
+        // Lower case.
+        assertFailsWith<IllegalArgumentException> { MACAddress( "aa-bb-cc-dd-ee-ff" ) }
```
ltj, line 35: must pass

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/MACAddress.kt
Whathecode: What the 'official' format is was a bit ambiguous. I did find somewhere '-' was the main one recommended by IEEE. Upper case and lower case wasn't specified, but all examples were upper case.

Regardless, of course we want to support easier initialization, but the main point here is the TODO note at the top:

> TODO: Allow more flexible [address] input and conversion to the common form. Make sure equality is ensured!

patrickusiewicz:
Why did you update `coroutines` to `1.4.0-M1`? Is there something wrong with version `1.3.9`?

build.gradle
```diff
@@ -9,13 +9,13 @@ buildscript {
...
-        globalVersion = '1.0.0-alpha.22'
+        globalVersion = '1.0.0-alpha.23'
 
         versions = [
             // Kotlin multiplatform versions.
             kotlin:'1.4.10',
-            serialization:'1.0.0-RC2',
-            coroutines:'1.3.9',
+            serialization:'1.0.0',
+            coroutines:'1.4.0-M1',
```
patrickusiewicz, line 18: Why did you update `coroutines` to `1.4.0-M1`? Is there something wrong with version `1.3.9`?

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/MACAddressTest.kt
```diff
@@ -0,0 +1,39 @@
...
+
+        val json = createDefaultJSON()
+        val serialized = json.encodeToString( MACAddress.serializer(), mac )
+        val parsed = json.decodeFromString( MACAddress.serializer(), serialized )
+
+        assertEquals( mac, parsed )
+    }
+
+    @Test
+    fun cant_initialize_incorrect_MACAddress()
```
ltj, line 25: Tests should be updated if you agree to the more relaxed format

build.gradle
Whathecode: It is the version they claim is aligned with Kotlin 1.4.0. Since we are on Kotlin 1.4.0, I believe it makes sense to do so.

In general, I try to upgrade dependencies each release. That way we catch changes sooner and upgrading becomes a smaller task. For example, upgrading dokka has been postponed for ages since it is a big task. :(

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/MACAddressTest.kt
Whathecode: I agree to a more relaxed format. But introducing that would be a backwards compatible change. It is simply something I did not want to spend time on now. If you think it is essential, I can work on it as part of this release.

build.gradle
patrickusiewicz: Version `1.3.9` is already aligned with kotlin 1.4.0 see [link](https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.3.9). `1.4.0-M1` isn't a production release and it does contain some breaking changes. Therefor I think it makes sense to keep `1.3.9` until `1.4.0` is released.

Whathecode: Okay. Let me know when you need this update. Will leave it at 1.3.9 until I hear a request.

patrickusiewicz (approved):
Besides the comments made by Lars and I, everything looks good!

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/MACAddress.kt
ltj: Fine with leaving the more strict format as is for now, but unless you add the meta characters matching the string boundaries _or_ require a string of the correct length, then the regex will match any string _containing_ a valid mac

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/MACAddressTest.kt
ltj: Either way you need this test case:
`assertFailsWith<IllegalArgumentException> { MACAddress( "hey you and the rock-steady AB-CD-EF-12-34-45 crew" ) }`

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/MACAddress.kt
Whathecode: I did it in the meantime awaiting your answer. =)

I thought that as well (and a similar problem would exist for UUID validation), but that's why I added the 'too long' test:
`assertFailsWith<IllegalArgumentException> { MACAddress( "00-00-00-00-00-00-FF" ) }`

I think this specific API does full matching, but I'd have to verify.

ltj: It passes because of incorrect group size, no. But groups are determined by the delimiter.

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/MACAddressTest.kt
ltj: ...of course with no valid delimiter on either side of the mac. So get rid of the hyphen :)

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/MACAddress.kt
ltj: If you just use the boundary matching regex you can leave the first require statement entirely

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/MACAddressTest.kt
Whathecode: assertFailsWith<IllegalArgumentException> { MACAddress( "... AB-CD-EF-12-34-56 ..." ) } passes. `Regex.matches` always verifies the _entire_ input: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/-regex/matches.html

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/MACAddress.kt
ltj: Even better. You can just use my regex in its entirety. Your formatting will always ensure the field has the stricter format, but allow more relaxed input. win-win

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/MACAddressTest.kt
ltj: Ok interesting semantics there.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/MACAddress.kt
Whathecode: No. I also thought of that, but `... AB-CD-EF-12-34-56 ...` passes and are 6 groups. It passes because the `matches` API does not do what you think it does AFAIK.

The first require statement is there for different reasons; to disallow matching `AB:CD-EF:12-34-56`.

And, the regex is public. As is, the regex can be reused to also find matches _anywhere_ in a sentence. Adding ^ or $ wouldn't offer that reusability.

ltj: So you don't need the boundaries then bc `Regex` semantics. But the point remains. You don't need the grouping check. Simply match and format.

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/MACAddressTest.kt
Whathecode: https://www.baeldung.com/kotlin-regular-expressions This is the difference between `containsMatchIn` and `matches`. I am using `matches`, so it works as expected. No bug. :)

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/MACAddress.kt
ltj: parse() factory and strict constructor. boom

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/UUID.kt
```diff
@@ -20,15 +22,22 @@ expect class UUID( stringRepresentation: String )
 
     companion object
     {
+        /**
+         * Parse common [UUID] representations, also allowing upper case hexadecimal digits.
+         *
+         * TODO: It might be useful to allow even more flexible [uuid] entry (e.g., surrounded by curly braces, etc.).
```
ltj, line 28: Just as a note. Parse could allow uuid with or without hyphens (some systems like nginx uses this more compact format) and of course be case insensitive. But I would refrain from adding too much smarts to the parse function like finding uuids in arbitrary strings an such. Imho it cleaner if it behaves more like the value type parse methods or the date ditto.

ltj (approved):
yup

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/UUID.kt
Whathecode: We can always add extra functionality now in a backwards compatible way. That's why I left the TODO. :) {aaa-...} is also fairly common (Microsoft).

PR #187, opened 2020-10-26
Add `ProtocolFactoryService.createCustomProtocol`.

Whathecode:


benceburom (approved):

ltj:
While I do like this kind of simple, pragmatic extensibility I'm generally reluctant to the idea of storing arbitrary data structures in strings with zero meta-data to describe it. You could argue that we don't care because it's the responsibility of the device to know what to do, but the service host implies more that one actor is taking part in the exchange of this data. A string also implies encoding, and here it's implicitly UTF-8 unless configured otherwise. It would almost make more sense to require that the data can be serialized as a byte array if it is truly "unstructured".

The test seems to suggest json however, so another option if that really is the case, is to use a json type.

Whathecode:
> While I do like this kind of simple, pragmatic extensibility I'm generally reluctant to the idea of storing arbitrary data structures in strings with zero meta-data to describe it. You could argue that we don't care because it's the responsibility of the device to know what to do, but the service host implies more that one actor is taking part in the exchange of this data. A string also implies encoding, and here it's implicitly UTF-8 unless configured otherwise. It would almost make more sense to require that the data can be serialized as a byte array if it is truly "unstructured".
> 
> The test seems to suggest json however, so another option if that really is the case, is to use a json type.

What are the downsides? Now the end user, who decides how to interpret it, can use any UTF-8 format. The backend can't do anything extra afaik by restricting it to json, so that would be an arbitrary restriction.

It is UTF-8 by necessity since that is the format of the container within which it is stored. If they want binary they can use bson, I guess?

Whathecode:
> It is UTF-8 by necessity since that is the format of the container within which it is stored.

Mmm, coming to think of it, that is not right. That is a serialization concern only. I do know byte-level support was limited on multiplatform last time I checked, so that might be a practical reason to use string.

And, I somehow doubt that people using a custom format are more likely to use binary than string formats. 🤔

PR #188, opened 2020-10-27
Add `ClientManager.stopStudy()`.

Whathecode:


carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntime.kt
```diff
@@ -123,6 +131,7 @@ class StudyRuntime private constructor(
             deploymentInformation == null -> StudyRuntimeStatus.NotReadyForDeployment( id )
             remainingDevicesToRegister.isNotEmpty() ->
                 StudyRuntimeStatus.RegisteringDevices( id, deploymentInformation!!, remainingDevicesToRegister )
+            isStopped -> StudyRuntimeStatus.Stopped( id, deploymentInformation!! )
```
patrickusiewicz, line 134: When we want to support pause/resume data collection and add the state for it in `StudyRuntime`, ordering will be important in this `when` statement. We could merge `isDeployed`, `isStopped` and etc. into one Study State field using `enum` or the `StudyRuntimeState`. Then ordering shouldn't matter. It is something we can look at after this PR.

patrickusiewicz (dismissed):
:tada:

carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntime.kt
Whathecode: pause/stop state will be orthogonal to this not ready/registering/stopped. It will probably just become a boolean in `Deployed`.

PR #190, opened 2020-10-29
Start and stop data collectors for a `StudyRuntime`

Whathecode:
This is a temporary implementation which allows clients to implement 'all or nothing' data collection. Triggers are not yet evaluated and are simply ignored.

`startDataCollectors` and `stopDataCollectors` simply requests the relevant data collectors to `start` and `stop` collecting data for the data types that might ever be triggered in the study protocol.

Later, this will be replaced with starting/stopping data collection based on trigger evaluation, and based on a 'privacy mode' for which we still need to investigate requirements.


ltj (approved):

Whathecode:
This is out of date and was a temporary hackish PR. Therefore, I will close this.

Issue #172, opened 2020-10-12
Model whether `DataPoint` for a `DataType` expects a single point in time or time interval.


Whathecode:
Currently it can be unclear from `DataType` documentation whether the headers of `DataPoint`'s in which they are used should have just a start time set, or also the end time (representing a time interval).

Can we model this somewhere? Can both be allowed, or is it always one or the other?

Related (other `DataType` meta data): [Store whether or not data type pertains to a participant](https://github.com/cph-cachet/carp.core-kotlin/issues/132)

Issue #132, opened 2020-06-09
Store whether or not data type pertains to a participant


Whathecode:
Data points have an optional 'participation ID'. This does not need to be set when the data pertains to e.g., the weather, or smart home sensors (cupboard closed). But, it needs to be set if the data pertains to a participant, e.g., stepcount.

To enable validation in the infrastructure whether or not a data point is valid, it is relevant to know whether or not data is expected to pertain to a participant.

This might be relevant to store in [`DataTypeSamplingScheme`](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/carp.protocols.core/src/commonMain/kotlin/dk/cachet/carp/protocols/domain/data/DataTypeSamplingScheme.kt).

Whathecode:
This could in fact be specific to instances of sampling schemes. For example, while geolocation on a smart phone might pertain to a participant, it does not necessarily pertain to a participant in the case of a moving beacon (e.g., to track the movement of an object in a home).

Whathecode:
The `participantId` was omitted from data points when the data subsystem was added to carp core.

However, currently, I am in the process of introducing "participant roles" to study protocols. If we end up _inviting_ participants to take part in a study, they could be invited to use the same device. In that case, it could be useful to associate the participant role that is currently using the device to the data points.

This is not a common scenario we intend to support right now, but this issue may become relevant in the future.

PR #192, opened 2020-11-02
Split `ParticipantService` off from `StudyService`

Whathecode:
This is an early PR from an incomplete feature branch which so far has focused on introducing `ParticipantService` as a new application service separate from `StudyService`.

This is done because:
- Additional participant management functionality will be added next (participant meta-data).
- The only real dependency of `ParticipantService` on `StudyService` is `StudyRepository` to verify whether a study exists.
- `StudyServiceHost` no longer needs to depend on `DeploymentService`; only `ParticipantServiceHost` does.
- We can envision `ParticipantService` becoming a separate microservice down the line, with eventual consistency in terms of study deletion/creation (constraints on `studyId` to which participants are added to).

One other unrelated commit is `FullyQualifiedName` is introduced and used in `DataType`. (https://github.com/cph-cachet/carp.core-kotlin/commit/28881a6fdfee97455edf0e5d5dddb2ee35565c6b) I added this in anticipation of needing some universal key registration for participant metadata, which would also benefit from fully qualified names.


carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/ParticipantService.kt
```diff
@@ -0,0 +1,60 @@
...
+import dk.cachet.carp.common.UUID
+import dk.cachet.carp.studies.domain.ParticipantGroupStatus
+import dk.cachet.carp.studies.domain.users.AssignParticipantDevices
+import dk.cachet.carp.studies.domain.users.Participant
+
+
+/**
+ * Application service which allows adding participants to studies and creating deployments for them.
+ */
+interface ParticipantService
```
Whathecode, line 13: Nothing really new here. These are just methods moved from `StudyService` to this new interface.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/ParticipantServiceHost.kt
```diff
@@ -0,0 +1,180 @@
...
+import dk.cachet.carp.studies.domain.StudyRepository
+import dk.cachet.carp.studies.domain.users.AssignParticipantDevices
+import dk.cachet.carp.studies.domain.users.DeanonymizedParticipation
+import dk.cachet.carp.studies.domain.users.Participant
+import dk.cachet.carp.studies.domain.users.ParticipantRepository
+import dk.cachet.carp.studies.domain.users.deviceRoles
+import dk.cachet.carp.studies.domain.users.participantIds
+
+
+class ParticipantServiceHost(
```
Whathecode, line 20: Nothing really new here. These are just methods moved from `StudyService` to this new interface.

Most relevant is to note that this service depends on `StudyRepository` as well to verify whether the study exists, and to add participations when deploying participant groups.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyServiceHost.kt
```diff
@@ -1,35 +1,20 @@
...
 
 /**
  * Implementation of [StudyService] which allows creating and managing studies.
  */
-@Suppress( "TooManyFunctions" ) // TODO: Perhaps split up participation management from main interface.
-class StudyServiceHost(
-    private val repository: StudyRepository,
-    private val deploymentService: DeploymentService
-) : StudyService
+class StudyServiceHost( private val repository: StudyRepository ) : StudyService
```
Whathecode, line 17: `StudyService` no longer depends on `DeploymentService`!

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/ParticipantRepository.kt
```diff
@@ -0,0 +1,22 @@
+package dk.cachet.carp.studies.domain.users
+
+import dk.cachet.carp.common.UUID
+
+
+/**
+ * Store [Participant] instances, linked to study IDs.
+ */
+interface ParticipantRepository
```
Whathecode, line 9: Nothing new here. Participants are now simply stored in a separate repository from `StudyRepository` since we can envision it becoming a separate microservice down the line.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/FullyQualifiedName.kt
```diff
@@ -0,0 +1,74 @@
...
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+
+
+/**
+ * A uniquely named entity within a [namespace].
+ * TODO: This likely needs more restrictions in line with other standards. Disallow certain symbols? Lower-case only?
+ */
+@Serializable( with = FullyQualifiedNameSerializer::class )
+data class FullyQualifiedName(
```
ltj, line 17: I don't think this is as general as you imply. It's only used for data types. If you want to use the same validation elsewhere, then move it out into an `FQNValidator` or something. DataType becomes an unnecessary wrapper around a generalization that, well, is its own identity.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/ParticipantService.kt
```diff
@@ -0,0 +1,60 @@
...
+{
+    /**
+     * Add a [Participant] to the study with the specified [studyId], identified by the specified [email] address.
+     * In case the [email] was already added before, the same [Participant] is returned.
+     *
+     * @throws IllegalArgumentException when a study with [studyId] does not exist.
+     */
+    suspend fun addParticipant( studyId: UUID, email: EmailAddress ): Participant
```
ltj, line 21: We do currently support usernames instead of emails. Is the endpoint for adding username participants deliberately left out?

ltj (commented):
Overall OK. I left a few comments. Nothing big, but I don't like the FullyQualifiedName as a general abstraction

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/FullyQualifiedName.kt
Whathecode: I added this because the next feature I'll add likely also needs FQN.

> I added this in anticipation of needing some universal key registration for participant metadata, which would also benefit from fully qualified names.

A `FQNValidator` does sound like a likely cleaner solution!

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/ParticipantService.kt
Whathecode: I guess so. Afaik, it was never there before.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/data/DataType.kt
```diff
@@ -16,64 +17,34 @@ import kotlinx.serialization.encoding.Encoder
...
-     * To prevent conflicts, a reverse domain namespace is suggested: e.g., "org.openmhealth" or "dk.cachet.carp".
-     */
-    val namespace: String,
     /**
      * Describes the data being collected (e.g., "acceleration", "stepcount", "audio"), but not the sensor (e.g., "accelerometer, "pedometer").
-     *
-     * The name may not contain any periods. Periods are reserved for namespaces.
      */
-    val name: String
+    val name: FullyQualifiedName
```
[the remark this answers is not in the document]
Whathecode: What needs to be refactored? Maybe the constructor? I could reintroduce the secondary constructor with `namespace`, `name`. Wrapper or not, that might be useful.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/FullyQualifiedName.kt
Whathecode: In essence, `DataType` _is_ a FQN. I don't see it deviating from it. Anything that might deviate from it is meta-data attached to that FQN. 🤔

Whathecode: Related:
- https://github.com/cph-cachet/carp.core-kotlin/issues/172
- https://github.com/cph-cachet/carp.core-kotlin/issues/132

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/data/DataType.kt
```diff
@@ -16,64 +17,34 @@ import kotlinx.serialization.encoding.Encoder
...
-        {
-            val segments = fullyQualifiedName.split( '.' )
-            require( segments.count() > 1 ) { "A namespace needs to be specified." }
-
-            val namespace = segments.subList( 0, segments.size - 1 ).joinToString( "." )
-            val name = segments.last()
-
-            return DataType( namespace, name )
-        }
+        fun fromFullyQualifiedName( fullyQualifiedName: String ): DataType =
```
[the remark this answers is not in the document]
Whathecode: Well-spotted! I think it makes sense to use `fromString/toString` everywhere. It is clear they match, ... regardless of how common `parse` is.

ltj (approved):
👍

PR #195, opened 2020-11-04
Consistency in fromString/toString and parse/format

Whathecode:
The types in `carp.common` used `toString`, `fromString`, and `parse` inconsistently. This pull request tries to standardize their meaning:

- `toString` and `fromString` should be lossless.
- `parse` and `format` are lossy and may support non-default (lenient) formats.

In addition, @ltj and I agreed that serializers should use 'strict' formats, not lenient alternative formats. I took advantage of that to create a reusable factory method to create primitive string serializers.

Lastly, I noticed that `DateTime` was being serialized as a string, even though it serializes a long. I fixed that and also introduced a factory method to create primitive long serializers.

We can now use these factory methods to easily create primitive serializers for the remaining 'primitive' types in common:

- [x] `EmailAddress`
- [x] `TimeSpan`
- [x] `Trilean` (ignored, since this is currently not serialized anywhere)


carp.common/src/commonMain/kotlin/dk/cachet/carp/common/DateTime.kt
```diff
@@ -17,27 +13,22 @@ expect class DateTime( msSinceUTC: Long )
...
     {
         fun now(): DateTime
     }
 
+
     /**
      * Output as ISO 8601 UTC date and time in extended format with day and seconds precision with 3 decimal places following a period.
      * E.g., "2020-01-01T12:00:00.000Z"
      */
-    override fun toString(): String
```
ltj, line 29: almost all date/libs will return the ISO string on `toString() ` . It is even lossless in the sense that the seconds fraction will provide the precision available. I think you should stick to that convention. It will not violate the (sound) principles you outline in the PR description.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/DateTime.kt
```diff
@@ -17,27 +13,22 @@ expect class DateTime( msSinceUTC: Long )
...
+    fun defaultFormat(): String
```
ltj, line 27: Output to specific and lossy representations should be named after the representation i.e. `toSeconds()`, `toISOString()` etc.

I guess all this will be fixed with kotlinx datetime ;)

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/EmailAddress.kt
```diff
@@ -1,11 +1,22 @@
...
 import kotlinx.serialization.Serializable
 
 
 /**
  * Represents the address of an electronic mail sender or recipient.
  * TODO: Validate so only valid email addresses can be passed.
  */
-@Serializable
+@Serializable( EmailAddressSerializer::class )
 data class EmailAddress( val address: String )
```
ltj, line 13: Email should have some kind of input validation. An email regex for instance

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/RecurrenceRule.kt
```diff
@@ -43,13 +45,62 @@ data class RecurrenceRule(
...
 
-    init
-    {
-        require( interval >= 1 ) { "Interval needs to be 1 or more." }
+        /**
+         * Initialize a [RecurrenceRule] based on a [rrule] string.
+         */
+        fun fromString( rrule: String ): RecurrenceRule
+        {
+            require( RecurrenceRuleRegex.matches( rrule ) ) { "Invalid or unsupported RecurrenceRule string representation." }
```
ltj, line 54: Just a thought. We could define validators as external functions `f(input: String): Boolean` - leaving you with a single line require in the init block and a reusable function that can be used elsewhere like UI input validation.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/serialization/CarpPrimitiveSerializer.kt
```diff
@@ -0,0 +1,53 @@
...
+
+
+const val NAMESPACE = "dk.cachet.carp.common"
+
+
+/**
+ * Creates a primitive serializer for CARP types in the [dk.cachet.carp.common] namespace
+ * which are serialized using `toString` and deserialized using [fromString].
+ */
+internal inline fun <reified T> createCarpStringPrimitiveSerializer( noinline fromString: (String) -> T ) =
```
ltj, line 18: Very neat idea.

ltj (approved):
Nice. A couple of comments, and generally I think kotlinx.datetime is about being ready enough for out risk appetite :)

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/DateTime.kt
Whathecode: I was wondering about that! I tried it and got an equality exception which I expected was due to rounding, but maybe there was another cause.

Will revisit.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/EmailAddress.kt
Whathecode: The TODO is there. Out of scope for this PR.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/DateTime.kt
Whathecode: Solved in https://github.com/cph-cachet/carp.core-kotlin/pull/195/commits/a390e8ce7763a1d22526067a784506b2948a8b93

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/DateTime.kt
Whathecode: Solved in https://github.com/cph-cachet/carp.core-kotlin/pull/195/commits/a390e8ce7763a1d22526067a784506b2948a8b93

`toString` is simply used now and ISO is the default.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/RecurrenceRule.kt
Whathecode: Isn't this an 'external' function? This is defined in a companion object. The only difference would be it would return true/false instead of `RecurrenceRule`. So, if there is ever a use case in which you expect erroneously formatted RRule's this would make sense.

I don't see that right now.

ltj (approved):
🌮

PR #197, opened 2020-11-06
Split `ParticipationService` off from `DeploymentService`

Whathecode:
Similar to how I split off `ParticipantService` from `StudyService`, I split off a `ParticipationService` from `DeploymentService` so that the code is more maintainable. Only `ParticipationService` depends on `AccountService`.

I need to add profile information which can be edited both by researchers and participants to `Participation`. I intend to add this logic to `ParticipationService` after this PR.

As part of this, I moved the existing `ParticipationService` (which is an internal _domain service_) function closer to where it is used, since I wanted to use the name `ParticipationService` for the new _application service_.

This PR also contains a trivial new `getParticipant` endpoint.


ltj (approved):
a-ok

Whathecode:
> a-ok

I realized this morning I forgot to update the README. And, maybe the repositories need to be split as well. Will investigate.

PR #201, opened 2020-11-22
Add domain model for `ParticipantAttribute`

Whathecode:
This functionality will allow registering required `ParticipantAttribute`s which are linked to expected `InputDataType`.

In a central `InputDataList`, `InputDataType`s are linked to `InputElement`s which define input constraints, and converter functions which enable converting the underlying data type for an `InputElement` to a `Data` object.

Lastly, there is also functionality included to support custom `ParticipantAttribute`s added by the researcher.

For ease of reviewing, it makes sense to walk over each separate commit independently; they build on top of one another.


carp.common/src/commonMain/kotlin/dk/cachet/carp/common/ddd/AggregateRoot.kt
```diff
@@ -22,6 +22,12 @@ abstract class AggregateRoot<TRoot, TSnapshot : Snapshot<TRoot>, TEvent : Domain
      */
     protected fun event( event: TEvent ) = events.add( event )
 
+    /**
+     * In case the instance this is called on is true, [createEvent] and add it to the event queue of this aggregate root.
+     */
+    protected fun Boolean.eventOnSuccess( createEvent: () -> TEvent ): Boolean =
```
ltj, line 28: The fp in me want's an even more generalized approach here that might even be defined in `common`. Some specific downsides to the concrete implementation are 

1. Unclear semantics: the function returns a boolean but it's always `this` re-emitted.
2. It takes a function as argument that returns an event type, but it is never used.

My suggestion is to add extensions to `Boolean` such as `andThen(block: () -> Any):  Boolean` where the return value must me provided at the call. This would make sense as you sometimes use the returned boolean, sometimes discard it.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/data/input/element/Text.kt
```diff
@@ -0,0 +1,17 @@
...
+import dk.cachet.carp.common.data.input.InputElement
+import kotlinx.serialization.Serializable
+import kotlin.reflect.KClass
+
+
+/**
+ * Text entry by the user.
+ */
+@Serializable
+data class Text( override val name: String ) : InputElement<String>
```
ltj, line 12: Maybe (future) validation rule parameters can be passed as object for the constructor here

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/users/ParticipantAttribute.kt
```diff
@@ -0,0 +1,100 @@
...
+{
+    /**
+     * Uniquely identifies the type of data represented by this participant attribute.
+     */
+    abstract val inputType: InputDataType
+
+    /**
+     * The unique identifier for this participant attribute.
+     */
+    abstract val id: NamespacedId
```
ltj, line 30: Is it just me or is only one ever needed? The name and namespace is simply only set to a unique value in the case of the custom attribute.

ltj (approved):
Oki doki. Left some comments. No deal-breakers per se

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/data/input/element/Text.kt
Whathecode: We can look at this once we start modeling constraints. I don't think we know enough here yet to make informed choices.

I.e., rules might be modular, but might still need to be initialized through configuration properties. Or, rules are polymorphic and serialized alongside as inner objects of `InputElement`.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/users/ParticipantAttribute.kt
Whathecode: Both are needed. All custom input is of type `CUSTOM_INPUT_TYPE`, which is required to be able to evaluate whether or not this is custom input in `inputToData`

The `id` serves a different purpose and just so happens to overlap for the default types (why use yet a different `NamespaceId`?)

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/ddd/AggregateRoot.kt
Whathecode: 1. The returned `Boolean` is never used, but it _could_ be useful for chaining, e.g.:

```
doSomething()
    .eventOnSuccess( Event.SomethingSuccess() )
    .eventOnFailure( Event.SomethingFailed() )
```

2. It does not return an event type, it returns an _event_. This is the main point of this helper; the event which is created is emitted.

Does this make it clearer? Any suggestions to make the semantics clearer? The suggestion doesn't seem to match the current behavior. A more generic solution would be something akin to: `doSomething().onSuccess( event( Event.SomethingSuccess() ) )` which is more verbose.

ltj: ad. 1 You're using the return value in multiple places in carp.protocols.core/src/commonMain/kotlin/dk/cachet/carp/protocols/domain/StudyProtocol.kt
ad. 2. I don't see I was that ambiguous here. It returns something of an event type == an event. Just saying this could be extended to all types, as a general concept of running a block of code _if_ the value is `true`

Perhaps putting this as an extension for Boolean was a bit of a stretch in the first place. If it should be focused on this very specific action, why not a helper function in the `AggregateRoot`

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/users/ParticipantAttribute.kt
ltj: or all custom input is in the _namespace_ `CUSTOM_INPUT_TYPE` and has a unique name (from uuid) which btw is how the id is implemented. The use of id is a bit confusing since it is just the data type after all. It needs to be bette documented exactly why you need both at least

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/ddd/AggregateRoot.kt
Whathecode: 1. Correct. Overlooked those cases.

2. I'm still unclear on the alternative you suggest or what that would look like. If not a (protected, thus only accessible by `AggregateRoot`) extension for Boolean, I think it would necessarily involve re-introducing the `if` checks this helper method abstracts away? https://github.com/cph-cachet/carp.core-kotlin/pull/201/commits/34c0deda80a8e9716bf76ecb19e630d3e693d9b8

Whathecode: Maybe a more generalizable abstraction is `eventOn( objectToCompare, createEvent )`? Thus, `.eventOn( true ) { Event.SomethingSuccess() }`

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/users/ParticipantAttribute.kt
Whathecode: I considered that, but then we can't use `CUSTOM_INPUT_TYPE_NAME` as custom `@SerialName`.

But, maybe we don't have to ...

PR #204, opened 2020-12-06
Get and set participant data

Whathecode:
This enables storing participant metadata linked to `StudyDeployment`s.

- Expected participant data is stored in the form of `ParticipantAttribute`s in `StudyProtocol`.
- `InputDataTypeList` now also registers converters to convert `Data` objects to input representations. This will be useful to render retrieved type safe data in the UI in a uniform manner.
- `ParticipantAttribute` now has a `isValidInput` and `isValidData` method:
  - `isValidInput` checks whether input matches the constraints of the corresponding `InputElement`, and that it can thus be converted to a data object.
  - `isValidData` checks whether the type of the passed data matches the expected type, while also evaluating the constraints of `InputElement` (a conversion to input is done to enable this).
- `ParticipationService` now has a `getParticipantData` and `setParticipantData` endpoint which will only take valid and expected data as configured in the study protocol. This is stored in a new `ParticipantGroup` which will also become the placeholder for consent.

This also includes other commits of things I ran into I wanted to clean up:

- Improved exception message for `ExtractUniqueKeyMap`
- Removed the use of custom `InvalidConfigurationError`
- Minor bugfix: some events were left behind in `fromSnapshot` initializers. I don't think anyone used them.

Still needed in subsequent PRs:

- Get participant data through `ParticipantService` in study subsystem (probably as part of the participant group status request).
- Potentially some factory methods in `ProtocolFactoryService` (or in `ProtocolService`) to add expected participant metdata without having to use the `StudyProtocol` domain objects.
- Add TypeScript descriptors where necessary so all of this can be accessed from the study manager.
- Move adding/getting participations from `StudyDeployment` into `ParticipantGroup`. This will allow removing the `DeploymentRepository` dependency in `ParticipationService` and depend on `DeploymentService` instead.


carp.common/src/commonMain/kotlin/dk/cachet/carp/common/ExtractUniqueKeyMap.kt
```diff
@@ -2,7 +2,8 @@ package dk.cachet.carp.common
 
 
 /**
- * A [Map] which automatically extracts the key of added values and throws an error when attempting to add an element with a key which is already present.
+ * A [Map] which automatically extracts the key of added values and
```
ltj, line 5: I don't fully understand the need for this construct. The semantics of putting an existing key into a mutable map is overwriting. If you don't want to able to overwrite keys you can use an immutable map. If what you want is really an "append only mutable map" then that should be its name: `AppendOnlyMutableMap`.

An error free version could look something like this:
`class AppendOnlyMap<K, V> : LinkedHashMap<K, V>() {
    override fun put(key: K, value: V): V? {
       return putIfAbsent(key, value)
    }

    override fun putAll(from: Map<out K, V>) {
        from.forEach { (k, u) -> putIfAbsent(k, u) }
    }
}`

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/users/ParticipantAttribute.kt
```diff
@@ -55,6 +54,26 @@ sealed class ParticipantAttribute
...
 
+    /**
+     * Determines whether [input] is valid and can be converted to a matching [Data] object associated to this attribute
+     * as registered in [registeredInputDataTypes] or [CustomInput] in case this is a [CustomParticipantAttribute].
+     *
+     * @throws UnsupportedOperationException when no input element is registered for this attribute.
+     */
+    fun <TInput> isValidInput( registeredInputDataTypes: InputDataTypeList, input: TInput ): Boolean
+    {
+        @Suppress( "UNCHECKED_CAST" )
```
ltj, line 65: Is it possible to avoid these by using the kotlin safe cast `as?` ?

carp.deployment.core/src/commonMain/kotlin/dk/cachet/carp/deployment/application/DeploymentService.kt
```diff
@@ -18,7 +17,7 @@ interface DeploymentService
     /**
      * Instantiate a study deployment for a given [StudyProtocolSnapshot].
      *
-     * @throws InvalidConfigurationError when [protocol] is invalid.
+     * @throws IllegalArgumentException when [protocol] is invalid.
```
ltj, line 20: We should really consider still having a hierarchy of exceptions pertaining to the domain. It is easier to handle these in a special way if needed in the implementations and they can be enriched with information and behavior specific to the domain unlike the generic unchecked exceptions

ltj (approved):
👍

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/ExtractUniqueKeyMap.kt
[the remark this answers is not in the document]
Whathecode: The main differences are:

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/users/ParticipantAttribute.kt
Whathecode: No. This has to do with limitations of generics on JVM. :( I could _not_ do the cast, in which case it would be `InputElement<*>` which can be inferred automatically.

But, then a cast is needed of `input` at:
`return isExpectedDataType && inputElement.isValid( input )`

This is thus working around generics limitations of the JVM runtime by using `inputElement.getDataClass().isInstance( input )` to verify whether or not the types match. `getDataClass` wouldn't be needed with true generics.

carp.deployment.core/src/commonMain/kotlin/dk/cachet/carp/deployment/application/DeploymentService.kt
Whathecode: Agreed we need to talk about this, but `InvalidConfigurationError` was mostly used haphazardly, so for now opted to simply have `IllegalArgumentException` uniformly for anywhere the _arguments_ are incorrectly.

Broadly speaking every `IllegalArgumentException` that comes out of services right now should safely map to a 400 of corresponding web requests. The problem lies in when the underlying _domain objects_ throw `IllegalArgumentException`'s which do not map to illegal arguments of the service requests.

For that reason it would make sense to have domain specific `IllegalArgumentException`s for application services, so that we can uniformly map domain-specific exceptions to web errors. In fact, we talked about this before in the past. I will add this to the `tech-meeting` backlog and see what previous information I can gather.

Whathecode: https://app.zenhub.com/workspaces/mobile-tech-meetings-5fcf567a99061f00110291ae/issues/imotions/mobile-project-management/26

PR #220, opened 2021-01-19
`EventBus` abstraction and groundwork for `StudyService.remove`

Whathecode:
`StudyService.remove` currently only removes the study from `StudyRepository`. This change needs to be propagated to `ParticipantService` and subsequently the deployments subsystem.

Therefore, I started adding the `EventBus` abstraction, which is the main bulk of this PR I already talked about extensively with @ltj . @ltj ; you could have a quick look at the first commit, which is the only one you did not yet look at. :)


ltj (commented):
Looked at the study service part and looks a-ok. Nice work on the remaining event bus stuff. Now we just need to figure out how decouple the events from the actual application services without sacrificing cohesion too much :)

PR #221, opened 2021-01-25
Propagate study removal to `ParticipantService`.

Whathecode:
`StudyService` and `ParticipantService` now only have indirect dependencies via an `EventBus` implementation.

Whenever a `Study` gets created, the `ParticipantService` creates a corresponding `Recruitment` object, and when the `Study` goes live, deployment for that recruitment can commence. This was the only dependency across these two services.

When a `Study` gets removed, the `Recruitment` also needs to be removed, which is the main reason for this PR. The continued decoupling is refactoring, although it does have a big impact on the repo implementations. `Participations` are now stored in `Recruitment`, and no longer in `Study`. And, `Participants` are now also stored in `Recruitment`, as opposed to directly in the repository.

**Out of scope for this PR:**

Study removal still needs to be propagated further to `DeploymentService` and `ParticipationService` in the deployment subsystem. But, I think a similar decoupling as done for `StudyService/ParticipantService` needs to happen here, and I would thus like to do that as a separate PR.

Furthermore, I think it would be beneficial to already release these changes, even though `StudyService.remove` is not complete yet (and also not documented in README) so that we can validate the current `EventBus` abstraction prior to applying it to more services.

**Considerations:**

- Now that `Participant` has been moved inside `Recruitment`, should we consider renaming `ParticipantRepository` to `RecruitmentRepository`? This might be premature, given that we do intend to add extra functionality such as storing participant data which might impact the repo.

- Some future use cases to consider: we expect to store participant study data in `Participant` in the future which can then help to determine recruitment goals. There are thus invariants to be handled by `Recruitment`. But, we might not want to load _all_ participant data whenever we retrieve `Recruitment`. But, that seems like a potential optimization for later. Right now, we don't even have participant data in study yet, and it is only stored (and can only be added) in deployment.


carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/ParticipantServiceHost.kt
```diff
@@ -94,9 +119,9 @@ class ParticipantServiceHost(
         require( group.isNotEmpty() ) { "No participants to deploy specified." }
 
         // Verify whether the study is ready for deployment.
-        val study: Study = getStudyOrThrow( studyId )
-        check( study.canDeployToParticipants ) { "Study is not yet ready to be deployed to participants." }
-        val protocolSnapshot = study.protocolSnapshot!!
+        val recruitment = getRecruitmentOrThrow( studyId )
+        check( recruitment.canAddParticipations ) { "Study is not yet ready to be deployed to participants." }
```
Whathecode, line 123: For consistency, this should maybe be renamed to `recruitment.isReadyForDeployment`? Essentially, it is after calling `recruitment.readyForDeployment` this returns true.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/Recruitment.kt
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+
+    companion object
+    {
+        fun fromSnapshot( snapshot: RecruitmentSnapshot ): Recruitment
+        {
+            val recruitment = Recruitment( snapshot.studyId )
+            recruitment.creationDate = snapshot.creationDate
+            if ( snapshot.studyProtocol != null && snapshot.invitation != null )
```
Whathecode, line 31: This can probably just be a `getter` on snapshot for readability.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/Recruitment.kt
```diff
@@ -0,0 +1,109 @@
...
+import dk.cachet.carp.deployment.domain.users.StudyInvitation
+import dk.cachet.carp.protocols.domain.StudyProtocolSnapshot
+import dk.cachet.carp.studies.domain.users.DeanonymizedParticipation
+import dk.cachet.carp.studies.domain.users.Participant
+
+
+/**
+ * Represents a set of participants recruited for a [Study] identified by [studyId].
+ */
+class Recruitment( val studyId: UUID ) :
```
Whathecode, line 16: Do we want to also store `Participant`s, or at least their identities, in `Recruitment`?

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/Recruitment.kt
```diff
@@ -0,0 +1,109 @@
...
+     * The snapshot of the study protocol to which participants in this recruitment can be invited.
+     */
+    var studyProtocol: StudyProtocolSnapshot? = null
+        private set
+
+    /**
+     * The invitation to send to invited participants.
+     */
+    var invitation: StudyInvitation? = null
+        private set
```
Whathecode, line 55: This data is now duplicated from the `StudyService`. I don't think that is a problem. Within `StudyService` it is mutable, until a `Recruitment` is created within the scope of which it is immutable.

Likely, a similar relation can be made later once we implement the 'pilot' phase in `Study`. During this time, these could be modified, but each modification will likely result in a new corresponding `Recruitment` object. I think this confirms the relevance of having split these two.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/Recruitment.kt
Whathecode: No. This would duplicate this logic to two places. Better keep it decoupled.

Whathecode:
Looking more at this, I think I can easily refactor this some more to move some of the current `ParticipantServiceHost` logic as value objects inside of `Recruitment`. In particular, the logic which extracts master device role names from the protocol and verifies all of them are assigned in `AssignParticipantDevices`.

All of this is a very good sign. It means that decoupling `ParticipantService` from `StudyService` was a good separation of concerns which now allows pushing constraint logic into domain models where it belongs, rather than in application services.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/Recruitment.kt
Whathecode: Looking more at this, I think I can easily refactor this some more to move some of the current ParticipantServiceHost logic as value objects inside of Recruitment. In particular, the logic which extracts master device role names from the protocol and verifies all of them are assigned in AssignParticipantDevices.

All of this is a very good sign. It means that decoupling ParticipantService from StudyService was a good separation of concerns which now allows pushing constraint logic into domain models where it belongs, rather than in application services.

ltj (dismissed):
Nice work @Whathecode I left some comments and looking forward to see how this plays out in concrete implementations.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/ddd/ApplicationServiceEventBus.kt
```diff
@@ -20,22 +20,22 @@ class ApplicationServiceEventBus<
     /**
      * Subscribe to events of [eventType] belonging to [applicationServiceKlass] and handle them using [handler].
      */
-    suspend fun <
+    fun <
```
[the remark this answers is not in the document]
Whathecode: I tried making a bit more legible by using shorter generic names. But, the main reason for adding the distinct generic types on separate lines is because they use nested generics.

I could shorten the generic types even futher, but overall, even if that allows me to place it all on one line, I personally don't find `<OS : ApplicationService<OS, OSE>, OSE: IntegrationEvent<OS>>` more legible.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/ParticipantServiceHost.kt
```diff
@@ -4,29 +4,52 @@ import dk.cachet.carp.common.EmailAddress
...
 import dk.cachet.carp.studies.domain.users.participantIds
 
 
 class ParticipantServiceHost(
-    private val studyRepository: StudyRepository,
     private val participantRepository: ParticipantRepository,
     private val deploymentService: DeploymentService,
-    private val participationService: ParticipationService
+    private val participationService: ParticipationService,
+    private val eventBus: ApplicationServiceEventBus<ParticipantService, ParticipantService.Event>
```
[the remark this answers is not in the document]
Whathecode: The generic type constrains what _can_ be published. The _subscribe_ method infers from the type what is _being_ published. 

We can certainly infer what is being published from the type: that is what happens in `EventBus`. But, I don't see any other way at compile time of constraining what _can_ be published. That information needs to be provided somewhere through the type system.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentSnapshot.kt
```diff
@@ -0,0 +1,54 @@
...
+import dk.cachet.carp.common.DateTime
+import dk.cachet.carp.common.UUID
+import dk.cachet.carp.common.ddd.Snapshot
+import dk.cachet.carp.deployment.domain.users.StudyInvitation
+import dk.cachet.carp.protocols.domain.StudyProtocolSnapshot
+import kotlinx.serialization.Serializable
+
+
+@Serializable
+data class RecruitmentSnapshot(
```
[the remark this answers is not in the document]
Whathecode: Yeah, would be cool. I'm getting annoyed writing and updating them.

PR #223, opened 2021-01-27
Fix Detekt task classpath

cortinico:
This is a sample PR to verify a correct configuration of the Detekt classpath for Type Resolution (and to avoid false positives on `UnnecessarySafeCall`).

Feel free to disregard and incorporate the changes in a different way.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -326,12 +326,14 @@ configure( rootProject )
         source = fileTree("$rootDir")
         {
             include('**/src/**')
-            exclude('**/node_modules/**')
+            exclude('**/node_modules/**', '**/resources/**')
         }
         config.from("$rootDir/detekt.yml")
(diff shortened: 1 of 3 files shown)


cortinico:
See https://github.com/detekt/detekt/issues/3414 for reference

Whathecode:
Thank you for all the information! Extremely informative and at a glance seems like it will fix the issues. I will definitely incorporate these changes later!

Whathecode (approved):

Whathecode:
@cortinico Thank you again! I simply applied your changes in the main build file so that they work for all subprojects.

Looking forward to the next release so that I can remove the suppressed warnings for the `expect object` scenario (https://github.com/detekt/detekt/issues/3415). ;p

cortinico:
Awesome 🙏

PR #227, opened 2021-02-09
Handle participations in `ParticipantGroup` rather than `StudyDeployment`.

Whathecode:
Similar to [a previous PR](https://github.com/cph-cachet/carp.core-kotlin/pull/221) which split repository dependencies between `StudyService` and `ParticipantService`, this PR removes dependencies between `DeploymentService` and `ParticipationService`. They now only have indirect dependencies via `EventBus`. This means that `Participation` is now stored in `ParticipantGroup` rather than in `StudyDeployment`.

Whenever a `StudyDeployment` is created, a matching `ParticipantGroup` gets created. `StudyDeployment` does device management (deployment and in the future status management), whereas `ParticipantGroup` manages everything related to participants, such as assigning devices to participants, consent, and storing participant data. This seems like a reasonable separation of concerns which at the very least makes the code easier to comprehend. Whether or not `DeploymentService` and `ParticipationService` need to evolve into separate microservices is to be determined later.

This separation might feel superfluous given that `ParticipationService` listens to a lot of changes of `DeploymentService` (register/unregister device, deployment stopped) to allow removing the dependency to `DeploymentRepository`, but we expect that later additions to the two respective services are less likely to impact one another. Furthermore, these are events which in practice won't be raised often.

There is one semi-unrelated commit: https://github.com/cph-cachet/carp.core-kotlin/pull/227/commits/0b48f2631e49bf69ff58074fc0418f9761fd06e2 I noticed the `InMemoryParticipationRepository` was not tested yet; only the test for the interface were written. Apparently, some of the tests were wrong, which are now fixed.

Potential improvements:

- [ ] Should we pass `DeviceDescriptor`s for `getActiveParticipationInvitations`? I see no real reason why only the role is passed.
- [ ] Now that `ParticipantGroup` contains assigned master devices, we could also consider storing which devices are assigned to participants within the same data structure. This would need to be updated as part of `addParticipation`.

### DIFF
--- a/carp.client.core/src/commonTest/kotlin/dk/cachet/carp/client/ClientCodeSamples.kt
+++ b/carp.client.core/src/commonTest/kotlin/dk/cachet/carp/client/ClientCodeSamples.kt
@@ -7,6 +7,8 @@ import dk.cachet.carp.client.domain.data.DataListener
 import dk.cachet.carp.client.infrastructure.InMemoryClientRepository
 import dk.cachet.carp.common.UUID
 import dk.cachet.carp.common.data.CarpDataTypes
+import dk.cachet.carp.common.ddd.SingleThreadedEventBus
+import dk.cachet.carp.common.ddd.createApplicationServiceAdapter
 import dk.cachet.carp.common.users.Account
 import dk.cachet.carp.common.users.AccountIdentity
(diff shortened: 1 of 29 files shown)


ltj (approved):
Whew. Very nice @Whathecode

PR #229, opened 2021-02-12
Add TS declarations for deployment subsystem requests.

Whathecode:
This includes all TS declarations needed to initialize `DeploymentService` and `ParticipationService` requests, serialize them, and deserialize the responses of those requests.

Look at the `carp.core` codebase to see which types are returned by which requests. You can find TS declarations in the corresponding namespaces. To serialize/deserialize, look at the examples of the other requests. But, now use the `createDeploymentSerializer`. When you need a `Map` or `Set` of the objects to be serialized, use the List/Map/Set serializer constructor (see [example in existing unit tests](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/typescript-declarations/tests/carp.studies.core.ts#L130)).

I did use `any` here and there in type declarations where I did not think you would need to access the elements within it. `any` means you can still access them dynamically just like you would in Javascript. If we ever need type safety for those, I can add more type declarations. But, this should get you going!

Note that these are the declarations for the current `develop` branch, which is not backwards compatible with alpha 28. Next, I will create a hotfix for alpha 28 and modify these declarations where needed (not too much discrepancy).

### DIFF
--- a/typescript-declarations/@types/carp.core-kotlin-carp.deployment.core/index.d.ts
+++ b/typescript-declarations/@types/carp.core-kotlin-carp.deployment.core/index.d.ts
@@ -2,10 +2,18 @@ declare module 'carp.core-kotlin-carp.deployment.core'
 {
     import { kotlin } from 'kotlin'
     import ArrayList = kotlin.collections.ArrayList
+    import HashMap = kotlin.collections.HashMap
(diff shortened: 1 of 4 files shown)


ossi0004 (approved):

ossi0004:
Reviewed and tested

Issue #161, opened 2020-09-21
Check out new default polymorphic type handler in `kotlinx.serialization`


Whathecode:
In order to handle unknown types we use custom `UnknownPolymorphicSerializer` in this library.

It seems similar functionality is now supported out-of-the-box in `kotlinx.serialization`: https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/polymorphism.md#default-polymorphic-type-handler-for-deserialization

We should verify whether or not this provides all the functionality we currently implement ourselves, and replace our implementation with `kotlinx.serialization` in case it does.

Whathecode:
If this would work, it is likely we can support other formats than JSON as well: https://github.com/cph-cachet/carp.core-kotlin/issues/33

Issue #9, opened 2018-06-27
Prevent deriving types from Measure having to specify DataTypeSerializer


Whathecode:
Currently, since `DataType` is polymorph, we use `DataTypeSerializer` in order to support loading any extending types unknown at runtime by wrapping them in `CustomDataType` upon deserialization. 

To make this work, all types extending from `Measure` need to specify to use the `DataTypeSerializer`:
```
@Serializable
data class StubMeasure(
    @Serializable( with = DataTypeSerializer::class )
    override val type: DataType = StubDataType() ) : Measure()
```

At the time of writing, [there seems no support to specify this serializer on the base property of the abstract `Measure` class](https://github.com/Kotlin/kotlinx.serialization/issues/162).

Can we deal with this in a nicer way?

Whathecode:
I believe this is only needed for deserialization, thus this is less important for extending types in external libraries outside of our control (serialization will still work correctly).

Concretely, this thus only impacts 'server-side'. Each time we introduce 'default' supported tasks and measures, we need to remember to apply these attributes correctly. In case there is no way to reuse the serializer defined on the base type, we could consider implementing a runtime check which triggers an exception in case the attribute is not defined.

Whathecode:
In Kotlin 1.3 the equivalent of this will become a `Polymorph` attribute which _needs to be applied_ for polymorph serialization to work.

Having looked more into this, I now believe this will most likely remain a requirement. The question then becomes, how can we enforce correct implementations of deriving types. We could check in a static initializers of such types whether the attribute is applied correctly, and throw a suitable runtime error in case it is not present.

PR #232, opened 2021-02-18
Register `UnknownPolymorphicSerializer`s as `default` in `SerializersModule`.

Whathecode:
By relying on the `default` fallback mechanism of polymorphic serialization to register the `UnknownPolymorphicSerializer`, it is no longer needed to specify the unknown serializers anywhere unknown types can be encountered in the codebase.

To make this work, the wrapper objects used to store unknown types at runtime also need to be registered as `subclass` for the polymorphic types. I updated the documentation accordingly.

In addition, this means that by registering a different `default` serializer (and `subclass` for the wrapper), unknown polymorphic serialization can now also be supported for different serialization formats than Json.

Making this work was a bit of a hack which required accessing `kotlinx.serialization` internals. An `Encoder` maintains state that the current object is being encoded as polymorphic or not. Depending on this state, `beginStructure` outputs type information or not. But, given that the wrapped object already contains type information, we never want to output this. But whether or not the `Encoder` is currently writing something as polymorphic is handled outside of the serializer. I wrote a tiny multiplatform accessor to modify this internal boolean state prior to encoding the wrapped json object. I will contact the Jetbrains folks to see whether or not they could consider this as a valid use case to open up their APIs a bit.

This fixes the following issues:

- [Make use of the new `default` feature of polymorphic `kotlinx.serialization`](https://github.com/cph-cachet/carp.core-kotlin/issues/161)
- [Prevent deriving types from types which allow unknown extending classes having to specify custom serializers wherever they are used](https://github.com/cph-cachet/carp.core-kotlin/issues/9) (e.g., `Measure`, and `AnyDeviceDescriptor`).

### DIFF
--- a/carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntimeSnapshot.kt
+++ b/carp.client.core/src/commonMain/kotlin/dk/cachet/carp/client/domain/StudyRuntimeSnapshot.kt
@@ -6,7 +6,6 @@ import dk.cachet.carp.common.ddd.Snapshot
 import dk.cachet.carp.deployment.domain.MasterDeviceDeployment
 import dk.cachet.carp.protocols.domain.devices.AnyDeviceDescriptor
 import dk.cachet.carp.protocols.domain.devices.AnyMasterDeviceDescriptor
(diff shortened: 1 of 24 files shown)


ltj (approved):
Very cool @Whathecode . It's coming together as much more coherent solution now.

PR #246, opened 2021-04-22
Stabilize tasks/measures/sampling types

Whathecode:
This is the groundwork for most of the changes we agreed on in terms of tasks, measures, and sampling configuration/schemes.

There is still a need for additional fields and concrete types, but some of these require further discussion (such as `InteractiveTask`). Instead, this PR focuses on changing the current API to the point where only _additions_ should be needed now, with the exception of battery-aware sampling schemes. The latter still requires more thinking and I will dedicate a separate PR to this. Likewise, we can talk about the new needed additions in subsequent PRs.

The commits and linked issues mostly speak for themselves, but here is a short overview.

`DeviceDescriptor`:

- `SamplingScheme`s are no longer strongly-linked to `DataType`. Instead, `DeviceDescriptor`s decide on the `DataTypeSamplingScheme` for each of the `DataType`s they support.
- Implementations of `DeviceDescriptor` need to list `Sensors` (`DataTypeSamplingSchemeList`) and `Tasks` (`TaskDescriptorList`). This is statically available information on device capabilities which _is not serialized_ since it is not configurable by the researcher defining the study protocol.

Tasks / measures:

- Specializations of `Measure` are removed and it is now a concrete class.
- `Measure`s are initialized through `DataTypeSamplingScheme` and no longer through separate functions accessible from the `DeviceDescriptor`. In addition, the API is extended to support checking whether or not a constructed `Measure` is _valid_ for the related `DataTypeSamplingScheme`. E.g., to check whether or not a set interval exceeds a minimum/maximum value (this is not implemented yet, but the location where to do so should now be clear).
- Likewise, `TaskDescriptor`s can be constructed through the `DeviceDescriptor`'s `TaskDescriptorList`.
- `ConcurrentTask` is renamed to `BackgroundTask`, and the documentation is updated to reflect what it represents.

Triggers:

- Triggers can now also _stop_ tasks.
- `TriggeredTask` is renamed to `TaskControl`. The previous name did not work since it now defines whether a task needs to start or stop.
- Extension functions are added to `Trigger` to create related `TaskControl` objects. E.g.: `start` in `phone.atStartOfStudy().start( measures, phone )` is an extension function on `Trigger`. Extension functions are used rather than an actual function on `Trigger` in order to maintain the distinction between the `application` and `domain` namespace. The created `TaskControl` is the one used in the domain model.


carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/sampling/SamplingConfiguration.kt
```diff
@@ -23,12 +23,27 @@ interface SamplingConfiguration
  * as part of setting up a [DeviceDescriptor].
 */
 @DeviceDescriptorBuilderDsl
-interface SamplingConfigurationBuilder
+interface SamplingConfigurationBuilder<TSamplingConfiguration : SamplingConfiguration>
```
ltj, line 26: I would like generic types to be one capital letter as they are just a placeholder. The very verbose type labels actually make it less legible

ltj (approved):
👌

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/sampling/SamplingConfiguration.kt
Whathecode: Mmmm ... agree to disagree. :) If there are generic type constraints, I do prefer to keep something akin to an interface name in code. Especially when there are multiple generics (not the case in this class, though, but there is a `TBuilder` lower in the file).

ltj: `TConfig` then

Whathecode: Resolved in next PR (which targets this PR).

bardram:
> There is still a need for additional fields and concrete types, but some of these require further discussion (such as `InteractiveTask`). Instead, this PR focuses on changing the current API to the point where only _additions_ should be needed now, with the exception of battery-aware sampling schemes. The latter still requires more thinking and I will dedicate a separate PR to this. Likewise, we can talk about the new needed additions in subsequent PRs.

I agree to this approach.

> `DeviceDescriptor`:
> 
> * `SamplingScheme`s are no longer strongly-linked to `DataType`. Instead, `DeviceDescriptor`s decide on the `DataTypeSamplingScheme` for each of the `DataType`s they support.
> * Implementations of `DeviceDescriptor` need to list `Sensors` (`DataTypeSamplingSchemeList`) and `Tasks` (`TaskDescriptorList`). This is statically available information on device capabilities which _is not serialized_ since it is not configurable by the researcher defining the study protocol.
> 

I simply do not understand the whole `DeviceDescriptor` model, so I really cannot judge this.

> Tasks / measures:
> 
> * Specializations of `Measure` are removed and it is now a concrete class.
> * `Measure`s are initialized through `DataTypeSamplingScheme` and no longer through separate functions accessible from the `DeviceDescriptor`. In addition, the API is extended to support checking whether or not a constructed `Measure` is _valid_ for the related `DataTypeSamplingScheme`. E.g., to check whether or not a set interval exceeds a minimum/maximum value (this is not implemented yet, but the location where to do so should now be clear).
> * Likewise, `TaskDescriptor`s can be constructed through the `DeviceDescriptor`'s `TaskDescriptorList`.

Again - I'm afraid that I really can't figure out the details of this and what the implications are.....

> * `ConcurrentTask` is renamed to `BackgroundTask`, and the documentation is updated to reflect what it represents.
> 

That makes sense.


> Triggers:
> 
> * Triggers can now also _stop_ tasks.
> * `TriggeredTask` is renamed to `TaskControl`. The previous name did not work since it now defines whether a task needs to start or stop.

Well - I work with four life cycles of execution of "everything" including triggers:

* initialized
* resumed
* pause
* stopped

There is a state machine controlling this - e.g. a stopped "thing" can't be resumed.

So - I have a `PeriodicTrigger` which resumes at a certain frequency and pauses after a certain time period.


> * Extension functions are added to `Trigger` to create related `TaskControl` objects. E.g.: `start` in `phone.atStartOfStudy().start( measures, phone )` is an extension function on `Trigger`. Extension functions are used rather than an actual function on `Trigger` in order to maintain the distinction between the `application` and `domain` namespace. The created `TaskControl` is the one used in the domain model.

hm...

I don't know -- I think the whole domain model is becoming extremely bloated... It's very very complex. I wanted to make support for specifying a very simple protocol, having a set of triggers, tasks, and measures... Four classes - that's it ;-)

I'm becoming more and more uncertain as to whether the "big" carp_core domain model makes sense to have on a phone...

We might need to talk.

Whathecode:
> four life cycles of execution of "everything" including triggers

These seem to be related to the _runtime_ (a state machine necessarily is), not so much the _description_ of the study. From the perspective of a researcher describing a research protocol, what is the use in disambiguating between "pausing" and "stopping" a task?

There might be a useful distinction, but I don't know of any use cases.

> I have a PeriodicTrigger which resumes at a certain frequency and pauses after a certain time period.

I wonder what the use case of that is. But, in Core the way to set this up would be using  [`ScheduledTrigger`](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/triggers/ScheduledTrigger.kt, which uses the [RRULE standard](https://tools.ietf.org/html/rfc5545#section-3.3.10) to specify a recurring schedule (your period, but with more capabilities, e.g., every Monday). The `duration` is defined on a `BackgroundTask`, since `duration` only really makes sense for background sensing. A `ScheduledTrigger` which triggers a `BackgroundTask` is thus equivalent to your `PeriodicTrigger` if I understand it correctly.

Combining them in one raises questions such as:

- What does it mean to use a `PeriodicTrigger` to trigger a survey with a specific duration?
- What is the difference with `PeriodicMeasure`? (I think `PeriodicMeasure` is what is now a `Measure` with an `IntervalSamplingConfiguration` in Core.)
- Why is there a need for `PeriodicMeasure`, `DelayedTrigger`, `RecurrentScheduleTrigger`, `PeriodicTrigger`, `CronScheduledTrigger`? This seems like unnecessary complexity, and many different ways of doing the same thing.

> I don't know -- I think the whole domain model is becoming extremely bloated.

The functions you commented on are simply Kotlin APIs to construct domain objects. They do not change anything about the domain objects themselves. All that got added was a `control` parameter to the previously named `TriggeredTask`, since, it is a valid use case that you want to _stop_ measuring something based on a condition (a `Trigger`).

> I'm becoming more and more uncertain as to whether the "big" carp_core domain model makes sense to have on a phone...

Core is definitely overly complex if _all_ you want to do is collect data on a single smartphone. Core already keeps in mind multiple devices, and thus triggers and tasks _across_ devices. We can definitely consider creating a subset/easier protocol definition which only allows collecting data on a smartphone. It does beg the question how _connected_ devices (as currently also modeled in Core) fits into that picture.

PR #249, opened 2021-05-05
Battery-aware sampling

Whathecode:
I decided to implement battery-aware sampling configurations as a specialization of `DataTypeSamplingScheme` (it is thus its own 'type' of configuration), as opposed to specifying multiple default sampling configurations for all data types (for different battery levels) on a device as we originally considered.

The reasoning is:

- Not all data streams can be made meaningfully battery-aware. And, not all devices even care about optimizing battery life.
- It resolves the question which sampling configuration wins out when a custom configuration is passed for `Measure`. E.g., suppose the battery is low, thus a 'low' sampling configuration is used for geolocation. But, now a `Measure` is started which specifies a custom configuration; which configuration should be used, the 'low' one or the custom one? This choice is now delegated to the researcher configuring the protocol. Since the expected configuration in this case is battery-aware, so is the configuration which needs to be passed to `Measure`. The researcher can choose to only set one battery-level, or all in case a full override is desired.

After some consideration, I opted for adding both a non-battery-aware `GranularitySampling` and `AdaptiveGranularitySampling` which is battery-aware. The codebase is more consistent this way, and the API for developers is clearer.

Whether or not the intermediate `BatteryAwareSampling` abstraction will be reused in other specializations remains to be seen, but conceptually it seems to make sense. With minimal amount of code any sampling scheme can now be made 'battery-aware'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ val phone = Smartphone( "Patient's phone" )
 {
     // Configure device-specific options, e.g., frequency to collect data at.
     defaultSamplingConfiguration {
-        geolocation { interval = TimeSpan.fromMinutes( 15.0 ) }
+        geolocation { granularity = Granularity.Normal }
     }
 }
 protocol.addMasterDevice( phone )
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ val phone = Smartphone( "Patient's phone" )
 {
     // Configure device-specific options, e.g., frequency to collect data at.
     defaultSamplingConfiguration {
-        geolocation { granularity = Granularity.Normal }
+        geolocation { batteryNormal { granularity = Granularity.Balanced } }
     }
 }
 protocol.addMasterDevice( phone )
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/sampling/AdaptiveGranularitySampling.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/sampling/AdaptiveGranularitySampling.kt
@@ -1,0 +1,33 @@
+@file:Suppress( "MatchingDeclarationName" )
+
+package dk.cachet.carp.common.application.sampling
+
+import dk.cachet.carp.common.application.data.DataType
+import dk.cachet.carp.common.application.devices.DeviceDescriptor
+
+
+/**
+ * Sampling scheme which provides only indirect control over how data is sampled by specifying a desired level of [Granularity].
+ * The levels of granularity correspond to expected degrees of power consumption.
+ * By default, [Granularity.Balanced] is used; when the battery is low, [Granularity.Coarse]; when battery is critically low, sampling stops.
+ */
+class AdaptiveGranularitySamplingScheme( dataType: DataType ) :
+    BatteryAwareSamplingScheme<GranularitySamplingConfiguration, GranularitySamplingConfigurationBuilder>(
+        dataType,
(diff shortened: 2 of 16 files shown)


README.md
```diff
@@ -159,7 +159,7 @@ val phone = Smartphone( "Patient's phone" )
 {
     // Configure device-specific options, e.g., frequency to collect data at.
     defaultSamplingConfiguration {
-        geolocation { interval = TimeSpan.fromMinutes( 15.0 ) }
+        geolocation { batteryNormal { granularity = Granularity.Balanced } }
```
ltj, line 162: 🥇

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/sampling/AdaptiveGranularitySampling.kt
```diff
@@ -0,0 +1,33 @@
...
+import dk.cachet.carp.common.application.data.DataType
+import dk.cachet.carp.common.application.devices.DeviceDescriptor
+
+
+/**
+ * Sampling scheme which provides only indirect control over how data is sampled by specifying a desired level of [Granularity].
+ * The levels of granularity correspond to expected degrees of power consumption.
+ * By default, [Granularity.Balanced] is used; when the battery is low, [Granularity.Coarse]; when battery is critically low, sampling stops.
+ */
+class AdaptiveGranularitySamplingScheme( dataType: DataType ) :
```
ltj, line 14: I imagine an adaptive sampling scheme could adapt to something other than battery level? Just saying that this might be too generic a name for something that might have several specializations or be even more generalized in the future.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/sampling/DataTypeSamplingScheme.kt
```diff
@@ -8,7 +8,7 @@ import dk.cachet.carp.common.application.tasks.Measure
 /**
  * Specifies the sampling scheme for a [DataType], including possible options, defaults, and constraints.
  */
-abstract class DataTypeSamplingScheme<TSamplingConfigurationBuilder : SamplingConfigurationBuilder<*>>(
+abstract class DataTypeSamplingScheme<TConfigBuilder : SamplingConfigurationBuilder<*>>(
```
ltj, line 11: 🙏

ltj (approved):
I like the way this turned out. Nice job. Only a few comments/suggestions

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/sampling/AdaptiveGranularitySampling.kt
Whathecode: Mmmm. I imagine other conditions would likely be modeled as `Trigger`s which start a `Measure` which overrides the `SamplingConfiguration`.

Regardless, I also thought this might be too generic when I wrote it, but found the longer alternative too convoluted. I.e., `BatteryAwareGranularitySamplingScheme`. Or maybe not? 🤷 

Suggestions?

ltj: Maybe just `BatteryAwareSamplingScheme`. I think "aware" implies its adaptiveness. The granularity part sounds a bit convoluted :)

Issue #248, opened 2021-05-05
Event bus race condition in ParticipantService


ltj:
https://github.com/cph-cachet/carp.core-kotlin/blob/2990766f5e24da497c10029a6fe04c63be461f13/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/ParticipantServiceHost.kt#L139

In `ParticipantServiceHost.deployParticipantGroup` a study deployment is created via a client call. On completion the deployment service will emit an event to be consumed by the `ParticipationService` in order to create the `ParticipantGroup`. The following calls to add participations relies on the presence of the participant group and will race the aforementioned creation of this - this will like always fail since it's multiple event bus roundtrips vs a single client call (although HTTP).

Whathecode:
For this particular race condition, I guess the solution is to include the participations to the `createStudyDeployment` request and handling the addition of participations in the `StudyDeploymentCreated` event handler of `ParticipationServiceHost`.

But, the `Recruitment` stored by `ParticipantServiceHost` stores `DeanonymizedParticipation`s which need the `participationId` returned by `ParticipationService`.

However, I am now considering that the _deanonymization_ step should be handled by the `ParticipationService` either way. We did consider before that we should have a clear separation of anonymized and deanonymized endpoints. Right now, the `ParticipantServiceHost` endpoints all return `ParticipantGroupStatus` which contains the `participantId` to `participationId` mapping.

Maybe this is wrong. Should we make it the full responsibility of `ParticipationService` to do this mapping? The `participantId` would then be passed to this service. We can then add UAC on a new 'deanonymization' endpoint of `ParticipationService`. You could also choose to only deanonymize a subset of participants.

Also, while looking at this code I noticed potential (but unlikely) race conditions upon removal of study deployments related to retrieving participant data from `ParticipationService`. The typical way of handling these in DDD would be to [_copy_ the data based on incoming events from this service instead](https://enterprisecraftsmanship.com/posts/how-to-request-information-from-multiple-microservices/). However, an exception handler which infers that the deployment was deleted and propagates this is also a not-too-complicated possibility, imho.

PR #250, opened 2021-05-07
Make deanonymization process explicit in deployments subsystem

Whathecode:
Before, any endpoint in `RecruitmentService` which returned `ParticipantGroupStatus` leaked anonymous participation IDs since it returned `DeanonymizedParticipation` objects. Instead, deanonymization is now done through an explicit separate request: `ParticipationService.deanonymizeParticipations`.

This is a better decoupling of participation IDs, which are managed by `ParticipationService`, but were previously also stored in `RecruitmentService`. This caused temporal coupling of calls from `RecruitmentService` to `ParticipationService` and a subsequent race condition (#248) which is fixed as part of this PR.

To resolve this, `DeploymentService.createStudyDeployment` now also takes `ParticipantInvitation`s; adding participations is no longer managed through separate requests in `ParticipationService`. As a result, the `addParticipation` endpoint is now only called internally from within `ParticipationService` based on incoming events. We should consider removing this endpoint: https://github.com/cph-cachet/carp.core-kotlin/issues/251


carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/domain/users/AccountParticipation.kt
```diff
@@ -8,11 +8,13 @@ import kotlinx.serialization.Serializable
 
 
 /**
- * An account (identified by [accountId]) which was invited to a study deployment (using [invitation])
+ * A participant, uniquely identified by [externalParticipantId], linked to an account identified by [accountId],
```
ltj, line 11: Aren't we missing a type representing this sans account?

PR #253, opened 2021-05-12
Remove `ParticipationService.addParticipation` endpoint

Whathecode:
Inviting participants now _has to_ be done through `DeploymentService.createStudyDeployment`; it is no longer optional. Correspondingly, `ParticipationService.addParticipation` has been removed.

We envision that any intermediate 'participant group' management will be implemented in the studies subsystem once we need it. So far, this wasn't used; deploying a participant group simply translated to multiple `addParticipation` calls.

This refactoring also allowed to centralize the logic to validate whether a participant group is valid for a protocol in as an extension method in `ParticipantInvitation`: `StudyProtocolSnapshot.throwIfInvalid( invitations )`. By doing so, `createStudyDeployment` can now also easily validate invitations, which was missing before (a bug).


ltj (dismissed):
Cool stuff

carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/application/DeploymentService.kt
```diff
@@ -40,17 +40,23 @@ interface DeploymentService : ApplicationService<DeploymentService, DeploymentSe
...
-     *  - any of the master devices a participant in [invitations] is invited to is not part of [protocol]
+     *  - [invitations] is empty
+     *  - any of the assigned device roles in [invitations] is not part of the study [protocol]
+     *  - not all master devices part of the study [protocol] have been assigned a participant
      * @return The [StudyDeploymentStatus] of the newly created study deployment.
      */
     suspend fun createStudyDeployment(
         protocol: StudyProtocolSnapshot,
-        invitations: List<ParticipantInvitation> = emptyList()
+        invitations: List<ParticipantInvitation>
```
Whathecode, line 59: Since participants are now instantly invited as part of the `createStudyDeployment` call, we should also pass optional `DeviceRegistration` for the assigned devices. Otherwise "device preregistration" is supported in a weird way. Preregistering a device would require creating a deployment, and then quickly adding a registration for that deployment, before the client has a chance of retrieving the master device deployment.

Logically, it makes more sense the _pre_registration is known already when the deployment is created.

PR #254, opened 2021-05-12
Let caller of `createStudyDeployment` determine ID.

Whathecode:
This is more consistent with other functions which create new aggregate root objects. The identity should be determined by the caller, or by the infrastructure.

Trying to create an AR with an identity which already exists should always fail. Alternatively, we can return the previously created object, but is this really clearer/more useful to the caller?


ltj (commented):
👍

PR #265, opened 2021-05-27
Add `WebTask`

Whathecode:


thaiphandinh (commented):
It's all good. I have only one concern that do we take this format as the input for survey's url? 
`val url =            "$surveyUrl?" +            "participantId=${WebTask.UrlVariable.PARTICIPANT_ID.pattern}" +            "&amp;deploymentId=${WebTask.UrlVariable.DEPLOYMENT_ID.pattern}" +            "&amp;triggerId=${WebTask.UrlVariable.TRIGGER_ID.pattern}"`

IMO, could it be easier/ more convenience if we can do automatic adding url params instead of typing it manually. (Please correct me if I'm getting you wrong here)

Whathecode:
> It's all good. I have only one concern that do we take this format as the input for survey's url?
> `val url = "$surveyUrl?" + "participantId=${WebTask.UrlVariable.PARTICIPANT_ID.pattern}" + "&amp;deploymentId=${WebTask.UrlVariable.DEPLOYMENT_ID.pattern}" + "&amp;triggerId=${WebTask.UrlVariable.TRIGGER_ID.pattern}"`
> 
> IMO, could it be easier/ more convenience if we can do automatic adding url params instead of typing it manually. (Please correct me if I'm getting you wrong here)

This is just an example of what it could look like in the unit test. It is fully the website we link to which determines whether, and how, query parameters are accepted. We cannot automate anything here, unless we know the specific website we link to. That is not the case for `WebTask`, which is an abstraction that links to _any_ website.

PR #275, opened 2021-06-17
fix: DataTypeMetaData now gets added to the underlying list

ltj:
The `add` method in `DataTypeMetaDataList` shaded the `add` method in `EnumObjectList` and the former never ensured entries were actually added to the underlying list. Renamed method in `DataTypeMetaDataList` to `addMetaData` and make it call the parent `add` method.

### DIFF
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/data/DataTypeMetaDataList.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/data/DataTypeMetaDataList.kt
@@ -16,5 +16,5 @@ open class DataTypeMetaDataList : EnumObjectList<DataTypeMetaData>()
      * which can be displayed to the user using [displayName] and should be stored temporally as specified by [timeType].
      */
     fun add( fullyQualifiedName: String, displayName: String, timeType: DataTimeType ): DataTypeMetaData =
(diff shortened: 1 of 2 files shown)


Whathecode:
My god, that's a stupid bug. :O Thank you for reporting. Odd it didn't fail any tests where this was used (I guess it isn't). I remember thinking, "I don't have to add a unit test for this list since the base `EnumObjectList` is already tested" 🤦 .

But, isn't a better solution here to redirect to `super.add()`. Basically just appending: `.also { add( it ) }`?

ltj:
> My god, that's a stupid bug. :O Thank you for reporting. Odd it didn't fail any tests where this was used (I guess it isn't). I remember thinking, "I don't have to add a unit test for this list since the base `EnumObjectList` is already tested" 🤦 .
> 
> But, isn't a better solution here to redirect to `super.add()`. Basically just appending: `.also { add( it ) }`? Nothing is shaded, it takes different function parameters. It's overloaded.

yes overloaded, but I also think that it actually is shadowed :) I started using `also` but thought it was more convoluted. Changed now to more "classic idiomatic" use of super and retaining the `add` method name

Whathecode:
> Changed now to more "classic idiomatic" use of super and retaining the add method name

I just came to the same conclusion on my local branch here. 😂 just `add( DataTypeMetaData( ... ) )` indeed.

Whathecode (approved):
Great find! Thanks. 🙏

PR #276, opened 2021-06-21
Support `Measure.TriggerData` to indicate a task should measure triggers.

Whathecode:
This converts the previous `Measure` into `Measure.DataStream` in the codebase and adds a `Measure.TriggerData` (https://github.com/cph-cachet/carp.core-kotlin/commit/a622605c68294718c5f628a0e32a131f3cfcd805).

Currently, there is no API available yet to easily initialize `Measure.TriggerData`. This will be done in later commits to this draft PR.

Since this feature required a trigger to be identified by ID _before_ it is converted to a snapshot, I now assign trigger IDs when they are added to `StudyProtocol`. As a consequence, snapshots are no longer equal when they contain the exact same triggers added in a different order, since they will be assigned different IDs. If we ever need a "smarter" comparison check (I currently don't know of a use case), we can implement that then.

Since triggers now already have an ID in the protocol, there is also an opportunity to refactor `TaskControl` which currently gets replaced by a DTO (swapping out a `Trigger` reference with an ID) when converting a `StudyProtocol` into a snapshot.


Whathecode:
> Since triggers now already have an ID in the protocol, there is also an opportunity to refactor TaskControl which currently gets replaced by a DTO (swapping out a Trigger reference with an ID) when converting a StudyProtocol into a snapshot.

`TaskControl` is still very useful for the `StudyProtocol.addTaskControl` API, which is flexible enough to _add any new trigger_ that aren't registered in the protocol yet. Therefore, I will keep it.

carp.clients.core/src/commonMain/kotlin/dk/cachet/carp/clients/domain/StudyRuntime.kt
```diff
@@ -191,7 +192,10 @@ class StudyRuntime private constructor(
                     ?: throw UnsupportedOperationException( "Connecting to device of type \"$deviceType\" is not supported on this client." )
             }
 
-            val dataTypes = tasks.flatMap { it.measures }.map { it.type }.distinct()
+            val dataTypes = tasks
+                .flatMap { it.measures.filterIsInstance<Measure.DataStream>() }
```
ltj, line 196: I wonder if task measures should be split up rather than "overloaded" like this.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/data/NoData.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import kotlinx.serialization.Serializable
+
+
+/**
+ * Placeholder for generic `Data` types to indicate there is no associated data.
+ * This should not be serialized; instead, nullable `Data` should be used.
+ */
+@Serializable
+object NoData : Data
```
ltj, line 11: Maybe put this in the same file as `Data` - doesn't matter much, it just seems very old-school Java-like :)

ltj (dismissed):

carp.clients.core/src/commonMain/kotlin/dk/cachet/carp/clients/domain/StudyRuntime.kt
Whathecode: Both are measures. The other instance is simply `Measure.TriggerData` which should _always_ supported, and `Measure.TriggerData` does not contain a `dataType` (since it is _always_ the same data type). The difference is that for `Measure.DataStream` the researcher could incorrectly configure a `dataType` to be measured on a device, which is not supported at runtime.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/data/NoData.kt
Whathecode: Makes perfect sense! I also didn't add it in the docs.

Issue #156, opened 2020-09-04
Generate TypeScript declarations using new JS IR backend


Whathecode:
Kotlin 1.4.0 supports generating TypeScript (TS) declarations with the new JS IR backend. This would be nice, so we don't have to maintain the currently manually-written ones.

However, to create TS declarations the `@JsExport` needs to be applied to classes. This currently fails for classes with companion objects: https://github.com/Kotlin/kotlinx.serialization/issues/980#issuecomment-685874328

And, there are limitations we need to look into: e.g., `Long` cannot be exported (yet?).

Whathecode:
I now upgraded to Kotlin 1.5.10 (can't upgrade to 1.5.20 yet due to [a multimodule regression bug](https://github.com/Kotlin/kotlinx.serialization/issues/1580)). I tried enabling the IR backend for JS and got: `java.lang.IllegalStateException: Symbol with IrTypeParameterSymbolImpl is unbound`

This [might potentially be fixed in 1.5.20](https://youtrack.jetbrains.com/issue/KT-45236), but we'll have to wait to find out.

Whathecode:
Upgrading to Kotlin 1.5.30 now seems blocked by a compiler error for the JVM target: "Backend Internal error: Exception during IR lowering" of `StudyProtocolSnapshot`.

This should be fixed in the next release, after which we can attempt another switch to the JS IR backend: https://github.com/Kotlin/kotlinx.serialization/issues/1584

Whathecode:
Upgrading to Kotlin 1.6.0 still seems to result in a kotlinx.serialization compiler plugin error for the JVM target: "Backend Internal error: Exception during IR lowering".

While a bug with the same exception message is fixed in 1.6.0, the same error now seems to happen for a different reasons: https://github.com/Kotlin/kotlinx.serialization/issues/1770

Additionally, type inference fails after upgrading to 1.6.0: https://youtrack.jetbrains.com/issue/KT-49838 This is less of a blocker since the types can simply be specified manually as a workaround.

Whathecode:
1.6.10 fixed the previously reported issues. But, seemingly exporting enum classes isn't supported yet. This [seems to be slated for 1.6.20](https://youtrack.jetbrains.com/issue/KT-37916).

Whathecode:
Can't generate TS declarations yet for enums and interfaces. The following types therefore don't have TS declarations yet on the JS IR branch.

**common**:

- enums:
  - `Trilean`,
[...]

PR #284, opened 2021-07-12
Upgrade all the things

### LINKED ISSUE #28: Multiplatform documentation configuration
Currently, generating documentation only works for the JVM target (Javadoc). The plugin used to generate this documentation does not support Javascript targets yet. But, it seems like this will be supported in the future: https://discuss.kotlinlang.org/t/how-to-configure-dokka-for-kotlin-multiplatform/9834

Since Dokka currently does not support JS yet, a few hacks are included in the codebase:
- The main build file overrides default (failing) dokka behavior ('HACK' section).
- Individual project build files (e.g., 'carp.deployment.core') need to include sources of the other projects they depend on. Otherwise, these types are not included in the generated documentation.

This can likely be removed once Dokka supports multiplatform targets.

### LINKED ISSUE #83: Release using `nexus-publish-plugin` rather than `maven-publish`
The [nexus-publish-plugin](https://github.com/marcphilipp/nexus-publish-plugin/) provides a better integration with the Sonatype Nexus Repository to release packages to Maven.

> Gradle Plugin that explicitly creates a Staging Repository before publishing to Nexus. This solves the problem that frequently occurs when uploading to Nexus from Travis, namely split staging repositories.

Currently, the [gradle-nexus-staging-plugin](https://github.com/Codearte/gradle-nexus-staging-plugin) is used to close and release staging repositories after having successfully published to them using `maven-publish`. Although this succeeds, this gives a [deprecation warning that the staging repository ID has not been provided](https://github.com/Codearte/gradle-nexus-staging-plugin/issues/127) which is the recommended way moving forward. Currently, it uses some heuristics to decide which repository to close (I'm guessing it tries to close the latest one).

### LINKED ISSUE #154: Remove hack: `jsPackageJson` in gradle publish tasks
The applied hack: https://github.com/cph-cachet/carp.core-kotlin/commit/0d7c8bc3d7c3119ad41ea3918433dce7fa0e35bf

The reason: https://youtrack.jetbrains.com/issue/KT-41658

### LINKED ISSUE #256: Upgrade to Kotlin 1.5.0 and serialization 1.2.1
When trying to upgrade to Kotlin 1.5.0, internal compiler errors are thrown when trying to process `CustomInput`. The problem seems the same as [this report on YouTrack](https://youtrack.jetbrains.com/issue/KT-46444).

However, even when removing the generic type parameter of `CustomInput`, I receive different compiler errors:

```
e: java.lang.NullPointerException
	at org.jetbrains.kotlinx.serialization.compiler.backend.ir.SerializableIrGenerator.getSerialDescriptorExpr(SerializableIrGenerator.kt:175)
	at org.jetbrains.kotlinx.serialization.compiler.backend.ir.SerializableIrGenerator.access$getSerialDescriptorExpr(SerializableIrGenerator.kt:38)
	at org.jetbrains.kotlinx.serialization.compiler.backend.ir.SerializableIrGenerator$generateInternalConstructor$1.invoke(SerializableIrGenerator.kt:121)
	at org.jetbrains.kotlinx.serialization.compiler.backend.ir
```

Currently, it seems we can't upgrade to 1.5.0 yet. Once the bug report on YouTrack is resolved/released, we can try again.

An upgrade to serialization 1.2.1 will also allow to fix https://github.com/Kotlin/kotlinx.serialization/issues/1323

Whathecode:
image

- Upgrade to Kotlin 1.5.21, serialization 1.2.2, and coroutines 1.5.1 (Closes #256).
- With Kotlin up-to-date, I attempted to switch to the JS IR backend to try generating TypeScript, but this still doesn't work (#156).
- Finally upgraded to dokka 1.5.0 to generate Javadoc for the Maven publications (Closes #28). With that in place, we are likely also ready to publish full multiplatform documentation next (#285).
- Switched to new sonatype nexus publication plugin; old one was no longer maintained and sometimes made publication fail (Closes #83). This very likely also closes #154.
- Various other minor dependency updates: detekt 1.18.0-RC1 and npm dependencies for the `verifyTsDeclarations` module/task.

Lastly, I also cleaned up the build output a bit by suppressing warnings which we know aren't a problem.

### DIFF
--- a/.github/workflows/publish-snapshots.yml
+++ b/.github/workflows/publish-snapshots.yml
@@ -30,4 +30,4 @@ jobs:
 
       - name: Deploy snapshot
         if: success()
-        run: ./gradlew jsPackageJson publishSnapshot
+        run: ./gradlew setSnapshotVersion publishToSonatype closeAndReleaseSonatypeStagingRepository
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,10 +28,6 @@ jobs:
         if: success()
         run: ./gradlew verifyTsDeclarations
 
-      - name: Publish to staging repo
+      - name: Publish to staging repo and release
         if: success()
-        run: ./gradlew jsPackageJson publishSigned
-      
-      - name: Release staged repo
-        if: success()
-        run: ./gradlew closeAndReleaseRepository
+        run: ./gradlew publishToSonatype closeAndReleaseSonatypeStagingRepository
--- a/README.md
+++ b/README.md
@@ -302,7 +302,7 @@ In case you want to contribute, please follow our [contribution guidelines](http
 We recommend using IntelliJ IDEA 2020, as this is the development environment we use and is therefore fully tested.
 
 - Open the project folder in IntelliJ 2020.
-- Install the Kotlin plugin for IntelliJ IDEA (203-1.4.31-release-IJ*): `Tools->Kotlin->Configure Kotlin Plugin Updates`
+- Install the Kotlin plugin for IntelliJ IDEA (211-1.5.21-release-*): `Tools->Kotlin->Configure Kotlin Plugin Updates`
 - To build/test/publish, click "Edit Configurations" to add configurations for [the included Gradle tasks](#gradle-tasks), or run them from the Gradle tool window.
 
 ### Gradle tasks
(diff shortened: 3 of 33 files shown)


ltj (commented):
👏

PR #291, opened 2021-08-09
Release 1.0.0-alpha.35

Whathecode:
Dependency updates:

- Kotlin 1.5.21
- kotlinx.serialization 1.2.2
- coroutines 1.5.1

Timestamp variables are now consistently named using the past tense followed by `on`, rather than a `date` or `time` suffix. E.g., `createdOn` rather than `creationDate`. This could cause confusion as to whether only a date or time of day was stored.

### Common
- Replaced previously custom `DateTime` with kotlinx-datetime `Instant`
- Replaced previously custom `TimeSpan` with kotlin's `Duration`
- `CustomInput` is no longer generic


ltj (approved):
🚀

PR #303, opened 2021-09-14
Support setting multiple input data for participant data at once.

### LINKED ISSUE #242: Participant data get response in JSON serialization is inconsistent with the format for setting it
When getting participant data the following is returned:

```
{
  "some.inputdata.key": "Some value"
}
```

But, when setting it you need to specify `$type` and `value` and `inputDataType`.

should the returned data also contain `$type` and `value`? I.e.:

```
{
  "some.inputdata.key": { "$type": "some.inputdata.key", "value": "Some value" }
}
```

Whathecode:
Originally, `ParticipationService.setParticipantData` could only set one participant data field at a time. I believe this will still be the common use case. But, the following changes will allow setting multiple fields at once (e.g., to support a "save" action after having edited input multiple fields).

However, the primary reason for this change was consistency in the serialization format (#242). Because `getParticipantData` returned multiple fields, and `setParticipantData` only set one, the resulting return and request object seemed inconsistent.

Closes #242


carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/application/ParticipationService.kt
```diff
@@ -42,13 +42,13 @@ interface ParticipationService : ApplicationService<ParticipationService, Partic
...
      * @throws IllegalArgumentException when:
      *   - there is no study deployment with [studyDeploymentId]
-     *   - [inputDataType] is not configured as expected participant data in the study protocol
-     *   - [data] is invalid data for [inputDataType]
+     *   - one or more of the keys in [data] isn't configured as expected participant data in the study protocol
+     *   - one or more of the set [data] isn't valid for the corresponding input data type
      * @return All data for the specified study deployment, including the newly set data.
      */
-    suspend fun setParticipantData( studyDeploymentId: UUID, inputDataType: InputDataType, data: Data? ): ParticipantData
+    suspend fun setParticipantData( studyDeploymentId: UUID, data: Map<InputDataType, Data?> ): ParticipantData
```
ltj, line 53: Is it a necessary constrain to only allow one of each `InputDataType` ?

ltj, line 53: Also I'm not quite sure why `data` is nullable. If something is optional, the caller should simply not set it unless there actually is a value to set.

carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/application/ParticipationServiceHost.kt
```diff
@@ -101,18 +101,18 @@ class ParticipationServiceHost(
...
 
     /**
-     * Set participant [data] for the given [inputDataType] in the study deployment with [studyDeploymentId].
+     * Set participant [data] in the study deployment with [studyDeploymentId].
      *
      * @throws IllegalArgumentException when:
      *   - there is no study deployment with [studyDeploymentId]
-     *   - [inputDataType] is not configured as expected participant data in the study protocol
-     *   - [data] is invalid data for [inputDataType]
+     *   - one or more of the keys in [data] isn't configured as expected participant data in the study protocol
```
ltj, line 108: Just a thought: should we really replicate documentation verbatim on the implementations?

carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/application/ParticipationService.kt
Whathecode: Yes, this is by design and I believe where some of the confusion came from. `inputDataType` is supposed to be unique. That's why for custom fields (which share the same _data_ type), a unique `inputDataType` is generated.

The idea here is that what fully defines the data to be requested, e.g. "biological sex", is unique. This does presume data is requested _for one specific participant_. But, we currently take shortcuts here and have not yet designed multiple participants into core, though we do know where to fit it in afterwards.

Whathecode: Maybe an indicator this should be named differently? Or, documented better?

Whathecode: Good point. I did not do this purposefully. It was just a copy/paste. _But_, by making it nullable you can _unset_ values. Otherwise, this is not supported.

In other words, setting only a subset of fields is still possible. But passing null is to support removing previously set data.

carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/application/ParticipationServiceHost.kt
Whathecode: I've been wondering about this. I guess I need to check dokka/IntelliJ support.

I.e., does IntelliJ by default use the interface documentation if nothing is set on the concrete class? And, does dokka do the same when generating docs? I'll create an issue to look into this.

Whathecode: https://github.com/cph-cachet/carp.core-kotlin/issues/304

carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/application/ParticipationService.kt
ltj: > Maybe an indicator this should be named differently? Or, documented better?
It's more like `InputDataId` since that's really what it is.

ltj: Documentation should be updated with the additional "unset" semantics then :)

ltj (approved):

carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/application/ParticipationService.kt
Whathecode: That would be slightly inconsistent with `typealias DataType = NamespacedId`, however. I think the point here is that it's _usage_ here is as an _id_, but it also is a `DataType`.

Whathecode: I think `InputDataId` has its own problems (elsewhere in the code where this is used). Instead, I just updated the documentatoin of `InputDataType` slightly. I think any confusion here will best be addressed through documentation.

PR #308, opened 2021-09-20
Use custom application service exceptions

### LINKED ISSUE #298: Domain exceptions for application services
Broadly speaking every `IllegalArgumentException` that comes out of services right now should safely map to a 400 of corresponding web requests. The problem lies in when the underlying domain objects throw `IllegalArgumentException`'s which do not map to illegal arguments of the service requests.

For that reason it would make sense to have domain specific `IllegalArgumentException`'s for application services, so that we can uniformly map domain-specific exceptions to web errors. `IllegalArgumentException`'s can then safely be mapped to 500's to address the former problem.

Whathecode:
So far:

- Introduce custom `ApplicationServiceArgumentException` and `ApplicationServiceStateException`.
- Replaced previous exceptions with the new ones in `DeploymentService`.

This should already be sufficient for a draft review, as the remainder of the PR will simply replace exceptions in all remaining application services.

Closes #298

### DIFF
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/services/ApplicationServiceException.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/services/ApplicationServiceException.kt
@@ -1,0 +1,53 @@
+package dk.cachet.carp.common.application.services
+
+import kotlin.reflect.KCallable
+
+
+/**
+ * An argument passed to the application service [endpoint] is invalid.
+ *
+ * When the application service is exposed as a web service, this maps to response status code 400 (Bad Request).
+ */
+class ApplicationServiceArgumentException(
+    val endpoint: KCallable<*>,
+    cause: Throwable? = null,
+    message: String? = null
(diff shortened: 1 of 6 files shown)


carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/services/ApplicationServiceException.kt
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * The requested operation exposed by [endpoint] is invalid given the current state of the objects it acts on.
+ *
+ * When this application service is exposed as a web service, this maps to response status code 409 (Conflict).
+ */
+class ApplicationServiceStateException(
+    val endpoint: KCallable<*>,
+    cause: Throwable? = null,
+    message: String? = null
+) : IllegalArgumentException( "Illegal state while calling \"${endpoint.name}\": ${message ?: cause?.message}", cause )
```
Whathecode, line 27: If we extend from an exception, this should be `IllegalStateException`.

PR #311, opened 2021-09-20
Add method to get default `SamplingConfiguration` for a data type on a device

### LINKED ISSUE #310: Can't access `BatteryAwareSamplingScheme` defaults.
The `normal`, `low`, and `critical` configurations for `BatteryAwareSamplingScheme` are `private`.

But, master devices need to be able to access these to know how to sample by default.

It would also be useful to add a helper method to determine the sampling scheme which should be used based on the configured `defaultSamplingConfiguration` in `DeviceDescriptor`.

Whathecode:
@thaiphandinh can use this to simplify retrieving the default sampling configuration for any data type through `DeviceDescriptor`. This can facilitate a more generic implementation to pass the required `SamplingConfiguration` to the data collector.

Of course, this way the _concrete_ `SamplingConfiguration` type is unknown, so you will still need to cast to the expected configuration per data type. Hope this helps.

Closes #310 

I tried to also fix a TODO I encountered:

> TODO: Verify whether all configured data types are supported by this device (supported data streams), probably in init.
>   We might also want to check whether the sampling configuration instances are valid.

However, this proved too be too difficult for now. Instead, I created a new issue for this: https://github.com/cph-cachet/carp.core-kotlin/issues/312

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,12 @@ Two key **design goals** differentiate this project from similar projects:
 ## Table of Contents
 
 - [Architecture](#architecture)
+  - [Common](docs/carp-common.md)
+    - [Data types](docs/carp-common.md#data-types)
+    - [Device descriptors](docs/carp-common.md#device-descriptors)
+    - [Sampling schemes and configurations](docs/carp-common.md#sampling-schemes-and-configurations)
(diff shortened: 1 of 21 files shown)


thaiphandinh (approved):
Thank you for adding documentations. It made things more clear now.

PR #313, opened 2021-09-22
Add `ClientRepository.removeStudyRuntime()`.

### LINKED ISSUE #299: Add `removeStudyRuntime` to `ClientRepository`.
When quitting a study on a client, it should be possible to remove the runtime information for the `ClientRepository`.

Whathecode:
Closes #299

### DIFF
--- a/carp.clients.core/src/commonMain/kotlin/dk/cachet/carp/clients/domain/ClientRepository.kt
+++ b/carp.clients.core/src/commonMain/kotlin/dk/cachet/carp/clients/domain/ClientRepository.kt
@@ -42,4 +42,10 @@ interface ClientRepository
      * @throws IllegalArgumentException when no previous version of this study runtime is stored in the repository.
      */
     suspend fun updateStudyRuntime( runtime: StudyRuntime )
+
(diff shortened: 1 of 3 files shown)


thaiphandinh (approved):
all good! Thank you for adding this. 
I'm wondering do we need a function like `deleteAll() `for `studyRuntime` because we know that at some point  we want to have multiple studies running in parallel.

Whathecode:
> I'm wondering do we need a function like deleteAll() for studyRuntime because we know that at some point we want to have multiple studies running in parallel.

I don't understand how that is related to studies running parallel. Either way, `deleteAll` would be a simple API enhancement, since it's already possible to do so with the current API. I expect eventually it wouldn't be a common use case, so I didn't add it. It's unclear why you would remove studies if they are already filtered out as "stopped".

thaiphandinh:
I meant when we have more than one running study, somehow user might need to stop them all at the same time, like when sign out or hit a button to stop all the running study. (just a feature which doesn't exist yet). But you're right, we can even do it with the current API

PR #314, opened 2021-09-24
Add time when participant group was deployed and stopped

### LINKED ISSUE #255: Create a `StagingParticipantGroup` in the studies subsystem
The deployments subsystem currently has a `ParticipantGroup` which holds the set of participants and data related to them once the participant group _has been deployed_.

However, there are also use cases in the studies subsystem in which the researcher doesn't or can't define a full participant group "in one go". There might be a need to define a participant group with the intent of deploying it at a later point in time once the configuration is completed. Such configuration could be adding preregistered devices, etc.

Right now, the studies subsystem only has a `deployParticipantGroup` endpoint which requires a fully specified group as input and instantly sends out invitations to all participants. Once more functionality is needed, we should add a `StagedParticipantGroup` which allows the researcher to construct a `ParticipantGroup` (to be deployed) in an incremental fashion.

Whathecode:
A `StagedParticipantGroup` was added in `Recruitment` of the studies subsystem to store participant groups that aren't necessarily deployed yet. However, each call to `RecruitmentService.deployParticipantGroup` (now named `inviteNewParticipantGroup`) still creates the participant group while deploying it, but the inner domain classes are now prepared for exposing "staging" participant group endpoints.

I already implemented this since it was a related upcoming requirement to the main focus of this PR: specify timestamps when a participant (group) was _invited_, and when the associated deployment was _stopped_.

This is now exposed through `ParticipantGroupStatus` which is turned into a state machine with states `Staged`, `Invited`, `Running`, and `Stopped`.

`RecruitmentService.deployParticipantGroup` was renamed to `inviteNewParticipantGroup`. Two reasons:

-  the "deployment" logic of registering devices is now wrapped (it's less relevant to the researcher), though still accessible through the inner exposed `studyDeploymentStatus`. From the researcher's perspective, the action is more suitably named "invite".
- this prepares for disambiguation for new endpoints we will later need: `createParticipantGroup` and `inviteParticipantGroup`

Participant Group Status


carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/application/StudyDeploymentStatus.kt
```diff
@@ -50,6 +57,7 @@ sealed class StudyDeploymentStatus
      */
     @Serializable
     data class DeploymentReady(
+        override val createdOn: Instant,
         override val studyDeploymentId: UUID,
         override val devicesStatus: List<DeviceDeploymentStatus>,
         override val startedOn: Instant?
```
ltj, line 63: `readyOn`? I guess it marks the time when the client signal the fetched deployment is 'ready'. It appears the use of 'started' was deliberately avoided here in the first place since we cannot know whether the client is actually running the study.

ltj (approved):
👍

carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/application/StudyDeploymentStatus.kt
Whathecode: No, this is the time the deployment was created, coinciding with when invitations were sent out. `startedOn` is I think what you mean by `readyOn`. Maybe a good candidate for a rename 🤔 .

`startedOn` is set once all devices participating in the deployment have called `deploymentSuccessful`.

Whathecode: Oh, I now totally see you did comment on the correct line. I got confused because `createdOn` was highlighted as the change. :)

PR #317, opened 2021-10-04
Clearer `DeviceDescriptor` API for client subystem.

Whathecode:
When `DeviceDescriptor` was used in the client subsystem, the multiple "default" sampling configurations were confusing. These changes aim to make the distinction clearer.

In addition, for known supported data types on `DeviceDescriptor`, the sampling configuration is now validated (https://github.com/cph-cachet/carp.core-kotlin/issues/312).

I'm still not 100% satisfied with the naming, as exemplified by the force push I did just now. 😂

Should we use `defaultSamplingConfigurations` (plural) instead? It's singular right now because I thought originally in JSON and the builder singular looked better, e.g.:

```
val phone = Smartphone( "Patient's phone" )
{
    // Configure device-specific options, e.g., frequency to collect data at.
    defaultSamplingConfiguration {
        geolocation { batteryNormal { granularity = Granularity.Balanced } }
    }
}
```

PR #318, opened 2021-10-05
Validation for `SamplingConfiguration` and cleaner client API to obtain defaults

Whathecode:
Second attempt of #317


carp.protocols.core/src/commonMain/kotlin/dk/cachet/carp/protocols/domain/StudyProtocol.kt
```diff
@@ -127,7 +127,9 @@ class StudyProtocol private constructor( val ownerId: UUID, val name: String, va
      * Add a [masterDevice] which is responsible for aggregating and synchronizing incoming data.
      * Its role name should be unique in the protocol.
      *
-     * @throws IllegalArgumentException in case a device with the specified role name already exists.
+     * @throws IllegalArgumentException when:
+     *  - a device with the specified role name already exists
```
thaiphandinh, line 131: Why is it relevant?

thaiphandinh (approved):
All good. I'm still not hundred percent satisfy with the place in DataListener but it's fine for now. May be as you said it would be better to only have it on client side.

Whathecode:
> I'm still not hundred percent satisfy with the place in DataListener

Agreed. :) I actually thought of moving it to `StudyRuntime` after finishing this PR yesterday. 😂

carp.protocols.core/src/commonMain/kotlin/dk/cachet/carp/protocols/domain/StudyProtocol.kt
Whathecode: Device role name being unique? It's the unique key that determines the _source_ of the data. It's also [part of the identity of a "data stream"](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamId.kt#L13).

If it wouldn't be unique, e.g., you'd name two SENS patches "SENS patch", how would you know which one is the one collecting data for the right leg and which one is the one collecting data on the left leg? And, the data would become the same "data stream", thus data from both sensors would be bunched together and you could no longer tell them apart.

PR #321, opened 2021-10-17
Major rework of existing client functionality to adhere more to DDD principles

Whathecode:
`StudyRuntime` is split into `Study` and `StudyDeploymentProxy`. This makes `Study` a proper aggregate root with no dependencies on application services. Accordingly, the state machine of `Study` (represented through `StudyStatus`) changed considerably. It now reflects more detailed state based on incoming study deployment events. The only two methods on `Study` now are `deploymentStatusReceived` and `deviceDeploymentReceived`, also allowing incoming changes from the remote study deployment to come in.

Study state machine

Along with this, also for the client subsystem to adhere more to domain-driven design (DDD), the main API, accessed through `ClientManager`, has been moved to the `application` namespace.

Lastly, `ClientManager.addStudy()` now no longer calls `tryDeployment()`. You can add a study without trying to deploy it, and it will be in the `DeploymentNotStarted` state.

Next, `tryDeployment` should likely be split into something akin to `receiveDeviceDeployment` and `startStudy` so that sensor permissions can be obtained after the first step already, which currently isn't possibly. But, I will look into alternatives and consider this a separate PR.

I also discovered that the `StudyDeploymentProxy` tests should probably be moved to `ClientManager`, and different tests relying on mocks should be used in the proxy.


Whathecode:
@thaiphandinh I'm pulling in these changes since the PR is over a month old. It was resulting in too many conflicts with other branches I was working on in parallel.

You can still provide feedback, but I'll incorporate it in future PRs, so ideally reported as issues.

Issue #118, opened 2020-04-07
Which devices require registration and which one are 'dependent'?


Whathecode:
While continuing the implementation of `DeploymentService`, some questions arose in regards to the registration of devices, and when devices are considered 'dependent' on one another.

**Dependent devices**
The original reason why it was expected that some devices _require_ registration prior to being able to retrieve the `MasterDeviceDeployment` of another device was because they might need to communicate and information needs to be exchanged in order to establish that communication. **However**, we also envisioned that the back-end may decide to only support server/client communication in between multiple master devices; that is, the individual master devices do not depend on one another, but all depend on the server as the middleman to communicate with one another. Since this would be the default fallback mechanisms, this is the first form of communication we intend to implement. In case of such a deployment, master devices _do not depend on one another_. To contrast, the current implementation assumes they always do, even if no form of communication between them is configured in the study protocol.

- We should likely clarify what 'dependent' means. Functionally, the concept was introduced to withhold `MasterDeviceDeployment` in case the information on how to communicate with other devices was not yet available and could be made part of that deployment information.

**Required registration**
Some ambiguity in terms of what it means for a device 'requiring registration' arose. The concept was introduced to indicate the dependencies described above, and thus all top-level master devices by default required registration. However, now the question also arises whether _all_ devices should be registered prior to a study deployment being able to 'start'. Connected devices can be registered locally on a master device, but it might make sense to notify the back-end of that registration so that a complete history of which physical devices were used to run a study is available.

Whathecode:
The second part '**required registration**' is addressed in https://github.com/cph-cachet/carp.core-kotlin/pull/119.

> All devices require registration on the back-end before a deployment can be considered 'ready' (https://github.com/cph-cachet/carp.core-kotlin/commit/eb4aa8862be1cb24c8272a6dce7c8004b96ee782). The previous `requiresRegistration` is now specific to individual devices, represented as `remainingDevicesToRegisterBeforeDeployment` in `DeviceDeploymentStatus` (https://github.com/cph-cachet/carp.core-kotlin/commit/b067518d733e0886ea86865541b6d197aaf8aa9a).

Whathecode:
Another formulation I just wrote up while looking into the different definitions again:

- Dependent devices: all master devices another master device depends on the registration of to run the study deployment.
[...]

PR #323, opened 2021-11-02
Optional devices

Whathecode:
Devices can now be _optional_, specified as a `isOptional` property on `DeviceDescriptor`, thus configurable in the `StudyProtocol`. Normally, all master devices need to be deployed before a deployment is considered "running" (since we still assume they are "dependent" by default, [which we may want to change in the future](https://github.com/cph-cachet/carp.core-kotlin/issues/118)). Now, only non-optional master devices need to be deployed.

There are likely meaningful `DeploymentWarning`s that can be added to the protocols subsystem based on these changes. I.e., triggers coming from optional devices targeting other devices. But, I'll consider this out-of-scope for now.

I refactored unit tests so that I could add two tests which relied on having access to concrete instances (previously private) of all `DeviceDescriptor`s in the codebase:

- `all_device_descriptors_have_an_instance_for_testing`
- `all_device_descriptors_have_correct_defaults`: the default for `isOptional` should always be `false` for API consistency. Likewise `defaultSamplingConfiguration` should always be empty.


Whathecode:
[Validation when inviting participant groups](https://github.com/cph-cachet/carp.core-kotlin/blob/optional-devices/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt#L104) probably still needs to be updated as well. Calling `inviteNewParticipantGroup` will still fail, I think, with unassigned master devices: "not all master devices part of the study have been assigned a participant"

Whathecode:
Have a look whether we can add a `DeploymentError` when there are only optional devices in the protocol.

Whathecode:
@ltj I included the discussed protocol warning: `OnlyOptionalDevicesWarning`

Have a look. :)

Whathecode:
Merging this since maintaining 4 branches is taking up too much of my time and I need to start work on new features/branches. I want to prevent further merge conflicts that already exist on the other branches.

The warning is implemented with the feedback during our meetings in mind. If any further changes are needed we can still address them in future PRs.

Issue #330, opened 2021-11-20
Support raw acceleration sampling on Smartphone


Whathecode:
_Raw_ accelerometry data can be retrieved both on Android and iOS phones. However, [other than excluding gravity it is unclear _how_ raw such data is](https://stackoverflow.com/questions/69831128/how-raw-is-ios-accelerometry-data-retrieved-through-cmmotionmanager/69831332#69831332). Specifically, Android has two options to choose from.

Therefore, [we have decided to only expose the calibrated streams for now](https://github.com/cph-cachet/carp.core-kotlin/issues/322#issuecomment-970269189), which seems the same in intent for both devices.

Once we understand the use case for raw data better, we can add extra streams as new features.

Whathecode:
One possible use case is that [raw accelerometry consumes considerable less power](https://developer.android.com/guide/topics/sensors/sensors_motion):

> In general, the accelerometer is a good sensor to use if you are monitoring device motion. Almost every Android-powered handset and tablet has an accelerometer, and it uses about 10 times less power than the other motion sensors.

PR #331, opened 2021-11-20
Include acceleration and angular velocity sensors for `Smartphone`

Whathecode:
We already had acceleration defined as a data type, but it was ambiguous whether it included or excluded gravity. Now, it is defined as "non-gravitational acceleration", thus excluding gravity. If we want to support measuring other types of acceleration (including gravity, more 'raw', etc.) we first need to better understand what other data the APIs provide, [which isn't immediately clear for smartphones](https://github.com/cph-cachet/carp.core-kotlin/issues/330).

Similarly, angular velocity refers to _calibrated_ data. We can see later how to address uncalibrated data. However, it must be noted that on top of the calibration done by the phone, some _additional_ calibration may still seems needed.

> The linear acceleration sensor always has an offset, which you need to remove: https://developer.android.com/guide/topics/sensors/sensors_motion

I set the default sampling interval for non-gravitational acceleration and angular velocity to 200 ms based on reading the Android docs. This seems to correspond to the "normal" sampling frequency, which is very low, _but_, for continuous passive measuring probably appropriate as not to drain the battery too much.


thaiphandinh (approved):
I would say those new names made things better then before 💯

ltj (approved):
👍

PR #333, opened 2021-11-23
`SyncPoint` improvements

Whathecode:
Passing sync points is a complex use case only needed for devices which aren't already synchronized with UTC timestamps. Therefore, a default option `SyncPoint.UTC` is provided which indicates the timestamps corresponds to UTC timestamps in microseconds.

To help working with sync points, several helper functions will also be added.

### DIFF
--- a/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/SyncPoint.kt
+++ b/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/SyncPoint.kt
@@ -29,3 +29,13 @@ data class SyncPoint(
     @Required
     val relativeClockSpeed: Double = 1.0
 )
+{
+    companion object
+    {
+        /**
+         * The default [SyncPoint] for timestamps that are already synchronized to UTC time.
+         * A synchronization conversion using this sync point is a no-op.
+         */
+        val UTC: SyncPoint = SyncPoint( Instant.fromEpochSeconds( 0 ) )
+    }
+}
--- a/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamBatch.kt
+++ b/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamBatch.kt
@@ -52,8 +52,9 @@ class MutableDataStreamBatch : DataStreamBatch
     /**
      * Append a sequence to a non-existing or previously appended data stream in this batch.
      *
-     * @throws IllegalArgumentException when the start of the [sequence] range precedes the end of
-     *   a previously appended sequence to the same data stream.
+     * @throws IllegalArgumentException when:
+     *  - the start of the [sequence] range precedes the end of a previously appended sequence to the same data stream
+     *  - the sync point of [sequence] is older than that of previous sequences in this batch
      */
     fun appendSequence( sequence: DataStreamSequence )
     {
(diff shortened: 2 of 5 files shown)


pmasulli:
After discussing with @Whathecode, we agree that a better choice of data for a `SynchPoint` is:
- `synchronizedOn`: the UTC time at the moment of the synchronization
- `sensorTimeStampAtSynchPoint`: the sensor time at `synchronizedOn`
- `relativeClockSpeed`: the variation of the UTC time divided by the variation of sensor time.

With these definitions, the timestamp transformation formula becomes:
`synchronizedTimeStamp = relativeClockSpeed * (sensorTime - sensorTimeStampAtSynchPoint) + synchronizedOn`.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/Measurement.kt
```diff
@@ -40,6 +40,17 @@ data class Measurement<out TData : Data>(
     fun getDataTimeType(): DataTimeType =
         if ( sensorEndTime == null ) DataTimeType.POINT
         else DataTimeType.TIME_SPAN
+
+    /**
+     * Convert this [Measurement] to one synchronized to UTC time using [syncPoint].
+     */
+    fun synchronizeToUTC( syncPoint: SyncPoint ): Measurement<TData> =
+        copy(
+            sensorStartTime = syncPoint.synchronizeTimestamp( sensorStartTime ),
```
thaiphandinh, line 49: `syncPoint.synchronizeTimestamp` returns `syncedTimestamp` value based on `syncPoint` and `sensorStartTime` value. I doubt does this function really convert `sensorStartTime` and `sensorEndTime` to UTC value?

thaiphandinh (changes requested):
Thank you for this PR, it've helped me to get a better picture of SyncPoint. 

My biggest concern is the helper functions which converts to UTC values. As you can see I had several comments which showing how confuse I'm when following this. 

The Implementation to convert to UTC values doesn't convince me so far, the main reason was I didn't see any thing which _correctly_ converts a local `timezone` to `UTC` which I think it would exist when I read the document. Please correct me if I'm wrong.

To me either the comment/documentation or the actual implementation should be improved at some point.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/SyncPoint.kt
```diff
@@ -19,13 +20,38 @@ data class SyncPoint(
...
+         * A synchronization conversion using this sync point is a no-op.
+         */
+        val UTC: SyncPoint = SyncPoint( Instant.fromEpochSeconds( 0 ) )
+    }
+
+
+    /**
+     * Converts [timestamp] to UTC time in microseconds using this [SyncPoint].
+     */
+    fun synchronizeTimestamp( timestamp: Long ): Long
```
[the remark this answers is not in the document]
Whathecode: A UTC timestamp is not the same as UTC time. I think you correctly point out the documentation, and possibly method name, is misleading in that regard. This function coverts the sensor timestamp to microseconds since the Unix epoch.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/Measurement.kt
Whathecode: Yes. The `synchronizeTimestamp` function returns `syncedTimestamp`. This corresponds to microseconds since the Unix epoch, which is expressed as a moment in UTC time, and thus more or less stores UTC time indirectly
But I can see how the shorthand in the documentation did more harm then good. :) I will update.

thaiphandinh (dismissed):

ltj (approved):

PR #336, opened 2021-12-01
Adding WebTask and linked ts declarations

jeyjey626:
In effort to create a full `StudyProtocolSnapshot` ts declaration, WebTask declarations and related ts declarations such as `Trigger` and tests for their instances were added.


typescript-declarations/tests/carp.common.ts
```diff
@@ -52,6 +60,19 @@ describe( "carp.common", () => {
             UsernameAccountIdentity.Companion,
             [ "ParticipantAttribute", new ParticipantAttribute.DefaultParticipantAttribute( new NamespacedId( "namespace", "type" ) ) ],
             ParticipantAttribute.Companion,
+            new ElapsedTimeTrigger("device", Duration.Companion.INFINITE),
+            ["Trigger", new ElapsedTimeTrigger("device", Duration.Companion.INFINITE)],
+            ["Trigger", new ManualTrigger("device", "manual", "")],
```
Whathecode, line 65: You only need to register the abstract type once. I will fix it. :)

Whathecode (approved):
Excellent! Thanks!

Only one major comment, which I corrected. You only need to provide one instance to test abstract types; e.g., no need to add `[ "Trigger", ... instance ... ]` multiple times. The point is the test code needs an instance of `Trigger`, not of the concrete type, that is what the first parameter in the array does.

Formatting-wise I edited your additions to be in line with the rest of the code, but this is only a minor comment. I did not set up a linter for TS declarations to guide you in this.

Issue #185, opened 2020-10-22
Equality checking for `StudyProtocol`


Whathecode:
How is protocol 'identity' determined? The current `equals` method verifies whether all fields are equal, which can thus not be used to store different 'versions' of the same protocol.

How should `StudyProtocolRepository` verify whether an updated snapshot is the same protocol than a previous version?

Likely, identity is best defined by the owner and creation date. But, we need to investigate. 

Alternatively, maybe we need a proper UUID assigned upon creation.

Whathecode:
I remember now. From a user perspective the name (and the user who created the protocol) is the identifier.

That doesn't mean that the name can't be changed, though. It simply means that the name has to be unique (i.e., why would there ever be two protocols with the same name but not different version?).

Likely there is still room for enhancement in the API to make that identity more explicit.

Whathecode:
Given that study protocols are aggregate roots which are linked to from different places, even subsystems (the study subsystem), a UUID would really be more suitable here.

Right now, changing the name of a protocol would require updating the name in all places where it is used (i.e., studies) as well since otherwise they are no longer linked.

I guess this is why DDD practitioners always recommend UUID as ID for aggregate roots. :)

Whathecode:
After another meeting with @ltj in which a similar issue came up, we concluded the best way forward is really to use a `UUID` for `StudyProtocol`.

PR #338, opened 2021-12-06
Give cues to infrastructure on integration event ordering

### LINKED ISSUE #230: Document `EventBus` implementation requirements for different subsystems.
We now use an event bus abstraction to communicate across application services. For example, between `StudyService` and `ParticipantService` (https://github.com/cph-cachet/carp.core-kotlin/blob/develop/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/ParticipantServiceHost.kt#L32).

The current implementation of this event bus currently does not dictate whether events may be received out of order. In the example listed above, there is an expectation by the consumer (`ParticipantService`) events from `StudyService` are received in order. Otherwise we may get runtime exceptions on the `checkNotNull` calls.

This particular integration seems easiest to rely on in-order delivery of events. Both services live in the same subsystem as well, so this does not seem like a farfetched requirement (they are expected to be deployed side-by-side, or even on the same machine). However, for other communication of services, such as across subsystems, this might be a big demand on the event bus implementation which reduces scalability.

Can we enforce one implementation or another, or make a correct implementation more likely? At a minimum, this should be documented.

Whathecode:
Events related to aggregate roots (ARs) should be handled _in order_, otherwise business logic checks will fail at runtime. To both document this, and not require _all_ events to be ordered, this PR introduces scopes to `IntegrationEvent` using the `aggregateId` for which the infrastructure is expected to guarantee ordering of events.

The current use case is always to pass the AR ID to `aggregateId`; all current integration events are related to aggregate roots. We also expect this to cover all future use cases.

This PR also unifies the concept of `id` on `AggregateRoot`. Wherever ARs were previously identified by other identifiers (e.g., since there is a one-to-one mapping of `Recruitment` to `StudyDeployment`, `Recruitment` was identified by `studyDeploymentId`), they now have their own unique ID. This is important, since otherwise events for those ARs sharing the same ID would need to be _ordered_ (since they share the same `aggregateId` in `IntegrationEvent`). This defeats the purpose of ARs.

For consistency, I also moved `createdOn` to the AR constructor, similar to how I introduced `id`. This provides better encapsulation (albeit `createdOn` being protected was a minor issue).

Lastly, to better disambiguate between `get` functions, some repository and service get functions were renamed.

With these changes in place, we should now be ready to change the notion of equality for study protocol in a subsequent PR: #185


carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/services/IntegrationEvent.kt
```diff
@@ -3,16 +3,23 @@ package dk.cachet.carp.common.application.services
...
-abstract class IntegrationEvent<out TApplicationService : ApplicationService<out TApplicationService, *>>
+interface IntegrationEvent<out TApplicationService : ApplicationService<out TApplicationService, *>>
+{
+    /**
+     * All events related to the same aggregate ID are handled in order.
+     *
+     * In case the event pertains to an aggregate root,
+     * specify its ID to ensure correct handling of business logic in the domain.
+     */
+    val aggregateId: String?
```
ltj, line 24: Not sure what to make of the nullable `aggregateId` - If for some reason it is `null` when an event is passed to the event bus, does it mean that ordering is not important for this event and any (fallback) key can be used _or_ should it be treated exceptionally and result in an error?

Whathecode: Well-spotted!

The intent was to convey ordering is not important, but I see that is not properly documented. Neither would treating `null` as an `aggregateId` be desirable: we don't want to guarantee ordering for all integration events which pass `null` here.

I think I'm happy with passing `null` here for the API, unless you can come up with a better suggestion. So my suggestion is to simply make this clear in the documentation.

ltj: I'm onboard with clearing up the documentation, since event constructors ensure non-null ids where applicable

ltj (approved):
🎉

PR #343, opened 2022-01-04
Replace `StudyOwner` and `ProtocolOwner` with `UUID`.

### LINKED ISSUE #214: Do not use `StudyOwner` in study subsystem application service endpoints
Similar to how we removed `ProtocolOwner` from the `ProtocolService` endpoints, we should remove `StudyOwner` from the `StudyService` endpoints and replace it with the `UUID`.

`StudyOwner` might become an aggregate root at some point down the line.

Whathecode:
Trivial PR. Simply removes `StudyOwner` and `ProtocolOwner` which were nothing more than a wrapper for `UUID`.

If `StudyOwner` or `ProtocolOnwer` ever becomes more (i.e., an aggregate root), it would _still_ be the `UUID` that should be used in the application services API.

Alternatively, these are _value objects_ we'd want to add other things than `id` to later, but that seems highly unlikely.

Closes #214


carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyDetails.kt
```diff
@@ -15,9 +14,9 @@ import kotlinx.serialization.Serializable
 data class StudyDetails(
     val studyId: UUID,
     /**
-     * The person or group that created this study.
+     * The ID of the person or group that created this study.
```
ltj, line 17: 'principal' or 'entity' over "person or group"

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyService.kt
```diff
@@ -31,16 +30,18 @@ interface StudyService : ApplicationService<StudyService, StudyService.Event>
 
 
     /**
-     * Create a new study for the specified [owner].
+     * Create a new study for the specified person or group with [ownerId].
```
ltj, line 33: 'principal' or 'entity' over "person or group"

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/Study.kt
```diff
@@ -8,25 +8,27 @@ import dk.cachet.carp.protocols.application.StudyProtocolSnapshot
...
 
 /**
- * Represents a study which can be pilot tested and eventually 'go live', for which a recruitment goal can be set, and participants can be recruited.
+ * Represents a study which can be pilot tested and eventually 'go live',
+ * for which a recruitment goal can be set, and participants can be recruited.
  */
 class Study(
     /**
-     * The person or group that created this [Study].
+     * The ID of the person or group that created this [Study].
```
ltj, line 21: 'principal' or 'entity' over "person or group"

ltj (approved):
yes. One thing we could consider is a general rename of `ownerId` - since this is really an access scope id

Whathecode:
> yes. One thing we could consider is a general rename of `ownerId` - since this is really an access scope id

This is the _domain_ model. Access scopes are an infrastructure concern. The question is which requirements, using the domain language, you _scope_ the access to. I think the closest to that is still the "owner" of the study.

> 'principal' or 'entity' over "person or group"

Which meaning of "principal" are you thinking of here? A simple dictionary lookup doesn't yield anything relevant for me. "Entity" could work, but what exactly do you intend that isn't encompassed by "person or group"? Organization? I picked "group" since not all groups are organizations, but an organization, to me, could be considered a group.

ltj:
> This is the _domain_ model. Access scopes are an infrastructure concern. The question is which requirements, using the domain language, you _scope_ the access to. I think the closest to that is still the "owner" of the study.

well, I'd argue that it boils down to access regardless. Owner also implies an entity outside the control of the domain. But I'm fine with it as long as the documentation reflects that this is not limited to a person or a "group".

> Which meaning of "principal" are you thinking of here? A simple dictionary lookup doesn't yield anything relevant for me. "Entity" could work, but what exactly do you intend that isn't encompassed by "person or group"? Organization? I picked "group" since not all groups are organizations, but an organization, to me, could be considered a group. Or, do you envision computer-generated studies and computer system owners? I can't think of realistic scenarios of the latter.

https://en.wikipedia.org/wiki/Principal_(computer_security)

PR #346, opened 2022-01-08
Allow changing `StudyProtocol` name and description

### LINKED ISSUE #185: Equality checking for `StudyProtocol`

Whathecode:
In a previous PR, it was determined that the _identity_ of `StudyProtocol` should be a `UUID`, as opposed to how it was originally defined: a composite key of study owner and study name.

The changes in this PR use that change to allow changing a protocol's name. In addition, I made `description` mutable, as I don't recall why I ever decided for it to be immutable.

But, like before, study names still need to be unique to owners. This is a constraint which needs to be implemented by the repository. Check the updated repository unit tests for guidance what is expected of concrete infrastructure implementations.

Closes #185


ltj (approved):
Looks fine. No problem regarding repo changes

PR #350, opened 2022-01-12
Add a display name to `DeviceRegistration`

### LINKED ISSUE #281: Display name for `DeviceRegistration`
A common use case will be to view the "type" of device a user is using as part of the study. In case of smartphones, for example "iPhone 12 (iOS 12.5)"

`DeviceRegistration` is designed to store this type of data. However, there is no common abstract field to _display_ it. Knowing the concrete type shouldn't be necessary to do so.

Whathecode:
The main reason for adding this is to display a concise representation to researchers in the study manager of the devices being used. E.g., "Pixel 6 Pro (Android 12)" for a `Smartphone`.

Even though manufacturers and Android operating system will probably be added as dedicated fields to a future `SmartphoneDeviceRegistration`, a general way of rendering _any_ device registration will be needed. The frontend shouldn't have to typecheck concrete `DeviceRegistration` types to render a useful column providing this information.

After some consideration, I made this field _optional_. Initially, it wasn't, leading me to having to add for example "Smartphone" as the default name for a Smartphone device. However, displaying fallback values is a frontend concern. While "Pixel 6 Pro (Android 12)" will work just fine in any locale, something like "Unknown device" won't.

In addition, I realized some mistakes I made in the current TS declarations. @ltj feel free to ignore those. @jeyjey626 could you have a look 👀 ?

Closes #281

### DIFF
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/DefaultDeviceRegistration.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/DefaultDeviceRegistration.kt
@@ -2,6 +2,7 @@ package dk.cachet.carp.common.application.devices
 
 import dk.cachet.carp.common.application.UUID
 import dk.cachet.carp.common.infrastructure.serialization.NotSerializable
+import kotlinx.serialization.Required
 import kotlinx.serialization.Serializable
 
 
@@ -11,20 +12,24 @@ import kotlinx.serialization.Serializable
  * The base class can't be made concrete since this would prevent being able to serialize extending types (constructors may not contain parameters which are not properties).
  */
 @Serializable
-data class DefaultDeviceRegistration( override val deviceId: String ) : DeviceRegistration()
+data class DefaultDeviceRegistration(
+    @Required
+    override val deviceId: String = UUID.randomUUID().toString()
+) : DeviceRegistration()
 
 
 /**
  * A default device registration builder which solely involves assigning a unique ID to the device.
  * By default, a unique ID (UUID) is generated.
- *
- * @param deviceId
- *   Override the default assigned UUID which has been set as device ID.
- *   Make sure this ID is unique for the type of device you are creating a registration for.
  */
 @Serializable( with = NotSerializable::class )
-class DefaultDeviceRegistrationBuilder( var deviceId: String = UUID.randomUUID().toString() ) :
-    DeviceRegistrationBuilder<DefaultDeviceRegistration>
+class DefaultDeviceRegistrationBuilder : DeviceRegistrationBuilder<DefaultDeviceRegistration>
 {
+    /**
+     * Override the default assigned UUID which has been set as device ID.
+     * Make sure this ID is unique for the type of device you are creating a registration for.
+     */
+    var deviceId: String = UUID.randomUUID().toString()
+
     override fun build(): DefaultDeviceRegistration = DefaultDeviceRegistration( deviceId )
 }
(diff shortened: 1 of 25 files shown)


typescript-declarations/@types/carp.core-kotlin-carp.deployments.core/index.d.ts
```diff
@@ -38,25 +38,25 @@ declare module 'carp.core-kotlin-carp.deployments.core'
             abstract class NotDeployed
```
jeyjey626, line 38:
```suggestion
            abstract class NotDeployed extends DeviceDeploymentStatus
```
This is something we could also fix with this PR, otherwise I'll add it in a later commit.

Whathecode: Thanks! But, I prefer to keep commits cohesive. I suggest you simply change this as part of a relevant commit on your future PR.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/DefaultDeviceRegistration.kt
```diff
@@ -11,20 +12,26 @@ import kotlinx.serialization.Serializable
  * The base class can't be made concrete since this would prevent being able to serialize extending types (constructors may not contain parameters which are not properties).
  */
 @Serializable
-data class DefaultDeviceRegistration( override val deviceId: String ) : DeviceRegistration()
+data class DefaultDeviceRegistration(
+    @Required
+    override val deviceDisplayName: String? = null,
+    @Required
+    override val deviceId: String = UUID.randomUUID().toString()
```
ltj, line 19: I think deviceId should come before display name. It would be more consistent with the other implementing classes

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/application/TestInstances.kt
```diff
@@ -49,7 +49,7 @@ val commonInstances = listOf(
     Smartphone( "User's phone" ),
 
     // Shared device registrations in `devices` namespace.
-    DefaultDeviceRegistration( "Some device" ),
+    DefaultDeviceRegistration(),
```
ltj, line 52: At first sight, it's not clear to me why you remove parameters when constructing `DefaultDeviceRegistration` everywhere in tests

Whathecode: Long story.

I was playing around/considering whether `deviceId` and `deviceDisplayName` should be optional or not, and that also impacted the "logical" ordering of them as parameters of `DefaultDeviceRegistration`.

From an API perspective, I concluded it makes more sense to have `deviceDisplayName` as the first parameter, and that in the majority of use cases you generate a `deviceId`.

This resulted in me having to update the tests several times, which I grew tired off. :) So, I settled on these parameters being optional, and non-essential for tests, which I committed in the first commit, to only then include `deviceDisplayName` without having to modify the tests.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/DefaultDeviceRegistration.kt
Whathecode: Interesting you should notice this. I went back on forth on this. As in the other PR comment, I concluded when using `DefaultDeviceRegistration` in the majority of use cases you want a generated `UUID`.

In essence, this is really a placeholder type to be replaced with a more concrete `DeviceRegistration` once we get round to modeling that, e.g., including manufacturer name, OS version etc. in a `SmartphoneDeviceRegistration`. Perhaps, at that point, we can only come up with a non-generated `deviceId` for smartphones.

Which does remind me again of potential privacy issues and need to deanonymize `DeviceRegistration` for anonymous exports. 🤔

ltj: oki doki

ltj (approved):
👍

PR #351, opened 2022-01-17
Add project to generate example JSON requests and responses

### LINKED ISSUE #233: Massive problems with Kotlin JSON serialization
I'm experiencing massive problems with the JSON coming out of the Kotlin core classes. Here are just some example.

The JSON for a [`MasterDeviceDeployment`](https://github.com/cph-cachet/carp.core-kotlin/blob/master/carp.deployment.core/src/commonMain/kotlin/dk/cachet/carp/deployment/domain/MasterDeviceDeployment.kt) looks like this:

```
{
    "deviceDescriptor": {
        "$type": "dk.cachet.carp.protocols.domain.devices.Smartphone",
        "isMasterDevice": true,
        "roleName": "phone",
        "samplingConfiguration": {},
        "supportedDataTypes": [
            "dk.cachet.carp.geolocation",
            "dk.cachet.carp.stepcount"
        ]
    },
    "configuration": {
        "$type": "dk.cachet.carp.protocols.domain.devices.DefaultDeviceRegistration",
        "registrationCreationDate": "2021-01-21T13:50:39.391Z",
        "deviceId": "114faa76-f00d-4665-b71b-d9b9508786d3"
    },
    "connectedDevices": [],
    "connectedDeviceConfigurations": {},
    "tasks": [
        {
            "$type": "dk.cachet.carp.protocols.domain.tasks.ConcurrentTask",
            "name": "task",
            "measures": [
                {
                    "$type": "dk.cachet.carp.protocols.domain.tasks.measures.PhoneSensorMeasure",
                    "type": "dk.cachet.carp.geolocation",
                    "duration": -1000
                }
            ]
        }
    ],
    "triggers": {
        "0": {
            "$type": "dk.cachet.carp.protocols.domain.triggers.ElapsedTimeTrigger",
            "sourceDeviceRoleName": "phone",
            "elapsedTime": 10000
        }
    },
    "triggeredTasks": [
        {
            "triggerId": 0,
            "taskName": "task",
            "destinationDeviceRoleName": "phone"
        }
    ],
    "lastUpdateDate": "2021-01-21T13:50:39.391Z"
}
```

This has several problems:

* There is inconsistency in using the `$type` -- some uses it, some don't and it's impossible to know when.
* The [`DataType`](https://github.com/cph-cachet/carp.core-kotlin/blob/master/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/data/DataType.kt) is serialized to a String, and not to the definition of a `namespace` and a `name` 
* [TriggeredTask](https://github.com/cph-cachet/carp.core-kotlin/blob/master/carp.protocols.core/src/commonMain/kotlin/dk/cachet/carp/protocols/domain/triggers/TriggeredTask.kt) does not in any way serialize as defined... where does the `triggerId`, `taskName`, etc. come from?
* The `Int` in the `triggers` map serialize as a `String` (e.g., `"0"`) and not as an integer.

And this is ONLY the problems with one class......

Whathecode:
Sets up a new project which:
- iterates over all `ApplicationService` implementations in the project
- retrieves the matching request objects
- serializes an example request and response for the request in JSON
- outputs it to `build/rpc-examples/<subsystem>/<SomeServiceRequest>/<someRequest>.json` and `<someRequest>-response.json`

In addition, I added JSON schemas for the entire API. I added a unit test which uses the available schemas to validate the generated JSON examples. Since these are generated, they represent the ground truth. In this way the JSON schemas can be manually maintained and tested as part of CI.

Note that this does not yet _publish_ the generated JSON examples and schemas. I may also look at how to include these files as a separate asset on GitHub releases.

This probably also closes #233


rpc/schemas/common/data/AngularVelocity.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "type": "object",
+  "properties": {
+    "$type": { "const": "dk.cachet.carp.angularvelocity" },
```
ltj, line 5: I think we decided on `__type` - should be consistently used here as well

Whathecode: Yup. Thought of that while adding `$type` everywhere. Unfortunately JSON Schema doesn't have variables, so I'll have to change it everywhere after.

Just didn't belong in this PR. Unit tests should catch any omissions when I change this.

ltj (approved):
Nice

PR #354, opened 2022-02-01
Preparation for versioned application services

Whathecode:
This PR prepares for setting up a versioning strategy for application service requests and integration events.

For now, it sets up unit tests per application service which replay the existing application service unit tests and capture all requests, events, and responses. This whole setup could already be reviewed at this stage.

It also already outputs these logged requests to the build folder.

The next step is writing unit tests that try to deserialize the previously generated test sources for specific _API versions_, and includes a migration proxy. If the request succeeds after migrating, it validates the migration proxy. Some more work will be needed to validate migration of responses and events, which will likely require injecting custom `DateTime.now()` and `UUID.random` generators so they are constant.


carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/DataStreamServiceLog.kt
```diff
@@ -0,0 +1,48 @@
...
+import dk.cachet.carp.data.application.DataStreamId
+import dk.cachet.carp.data.application.DataStreamService
+import dk.cachet.carp.data.application.DataStreamsConfiguration
+
+
+/**
+ * A proxy for a data stream [service] which notifies of incoming requests and responses through [log]
+ * and keeps a history of requests in [loggedRequests].
+ */
+class DataStreamServiceLog(
```
ltj, line 17: I think this really convinces me that it should be called `DataStreamServiceLoggingProxy` :)

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/infrastructure/services/ApplicationServiceRequest.kt
```diff
@@ -0,0 +1,13 @@
...
+import kotlinx.serialization.KSerializer
+
+
+/**
+ * A request for [TService] stored in memory, which can be invoked using [invokeOn].
+ */
+interface ApplicationServiceRequest<TService, out TReturn>
+{
+    fun getResponseSerializer(): KSerializer<out TReturn>
+    suspend fun invokeOn( service: TService ): TReturn
```
ltj, line 12: Thank you. This seems more right and is easier to reason about than the previous one via constructor delegation 👍

carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/infrastructure/DeploymentServiceLog.kt
```diff
@@ -0,0 +1,82 @@
...
+import dk.cachet.carp.deployments.application.users.ParticipantInvitation
+import dk.cachet.carp.protocols.application.StudyProtocolSnapshot
+import kotlinx.datetime.Instant
+
+
+/**
+ * A proxy for a deployment [service] which notifies of incoming requests and responses through [log]
+ * and keeps a history of requests in [loggedRequests].
+ */
+class DeploymentServiceLog(
```
ltj, line 20: Naming

carp.studies.core/src/jvmTest/kotlin/dk/cachet/carp/studies/infrastructure/versioning/OutputRecruitmentServiceTestRequests.kt
```diff
@@ -0,0 +1,21 @@
...
+import dk.cachet.carp.studies.application.StudyService
+import dk.cachet.carp.studies.infrastructure.RecruitmentServiceLog
+
+
+private val services = RecruitmentServiceHostTest.createService()
+
+class OutputProtocolServiceTestRequests :
+    OutputTestRequests<RecruitmentService>(
+        RecruitmentService::class,
+        RecruitmentServiceLog( services.first, services.third )
```
ltj, line 16: Documentation is needed when accessing generic tuple indexes like this I think.

rpc/src/main/kotlin/dk/cachet/carp/rpc/Main.kt
```diff
@@ -1,20 +1,33 @@
...
     // Create example requests for all request objects and responses of application service methods.
-    val exampleRequests = ApplicationServiceInfo.findInNamespace( "dk.cachet.carp" )
-        .associateWith { generateExampleRequests( it.serviceKlass, it.requestObjectKlass ) }
+    val exampleRequests = applicationServices
+        .associateWith { generateExampleRequests( it ) }
 
     // Output example requests.
-    FileUtils.deleteDirectory( JSON_EXAMPLES_FOLDER )
+    val jsonExamplesFolder = File( "build/rpc-examples/" )
+    FileUtils.deleteDirectory( jsonExamplesFolder )
```
ltj, line 27: You probably want to catch `IOException` here

ltj (commented):
Very nice work @Whathecode. Some general notes that may also have been addressed in my comments:
- General naming using the "Log" postfix. Think "LoggingProxy" is more appropriate
- (future) Concrete application service proxy implementations seems like something that can be generated. Vert.x service proxy codegen for inspiration https://vertx.io/docs/vertx-service-proxy/java/
- (future) Application service requests could potentially also be auto generated
- I noticed that the use of in-memory implementations is "hardcoded" everywhere. For shared test-sources we could consider making this pluggable, or otherwise just decide that we go for a test suite that replays the logged requests on any running instance of the infrastructure.

carp.studies.core/src/jvmTest/kotlin/dk/cachet/carp/studies/infrastructure/versioning/OutputRecruitmentServiceTestRequests.kt
Whathecode: Agreed. Replaced with data classes; a case where I do find self-documenting code is sufficient.

rpc/src/main/kotlin/dk/cachet/carp/rpc/Main.kt
Whathecode: I suspect this `rpc` application may become an entry point called as part of the Gradle build process to automate packaging test sources as part of releases.

If this fails, the build should fail. No clean way to handle the exception here for now. I think an exception by default results in an error code returned to the command line, which would be desired behavior. If the exception is unclear I can catch it to perhaps make it more meaningful, but will leave it as is for now.

PR #357, opened 2022-02-10
Backwards compatibility for application services

Whathecode:
Adds infrastructure support to transform old application service request objects to newer versions, and transform the returned responses to those expected by the caller.

The current state of this PR can generate test resources which when copied to the correct location can all be successfully replayed when `@Ignore` is removed for the tests in the base `BackwardsCompatibilityTest` class. I only intend to commit test resources once the 1.0 API stabilizes; otherwise they will cause tests to fail during intermediate updates, and we'd have to constantly update them as the API is still in flux.

Similarly, we will need transformation helpers to migrate integration events, but I consider this out of scope in this PR for now. This may need further discussion before proceeding.

### DIFF
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/services/ApiVersion.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/services/ApiVersion.kt
@@ -1,0 +1,12 @@
+package dk.cachet.carp.common.application.services
+
+
+/**
+ * Specifies the API version of an [ApplicationService].
+ * Within [major] versions, [minor] versions are backwards compatible.
+ *
+ * E.g. a 2.0 request will work on a 2.1 hosted service, but not on 1.0 or 3.0.
+ */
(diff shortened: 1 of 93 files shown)


carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/services/ApiVersion.kt
```diff
@@ -0,0 +1,25 @@
...
+annotation class ApiVersion( val major: Int, val minor: Int )
+{
+    companion object
+    {
+        fun toString( version: ApiVersion ) = "${version.major}.${version.minor}"
+
+        /**
+         * Determines whether [version] is more recent than [otherVersion].
+         */
+        fun isMoreRecent( version: ApiVersion, otherVersion: ApiVersion ): Boolean =
```
ltj, line 21: They are comparable by their string representation :)

Whathecode:
As discussed, I will merge this PR, but feedback would still greatly be appreciated and I will incorporate it in future PRs.

PR #359, opened 2022-03-01
Consistently use `statusList` for plural of `status`.

Whathecode:
image

The result of the naming poll was to simply use `status` for the plural of `status` in variables/functions/etc. However, this is not always possible and can thus not be applied _consistently_. E.g. the current application service method `DeploymentService.getStudyDeploymentStatus` needs to disambiguate from `DeploymentService.getStudyDeploymentStatusList`. We also have `RecruitmentService.GetParticipantGroupStatusList` which I personally would find odd to rename to `RecruitmentService.GetParticipantGroupStatus`.

We thus already use `List` suffixes throughout the codebase to prevent confusion. This PR applies that convention for all plurals of `status`.

Do we agree this is appropriate consistent naming?


ltj (approved):
Totally fine with that solution

PR #366, opened 2022-03-14
Add sensor timestamp constraints to measurement sequences

### LINKED ISSUE #355: Constraints on `sensorStartTime` and `sensorEndTime` of measurements within `DataStreamSequence`
## Start and end time ordering
`DataStreamSequence` does not constrain sensor start and end times added to sequences in any way. Is this correct?

This _feels_ wrong, since for a lot of typical data types such as geolocation and acceleration, we expect measurements to follow one another. This allows doing "naive" synchronization on the backend as data points come in simply by always applying the last sync point.

This is still "naive", since as sync points change, some new synced data may lie _before_ previously synced data. Our current thinking is simply to omit such data from the synced repository, as these are edge cases. Also, it would be "more wrong" to act as if these data points, which we _know_ have been collected at a later point in time, were collected before others. Later postprocessing can do smarter interpolation between sync points (e.g., spline interpolation), thereby maintaining all measurements and their order.

So, one benefit of this requirement is we can do a best effort real-time synchronization. If measurements are allowed to arrive out of order, a lot of data may very well be dropped erroneously for synced data since the original assumption does not apply. But, do we have any other reason to constrain "in order" measurements otherwise?

**Yes**; future back-end processing:

- Processed data streams:
  - Converting the current stepcount `TIME_SPAN` data which allows overlapping time spans intervals into a data stream of non-overlapping stepcount intervals.
  - Moving averages calculated on the backend. If we can't make any assumptions about which order data comes in, moving averages are pretty useless and constantly changing.
- Remote triggers may be arbitrary or at the whim of whichever order data is uploaded for such data streams. Consider how to implement the following anticipated triggers without any ordering assumptions:

image

- Real-time graph visualizations. Showing a graph of "incoming" data would be weird if "the past" changes over time.

In general, assuming the "past" (as determined by last received measurement) is immutable gives rise to many features which would otherwise become a nightmare to implement. 

**No**; flexibility edge cases:

- Back-end merging of data streams coming from different devices (e.g., multiple accelerometry patches); what if one device has been offline and hasn't uploaded data for a long time? Would this inhibit the stream processor merging multiple data streams until the device comes online again?
- Maybe some long-term measurements require long post-processing. What if the processing of one measurement of a more recent time interval is finished before that of a longer older one. Should we then prevent the client from uploading the data for which the measurement is already done simply because an older measurement is still in progress? This requires additional state management on the client.

**Conclusion**:

There are very few cases I can think of which require the ability to upload measurements out of order, and there are many potential benefits if we do enforce this requirement. It can be argued for the few cases in which it may be useful that the actual "measurement" (e.g., post-processing, or after the fact annotations) should be the point in time the values are specified/derived, and not the time to which those values apply. This is in line with the current `Measurement` documentation:

> The result of a measurement of [data] of a given [dataType] at a specific point or interval in time.

In that case, the `Data` object will probably contain the start/end time to which the measurement applies, even though it was "taken" at a different time. This may be a specific subtype of `Data` we can provide generic support for at a later point in time once requirements are better understood.

## Data time types

A measurement is said to be of `DataTimeType.POINT` when only start time is set, and of `DataTimeType.TIME_SPAN` when end time is also set.

We also have `DataTimeType.EITHER`, with the intent to denote a data stream that can contain both points and time spans. We have no real use cases in mind for this, and it is currently only used by "free form text":

> Text of which the interpretation is left up to the specific application.

It is worth noting the more idiomatic way in CARP would be to simply use application-defined data types for this and extending from `Data`. So this current data type is not the greatest example.

Should we drop it and require _all_ measurements in a stream to be either `POINT` or `TIME_SPAN`? For known data types we can validate based on what is _required_. For unknown data types (defined by the client) the first uploaded measurement automatically dictates the requirement.

The semantics of start and end time ordering are clear once we can assume there are only point or time spans in a data stream: 
- `POINT`: start time of subsequent points should always be > last point
- `TIME_SPAN`: end time should be > last end time; start time >= last start time

Once you mix and match, this becomes harder to reason about. It may be best to simply enforce these easier to understand semantics; especially if we can't think of any reasonable scenario in which you _would_ mix and match. The recommended approach would then simply be to split it into two data streams, so it still isn't impossible.

Whathecode:
- Removed `DataTimeType.EITHER` and `FreeFormText` which used it. For clarity I also removed `StubData` in tests, which used to be of `DataTimeType.EITHER`; I now mostly use `StubDataPoint` and `StubDataTimeSpan` where appropriate.

- Measurements in `DataStreamSequence` can now be validated by retrieving `DataTimeType` from `DataTypeMetaDataMap` and verifying time stamp ordering constraints:
  - `POINT`: start time of subsequent points should always be > last point
  - `TIME_SPAN`: end time should be > last end time; start time >= last start time

Closes #355 

Worth noting: adding a sequence to an existing `DataStreamId` in a batch which correctly lies after a previously added sequence ID but with timestamps that don't follow will make the sequence contained in the batch "invalid".


ltj (dismissed):
Ok for now

Whathecode:
In a tech meeting we agreed to make this type of verification _optional_. Thus, the current logic can become an external helper method operating on `DataStreamSequence`. It should not throw exceptions, but, instead, return information of any violations (in most basic form just `true`/`false`).

The backend infrastructure can then ignore any incoming sequences which don't validate and log these as errors. Or, can choose to still store these sequence but not incorporate them in synced data. In short, it allows for more elaborate fallback mechanisms and eliminates scope creep of `DataStreamSequence` which should really only be focusing on validating sequence IDs.

ltj (dismissed):

PR #369, opened 2022-03-21
Add `ACCELERATION` data stream to `Smartphone`

### LINKED ISSUE #356: Add acceleration including gravity data type
Right now we only have `NonGravitationalAcceleration`. Some devices do limited processing of acceleration measurements and still include gravity.

Whathecode:
Closes #356

### DIFF
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/data/CarpDataTypes.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/data/CarpDataTypes.kt
@@ -53,9 +53,15 @@ object CarpDataTypes : DataTypeMetaDataMap()
 
     internal const val NON_GRAVITATIONAL_ACCELERATION_TYPE_NAME = "$CARP_NAMESPACE.nongravitationalacceleration"
     /**
-     * Acceleration along perpendicular x, y, and z axes, excluding gravity.
+     * Change in velocity along perpendicular x, y, and z axes in the device's coordinate system, excluding gravity.
      */
     val NON_GRAVITATIONAL_ACCELERATION = add( NON_GRAVITATIONAL_ACCELERATION_TYPE_NAME, "Acceleration without gravity", DataTimeType.POINT )
+
+    internal const val ACCELERATION_TYPE_NAME = "$CARP_NAMESPACE.acceleration"
+    /**
+     * Change in velocity along perpendicular x, y, and z axes in the device's coordinate system, including gravity.
+     */
+    val ACCELERATION = add( ACCELERATION_TYPE_NAME, "Acceleration including gravity", DataTimeType.POINT )
 
     internal const val ANGULAR_VELOCITY_TYPE_NAME = "$CARP_NAMESPACE.angularvelocity"
     /**
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/data/CarpDataTypes.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/data/CarpDataTypes.kt
@@ -53,13 +53,13 @@ object CarpDataTypes : DataTypeMetaDataMap()
 
     internal const val NON_GRAVITATIONAL_ACCELERATION_TYPE_NAME = "$CARP_NAMESPACE.nongravitationalacceleration"
     /**
-     * Change in velocity along perpendicular x, y, and z axes in the device's coordinate system, excluding gravity.
+     * Rate of change in velocity, excluding gravity, along perpendicular x, y, and z axes in the device's coordinate system.
      */
     val NON_GRAVITATIONAL_ACCELERATION = add( NON_GRAVITATIONAL_ACCELERATION_TYPE_NAME, "Acceleration without gravity", DataTimeType.POINT )
 
     internal const val ACCELERATION_TYPE_NAME = "$CARP_NAMESPACE.acceleration"
     /**
-     * Change in velocity along perpendicular x, y, and z axes in the device's coordinate system, including gravity.
+     * Rate of change in velocity, including gravity, along perpendicular x, y, and z axes in the device's coordinate system.
      */
     val ACCELERATION = add( ACCELERATION_TYPE_NAME, "Acceleration including gravity", DataTimeType.POINT )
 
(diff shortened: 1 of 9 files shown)


carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/data/CarpDataTypes.kt
```diff
@@ -53,10 +53,16 @@ object CarpDataTypes : DataTypeMetaDataMap()
 
     internal const val NON_GRAVITATIONAL_ACCELERATION_TYPE_NAME = "$CARP_NAMESPACE.nongravitationalacceleration"
     /**
-     * Acceleration along perpendicular x, y, and z axes, excluding gravity.
+     * Change in velocity along perpendicular x, y, and z axes in the device's coordinate system, excluding gravity.
```
ltj, line 56: Rate of change of velocity over time = acceleration. Acceleration is still the accurate term here.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/data/CarpDataTypes.kt
```diff
@@ -53,10 +53,16 @@ object CarpDataTypes : DataTypeMetaDataMap()
...
      */
     val NON_GRAVITATIONAL_ACCELERATION = add( NON_GRAVITATIONAL_ACCELERATION_TYPE_NAME, "Acceleration without gravity", DataTimeType.POINT )
 
+    internal const val ACCELERATION_TYPE_NAME = "$CARP_NAMESPACE.acceleration"
+    /**
+     * Change in velocity along perpendicular x, y, and z axes in the device's coordinate system, including gravity.
```
ltj, line 62: Again, why call it change in velocity when it's acceleration just like stated by the type name.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/data/NonGravitationalAcceleration.kt
```diff
@@ -4,7 +4,7 @@ import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 
 /**
- * Holds acceleration data excluding gravity along perpendicular [x], [y], and [z] axes in meters per second squared (m/s^2).
+ * Holds change in velocity excluding gravity along perpendicular [x], [y], and [z] axes in meters per second squared (m/s^2).
```
ltj, line 7: Should be "Holds acceleration excluding..."

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/Smartphone.kt
```diff
@@ -55,20 +55,35 @@ data class Smartphone(
         val STEP_COUNT = add( NoOptionsSamplingScheme( CarpDataTypes.STEP_COUNT ) ) // No configuration options available.
 
         /**
-         * Acceleration along perpendicular x, y, and z axes,
-         * as measured by the phone's accelerometer and calibrated to exclude gravity and sensor bias.
+         * Change in velocity along perpendicular x, y, and z axes in the device's coordinate system, excluding gravity,
```
ltj, line 58: Again, acceleration is the better term

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/Smartphone.kt
```diff
@@ -55,20 +55,35 @@ data class Smartphone(
...
+         * as measured by the phone's accelerometer and calibrated to exclude sensor bias.
          * This uses the same coordinate system as the other sensors referring to x, y, and z axes.
          *
          * Android (https://developer.android.com/guide/topics/sensors/sensors_overview):
          * - This corresponds to `TYPE_LINEAR_ACCELERATION`.
          * - The sampling scheme's default measure interval corresponds to `SENSOR_DELAY_NORMAL` (200 ms).
-         * - The linear acceleration sensor always has an offset, which you need to remove: https://developer.android.com/guide/topics/sensors/sensors_motion
+         * - The acceleration sensor always has an offset, which you need to remove: https://developer.android.com/guide/topics/sensors/sensors_motion#sensors-motion-linear
```
ltj, line 65: acceleration sensor = accelerometer

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/application/TestInstances.kt
```diff
@@ -19,6 +19,7 @@ import kotlin.time.Duration.Companion.milliseconds
  */
 val commonInstances = listOf(
     // `data` namespace.
+    Acceleration( 42.0, 42.0, 42.0 ),
```
ltj, line 22: I still think it should be considered using constrained randomized values rather than the same constants forever. The unit should work with all values within a given range.

ltj (dismissed):
Approved but please consider my comments.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/data/CarpDataTypes.kt
Whathecode: Because defining a term using the term itself is circular logic. :) I was trying to add to the definition. If it is "rate of change of velocity" that's what I will change it into. Don't know where I found it stated it was simply "change in velocity". Will double-check.

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/application/TestInstances.kt
Whathecode: Here, that would be a horrible idea, since I think these values are even used to verify whether object creation is fully deterministic (always the same values).

For more fuzz testing I can look into that later. These `commonInstances` don't serve that purpose, though.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/data/CarpDataTypes.kt
ltj: These comments should not serve as definitions of the terms imho, but rather be specific about the contents if needed in this case that it is acceleration with or without a gravitational vector.

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/application/TestInstances.kt
ltj: As long as the values used to create the initial object are known to the assertions, I don't see the problem

PR #373, opened 2022-03-29
Include participant roles assigned to participant attributes and protocol devices

### LINKED ISSUE #283: Link `ParticipantData` to participant IDs
Currently, `ParticipantData` is stored _per deployment_ and it relates to the _entire_ deployment (e.g., "home address" of all participants in the deployment). But, a common use case will be to link it to a single _participant_ in that deployment (e.g., "birthdate"). For now we assumed this can be modeled by asking for multiple participant data types (e.g., "birthdate of father", "birthdate of child"), but then there is no strong reuse of standardized types anymore (e.g., "birthdate") as this requires introducing custom data types. 

Worth noting,  expected participant data is specified in the protocol, but currently we do not specify participants in the study protocol. Thus, this requires some thought on how to model participants as part of a study protocol before we can implement this.

Whathecode:
Closes #283 

- Optional participant roles can be defined in the `StudyProtocol`.
- Devices in `StudyProtocol` can be assigned participant roles.
- Invitations to studies are done using participant roles, rather than device role names.


carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/users/ExpectedParticipantData.kt
```diff
@@ -0,0 +1,66 @@
...
+)
+{
+    val inputDataType: InputDataType
+        get() = attribute.inputDataType
+}
+
+
+/**
+ * Determines whether the set contains any of the following conflicts:
+ * - contains differing participant attributes with the same input type
```
ltj, line 23: I think it should be clarified what 'type' means i.e. that it's more than just, say, a free-form text input.

ltj (approved):
You have my blessing

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/users/ExpectedParticipantData.kt
```diff
@@ -0,0 +1,66 @@
...
+ * - contains multiple attributes of the same input type which are assigned to the same role
+ *
+ * @throws IllegalArgumentException if [exceptionOnConflict] is set to true and the set contains a conflict.
+ */
+fun Set<ExpectedParticipantData>.hasNoConflicts( exceptionOnConflict: Boolean = false ): Boolean
```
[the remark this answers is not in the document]
Whathecode: Yeah. I had it on my radar. I even think I added it in a third place in the meantime. But, this would be internal refactoring that doesn't impact API, so later.

The advantage of the current approach is since `Set<ExpectedParticpipantData>` is used in serialized types, the serialized format is more sane than yet another class wrapper, without having to introduce some custom serializer. In other words, I agree, but postponed for now since it's not urgent.

carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/domain/users/ParticipantGroup.kt
```diff
@@ -179,28 +184,35 @@ class ParticipantGroup private constructor(
...
      */
-    fun setData( registeredInputDataTypes: InputDataTypeList, inputDataType: InputDataType, data: Data? ): Boolean
+    fun setData(
+        registeredInputDataTypes: InputDataTypeList,
+        inputDataType: InputDataType,
+        data: Data?,
+        /**
+         * The participant role [data] was assigned to; null if anyone can set the specified [inputDataType].
+         */
+        assignedToParticipantRole: String? = null
```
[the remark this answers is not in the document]
Whathecode: Worth considering. Right now, I tailor the API to the "default" notion of "single participant" studies. Either way, the API is a bit messy here. Ideal would be introducing the equivalent of `AssignedTo` but with a single `participantRoleName`.

This is a domain object, so not urgent for now. As long as it correctly enforces the current constraints.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/users/AssignedTo.kt
```diff
@@ -0,0 +1,23 @@
...
+ * Determines which participant roles to assign to something.
+ */
+@Serializable
+sealed class AssignedTo
+{
+    /**
+     * Assign this to anyone in the study protocol.
+     */
+    @Serializable
+    object Anyone : AssignedTo()
```
ltj, line 16: I wonder if this should be named `AssignedTo.All` since that is actually what it means as far as I can understand. Anybody could imply any _single_ one. Or we might need an additional type here since the semantics are potentially different between assigning to a single participant and all participants?

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/users/AssignParticipantRoles.kt
```diff
@@ -0,0 +1,31 @@
...
+import dk.cachet.carp.common.application.users.AssignedTo
+import kotlinx.serialization.Serializable
+
+
+/**
+ * Assign participant roles specified in a study protocol ([assignedRoles])
+ * to the participant identified by [participantId].
+ */
+@Serializable
+data class AssignParticipantRoles( val participantId: UUID, val assignedRoles: AssignedTo )
```
ltj, line 13: imho "assigned" would be more fitting

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/StudiesCodeSamples.kt
```diff
@@ -58,8 +58,8 @@ class StudiesCodeSamples
         // Once the study is live, you can 'deploy' it to participant's devices. They will be invited.
         if ( studyStatus.canDeployToParticipants )
         {
-            // Create a 'participant group' with a single participant, using the "Patient's phone".
-            val participation = AssignParticipantDevices( participant.id, setOf( patientPhone.roleName ) )
+            // Create a 'participant group' with a single participant, with no specific participant role assignment.
```
ltj, line 61: Maybe emphasize that the example protocol contains the primary smartphone device already and how this links to the participation created here.

ltj (approved):

PR #382, opened 2022-04-04
Release 1.0.0

Whathecode:
The first stable release of CARP core. The backend application services have been stabilized, and migration functionality is included so that older clients will still be able to call newer backends of the same major API level.

In addition, participants are now invited using _participant roles_, which dictate which devices they will be assigned to, and which data they are expected to input. This impacts various objects that previously held assigned device role names. (https://github.com/cph-cachet/carp.core-kotlin/pull/373) For the majority of studies that only contain a single participant role, no roles need to be specified, and `AssignedTo.All` can be used to have the same behavior as in older versions.

Dependency updates:

- kotlinx.serialization 1.3.2
- coroutines 1.6.0

### Common

- **Big rename**:
  - `TaskDescriptor` to `TaskConfiguration`
  - `Trigger` to `TriggerConfiguration`
  - Consistently use `statusList` for plural of `status` (https://github.com/cph-cachet/carp.core-kotlin/pull/359)
- Removed `FreeFormText`. This was a catch-all solution for extensibility, which was already possible by implementing custom `DataType`s.
- Functionality to migrate incoming API requests to the runtime version (https://github.com/cph-cachet/carp.core-kotlin/pull/367):
  - Each application service has a matching `ApplicationServiceApiMigrator`.
  - Call `migrateRequest` for incoming JSON requests, which can be old versions, to retrieve a `MigratedRequest` which holds the runtime version in `request`.
  - Call `MigratedRequest.invokeOn` to call the application service an return the JSON response in the version expected by the caller.
- Add `ACCELERATION`, including gravity, to `Smartphone`. (https://github.com/cph-cachet/carp.core-kotlin/pull/369)
- `Username` serialized as string primitive in JSON, rather than object (only used in `UsernameAccountIdentity`).
- Renamed `RRInterval` to `InterbeatInterval`, which is more general since not all interbeat intervals need to be based on RR peaks (https://github.com/cph-cachet/carp.core-kotlin/pull/380/commits/4083b64eb034ea08efc9738945e1a417c87214e4).

### Protocols

- Optional participant roles can be defined in `StudyProtocol`, and devices can be assigned to participant roles. By default, devices are assigned to "all" roles, meaning all participants in case no roles are specified. (https://github.com/cph-cachet/carp.core-kotlin/pull/373)
- Expected participant data now includes an optional participant role the `ParticipantAttribute` applies to. The equivalent of the old behavior is using `AssignedTo.All`, in which case the data can be edited by all participants in the deployment. When assigning data to one or more specific participant roles, the data can be set _per role_.

### Deployments

- Expected participant data is now included in `PrimaryDeviceDeployment`, so that clients have access to the necessary meta data to determine validity of data.
- `ParticipationService.setParticipantData` now includes a `inputByParticipantRole` parameter to set data assigned to a specific participant role.

### Studies

- Added `StudyService.RemoveProtocol`. (https://github.com/cph-cachet/carp.core-kotlin/pull/360)
- Added `studyProtocolId` to `StudyStatus`, which is `null` until study protocol is set. (https://github.com/cph-cachet/carp.core-kotlin/pull/365)
- `RecruitmentService.inviteNewParticipantGroup` now uses `AssignedParticipantRoles` rather than `AssignParticipantDevices`.
- `StudyService.setInternalDescription` can now set `description` to `null` to indicate no description is set (it is `null` to start out with).

### Data

- `DataStreamService.removeDataStreams` returns IDs of study deployments for which data streams were removed.
- Removed `DataTimeType.EITHER`: all data points in a data stream sequence should be `POINT` or `TIME_SPAN`, but not a mix of the two.
- Timestamps in sequences are now validated (https://github.com/cph-cachet/carp.core-kotlin/pull/366):
  - `POINT`: start time of subsequent points should always be > last point
  - `TIME_SPAN`: end time should be > last end time; start time >= last start time


pmasulli (approved):

ltj (commented):
🎉

PR #407, opened 2022-10-10
Interval sampling constraints

Whathecode:
For measures which support configuring an _interval_ at which to collect data, the introduced changes now support setting a fixed set of interval options in the corresponding sampling scheme.

In addition, while adding the `nearestOption` helper method to the interval sampling configuration builder, I noticed that the `Smartphone` sampling configuration builder did not yet expose accelerometry and angular velocity configurations.


thaiphandinh (approved):
Looks good to me!

pmasulli (approved):

PR #409, opened 2022-10-12
Add PPG measure data type

Whathecode:
Add a measure data type for photoplethysmogram data.

Considerations while choosing the current PPG format which includes data for multiple light sources, as opposed to a separate data type per light source:
- Some sensors use multiple LEDs: either different wavelengths of light (which corresponds to deeper penetration of the light into the skin), or at different locations to provide redundancy or light from different angles.
- The wavelength of LEDs for sensors is not always documented.
- Researchers interested in PPG are most likely interested in custom analysis/feature derivatives: e.g., heartrate or interbeat interval. Any _additional_ data which can help them perform a better analysis is likely valuable to them. Some wearables include multiple PPG channels. A common use case is to select the "best" signal out of multiple collected. Thus, when multiple channels are available, it makes sense to bundle samples collected approximately at the same time (sub-millisecond) together.
- There seems to be no agreed-upon unit for PPG data. But, it's very unlikely a given sensor would use multiple units for different channels.
- Some sensors allow turning data collection for individual light sources on/off.

Design decisions:
- The specific device should uniquely identify/name the light sources it collects data for. An early consideration was to use wavelength, but this would not allow to disambiguate between multiple LEDs of the same wavelength.
- PPG sensor specifications should at a later point in time be included in `DeviceRegistration` (https://github.com/imotions/carp.core-kotlin/issues/410):
  - The data is unitless, but not always. If the unit is known, it can be documented.
  - If wavelengths for LEDs are known, they can be documented.


pmasulli (approved):
looks good :)

PR #411, opened 2022-10-12
Allow adding sensor-specific data to measures

### LINKED ISSUE #293: Support adding "extra" infrastructure-specific data in data points
For example, stepcount currently only contains number of steps taken.

But, when stepcount is collected through the Apple Health API, it also carries a `deviceSource` field, which can be the phone, or for example a connected smartwatch.

How do we allow such extra data to be uploaded?

Whathecode:
Closes #293 

- Added `sensorSpecificData` to a new `SensorData` interface extending from `Data`.
- Updated `carp-common.md` documentation to reflect newly introduced `SensorData` and include better documentation for other data types. Be sure to [give this a proofread](https://github.com/imotions/carp.core-kotlin/blob/sensor-specific-data/docs/carp-common.md).
- Added migration to remove `sensorSpecificData` from `DataStreamService.getDataStream`. This wasn't caught by the current backwards compatibility tests since it is an optional field in JSON, but if a new backend has the data set, old clients calling `getDataStream` using the current default JSON serializer would fail. A specific test was introduced to test the migration. Additional functionality to implement this was also added to the migration builders. And, a backlog item was added to decouple the JSON transformation logic from the API migrations: #412 
- Added a reflection test to verify whether all implementations of `SensorData` set `sensorSpecificData` to `null`. Since this was was similar to `DeviceConfigurationReflectionTest`, a new common test class was created.


thaiphandinh (approved):
Looks good!

PR #418, opened 2022-11-26
Doc: adding a new connected device configuration

Whathecode:
Documentation on device configurations and how to add new connected device configurations. This is a more common use case than adding primary device configurations (like `Smartphone`) which I for now did not focus on, but in essence is the same, plus likely some additional considerations.


pmasulli (approved):
looks good!

PR #439, opened 2023-04-19
Fix: check for CustomData when serializing measurements

xelahalo:


Whathecode (approved):
Your changes broke a whole bunch of other unit tests. But, thank you for introducing the new unit tests which pointed out the problem!

As I suggested on Slack, the problem lies in the custom `MeasurementSerializer`, which overrides the default serializer behavior, which otherwise works perfectly. I therefore moved the fix there instead of an overloaded `getDataType`.

The `SerializerDerivedMethods` you modified that operate on the `Map` already took care of "unknown" types since unknown types wouldn't be registered in the map, in which case `getDataType` wouldn't be called. Similarly, you would only ever call `measurement` with types you have at runtime.

In addition, I made `CustomData` and other `Custom...` constructors internal to make clearer that these aren't types the client is supposed to initialize. I also rewrote the test slightly to reflect that, by which it becomes more representative of "the real runtime".

Whathecode (approved):

PR #441, opened 2023-05-05
Preparations for release 1.1.2

Whathecode:
- Stop publication of the helper project `publish-npm-packages`, which is rejected by Sonatype since it isn't configured for publication and made CD fail.
- Upgrade all dependencies: primarily Kotlin 1.8.21.

### DIFF
--- a/typescript-declarations/package.json
+++ b/typescript-declarations/package.json
@@ -13,19 +13,18 @@
     "carp-studies-core"
   ],
   "dependencies": {
-    "@js-joda/core": "5.2.0",
-    "big.js": "6.1.1"
+    "@js-joda/core": "5.5.3"
   },
   "devDependencies": {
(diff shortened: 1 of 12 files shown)


typescript-declarations/package.json
```diff
@@ -13,19 +13,18 @@
     "carp-studies-core"
   ],
   "dependencies": {
-    "@js-joda/core": "5.2.0",
-    "big.js": "6.1.1"
```
xelahalo, line 17: I noticed you removed this one, do we not need it?

Whathecode: Big JS no longer seems a dependency. I think it was used for the `long` implementation in kotlin standard library. Maybe they now use a different approach. All tests pass ...

xelahalo (approved):

PR #447, opened 2023-10-21
Upgrade to Kotlin 1.9.10 and other dependencies

Whathecode:
The new Kotlin version seems to require IntelliJ 2023.

And, as usual, the JS facades had to be updated; `serializer()` access through companion objects now also seems garbled, but luckily always to `.o14()`, so I could add a new facade to access serializers.

Lastly, a detekt upgrade resulted in some new formatting violations which were fixed.


Whathecode:
Kotlin 1.9.20 has been released, but updating to it currently is blocked by what seems to be a new TypeScript generation bug: https://youtrack.jetbrains.com/issue/KT-63184/JsExport-on-serializable-interface-creates-erroneous-TypeScript

For future reference, it also requires adding `it.freeCompilerArgs.add('-Xexpect-actual-classes') // https://youtrack.jetbrains.com/issue/KT-61573` to `build.gradle` where currently the other `optIn`'s live.

xelahalo (approved):
Yeah, most of it was just formatting, sorry for the late review!

Issue #451, opened 2023-11-08
Add Social Security Number as Input Data Type


bardram:
We need to support SSN - like the Danish CPR number. 

It should just be a string IMO.

Whathecode:
If all you need is a string, you could just add it as a [custom participant attribute](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/users/ParticipantAttribute.kt#L41) to the protocol.

As we discussed in person, there may be a requirement to still have full control over the data type name. I will check whether the `custom` suffix and uuid are essential for things to work or not. I think it may just be API convenience, in which case it already should work now if you provide the right JSON payload.

Whathecode:
The current constraints for participant data are managed upon protocol creation, and whenever the "expected participant data" is set. The logic for this [lives in `ExpectedParticipantData.kt`](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/users/ExpectedParticipantData.kt#L35):

> Determines whether the set contains any of the following conflicts:
> - contains differing participant attributes with the same input data type
> - contains multiple attributes of the same input data type which are assigned to the same role

As long as the input data type is unique, there should thus be no problem. When you set the expected participant data, you have to pass the `InputElement` by using `CustomParticipantAttribute` instead of `DefaultParticipantAttribute`. Default denotes the input element and validation is known at compile time; "custom" implies it isn't and thus needs to be specified by the one configuring the protocol.

I see that there is no integration test covering this, and thus also no example JSON requests generated for this. I will add this to clarify usage.

In addition, the Kotlin API currently doesn't allow to specify the `InputDataType` for `CustomParticipantAttribute`; it generates a unique ID in the format of `dk.cachet.carp.input.custom.9004a473ce624c49bcec34a238174118`. This is by design to protect the `dk.cachet.carp.input` namespace, but maybe is overly strict. The main goal here is to not pollute `dk.cachet.carp.input` with types for which support may be added later on, which would cause conflicts when comparing/handling data. E.g., if you use `dk.cachet.carp.ssn` now, but support gets added at with a stricter/different format. I think a suitable compromise is to allow you to set your own `InputDataType`, but restricting access to `dk.cachet.carp` sub-namespaces. @bardram Would this serve your need?

Whathecode:
[...]

PR #453, opened 2023-12-10
Add custom participant attribute API backwards compatibilty test

Whathecode:
Currently, there was no coverage for adding custom participant attributes to protocols in the backwards compatibility tests. The test resource also serves as an example on how such a JSON request looks like, which was a request in #451 .

Example of a protocol with both a default attribute (the server is expected to have the type information available at compile time), and a custom attributes (the type information is provided by the client in the form of an "input element").

```
{
    "id": "0d57f923-955e-4693-a606-cca9651aab4a",
    "createdOn": "2023-12-10T18:51:39.511538600Z",
    "version": 0,
    "ownerId": "27879e75-ccc1-4866-9ab3-4ece1b735052",
    "name": "Test protocol",
    "description": "Test description",
    "expectedParticipantData": [
        {
            "attribute": {
                "__type": "dk.cachet.carp.common.application.users.ParticipantAttribute.DefaultParticipantAttribute",
                "inputDataType": "dk.cachet.carp.input.sex"
            }
        },
        {
            "attribute": {
                "__type": "dk.cachet.carp.common.application.users.ParticipantAttribute.CustomParticipantAttribute",
                "input": {
                    "__type": "dk.cachet.carp.common.application.data.input.elements.Text",
                    "prompt": "Social Security Number"
                },
                "inputDataType": "dk.cachet.carp.input.custom.a1eda4bcd0d84ca0ba446cabfa4e84c8"
            }
        },
        {
            "attribute": {
                "__type": "dk.cachet.carp.common.application.users.ParticipantAttribute.CustomParticipantAttribute",
                "input": {
                    "__type": "dk.cachet.carp.common.application.data.input.elements.SelectOne",
                    "prompt": "Favorite OS",
                    "options": [
                        "Windows",
                        "Linux",
                        "Mac"
                    ]
                },
                "inputDataType": "dk.cachet.carp.input.custom.4ee5a0cc0ce5462b94baee4e59415f9b"
            }
        }
    ]
}
```

The protocol service does not validate `inputDataType`. Only when _setting_ the participant data on `ParticipationService`, errors will occur if no type information is available for the provided `inputDataType`. In the Kotlin APIs, the `inputDataType` for custom participant attributes is generated under the namespace `dk.cachet.carp.input.custom` with a GUID appended, but, this is currently no requirement for the API.

### DIFF
--- a/carp.protocols.core/src/commonTest/kotlin/dk/cachet/carp/protocols/application/ProtocolServiceTest.kt
+++ b/carp.protocols.core/src/commonTest/kotlin/dk/cachet/carp/protocols/application/ProtocolServiceTest.kt
@@ -1,7 +1,10 @@
 package dk.cachet.carp.protocols.application
 
 import dk.cachet.carp.common.application.UUID
+import dk.cachet.carp.common.application.data.input.CarpInputDataTypes
 import dk.cachet.carp.common.application.data.input.InputDataType
+import dk.cachet.carp.common.application.data.input.elements.SelectOne
+import dk.cachet.carp.common.application.data.input.elements.Text
 import dk.cachet.carp.common.application.triggers.TaskControl
 import dk.cachet.carp.common.application.users.ExpectedParticipantData
 import dk.cachet.carp.common.application.users.ParticipantAttribute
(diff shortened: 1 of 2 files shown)


bardram:
LGTM. Only comment / questions is whether it is necessary to use UUID for the custom `inputDataType`, like

```
"inputDataType": "dk.cachet.carp.input.custom.4ee5a0cc0ce5462b94baee4e59415f9b"
```

Can't I call them something useful, like 

```
"inputDataType": "dk.cachet.carp.input.custom.OperatingSystem"
```

bardram:
Another question is - when do you use lower versus upper case - what is the difference between these two?

```
  "attribute": {
                "__type": "dk.cachet.carp.common.application.users.ParticipantAttribute.DefaultParticipantAttribute",
                "inputDataType": "dk.cachet.carp.input.sex"
            }
```

Why is `sex` lower case and `DefaultParticipantAttribute` upper case?

bardram (approved):
See comments in PR

bardram:
Maybe it would also be a good idea to add an example of how to upload this custom participant data to the `ParticipationService`? Would you, for example, need to know these strange UUIDs?

Whathecode:
> Can't I call them something useful, like ...

You can, but as mentioned elsewhere, I recommend you to _not_ use CARP core namespaces to rule out any potential conflicts, and just to be clear about id scopes. That said, the `.custom` namespace is pretty safe as by design the goal is to only add guids there 

The reason they are generated by default is because when using an API exposed through a UI, you wouldn't have to directly interact with any ID. It is much more meaningful to display the prompt the user provided to help them identify which is which. In the API, you hold a reference to the object after deserializing, so no ID is needed either. That said, I am not opposed to an API which optionally allows you to set a custom ID.

Whathecode:
> Yes, but you receive these through the endpoint which tells you what data can be set.

I just noticed that although you do receive expected input elements on `PrimaryDeviceDeployment` through the `expectedParticipantData` property, this doesn't help a web interface as part of study management to _display_ these elements.

When you receive _known_ types and you have access to the CARP runtime, you can retrieve the statically defined input types. But, for custom types, you'd need access to the original protocol to discern the input elements (or derived types through which input elements are passed, such as `PrimaryDeviceDeployment`). This is undesirable coupling.

Incorporating support for retrieving `InputElement`'s as part of `ParticipationService` will make sense. Maybe they can optionally be requested as part of `getParticipantData`. Or, a new endpoint to retrieve input elements could be added. I'll give this some thought. Both seem backwards compatible.

Whathecode:
I moved the relevant tasks which emerged from this discussion to new issues, and will go ahead and merge this PR.

PR #455, opened 2024-01-31
Web compatible ts package

jakdan99:
The current TS sources don't work on browser targets. They seem to be targeting node. This PR fixes that.

Could not get it working with the `tsconfig.json` file, so I added and configured Babel to transpile the files into web compatible js files


Whathecode:
You mentioned on Slack you got the following error before this change:

> Uncaught Error: Dynamic require of "@cachet/carp-common-generated" is not supported

Can you elaborate what this actually means? When did this occur? What is your build environment, etc.? And. How do these changes fix that?

I looked a bit more into potential incompatibility with the browser given the current setup, and with my albeit limited knowledge, it should work as is. The current Kotlin build outputs UMD modules ([configured here](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/build.gradle#L90)). The wrappers written in TypeScript [also output UMD modules](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/typescript-declarations/tsconfig.json#L4). It should be possible [to consume these from the browser](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/global-plugin-d-ts.html#umd). How are you importing the library in React (I believe that's what you were using)? Are you certain you aren't trying to import it as an ES6 module?

The [latest version of Kotlin allows outputting ES6 modules](https://kotlinlang.org/docs/whatsnew19.html#experimental-support-for-es6-classes-and-modules), but that is still experimental and would bring along other bigger changes (seemingly it doesn't export namespaces). We could consider compiling to ES6 later, but would first like to understand the problem you are running into better.

Whathecode:
Rather than this PR, I'm [in the process of finding a nicer solution](https://github.com/cph-cachet/carp.core-kotlin/tree/typescript-changes). Pulling in more dependencies like babel should be unnecessary, and makes this project harder to maintain. From all I could read, this should work with latest versions of node, which can compile to ES modules and interop with CommonJS modules (which UMD modules expose).

I'm working on compiling the tests and wrappers/facades as ES modules, but they still import the generated UMD modules. As you can see in ongoing changes, [this does mean changing the way import/exports are currently done in the wrappers](https://github.com/cph-cachet/carp.core-kotlin/commit/c700b1c5d1158a456b883cd2be92cffc262d49d8). But, I think I got the hang of it, so may be able to wrap this up soon.

I'll create a new PR and give a shout here so you can try whether you have more luck with those sources.

Whathecode:
The alternate approach in #456 turns out to work, since this PR can be abandoned.

Issue #404, opened 2022-10-06
Add test to ensure valid JSON schema is available for all domain objects


Whathecode:
Currently, JSON schemas for all request objects are validated in `JsonSchemasTest`.

But, this does not include tests for all extending domain types.

An instance for all extending types is available in common test sources (`TestInstances`). And, a unit test verifies whether an instance is available for each of the domain types. If we could reuse these sources (#23), we could likely set up a test which verifies whether a schema for each is available, and which validates the schema.

PR #458, opened 2024-03-01
Add WebServer as a device

xelahalo:
A little background information:
We are trying to add a feature in @cph-cachet/carp-webservcies-spring, which would allow for a study to be hosted in an anonymous mode, meaning that participants would only receive a link in their inbox, and they wouldn't have to bother with the onboarding process. 

When we are generating the links with the help of our authorization service, we need to have a url that we can redirect users. At the time of creating the accounts and links we don't have any deployed groups and devices so I can really only rely on the protocol to contain some sort of information about where the study is hosted.

Do you think this is the right way to approach this @Whathecode? Or do you have any other ideas? Right now we would only be using anonymous invitations for ICAT, but in the future there may be other studies where we want to avoid using accounts and I also don't want to hardcode it on the CAWS level.

### DIFF
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/WebServer.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/WebServer.kt
@@ -1,0 +1,45 @@
+@file:Suppress("NON_EXPORTABLE_TYPE")
+
+package dk.cachet.carp.common.application.devices
+
+import dk.cachet.carp.common.application.Trilean
+import dk.cachet.carp.common.application.data.DataType
+import dk.cachet.carp.common.application.sampling.DataTypeSamplingSchemeMap
+import dk.cachet.carp.common.application.sampling.SamplingConfiguration
+import dk.cachet.carp.common.application.tasks.TaskConfigurationList
+import kotlinx.serialization.Serializable
+import kotlin.js.JsExport
+import kotlin.reflect.KClass
+
+typealias WebServerDeviceRegistration = DefaultDeviceRegistration
+typealias WebServerDeviceRegistrationBuilder = DefaultDeviceRegistrationBuilder
+
+@Serializable
+@JsExport
+data class WebServer(
+        val url: String,
+        override val roleName: String,
+        override val isOptional: Boolean = false
+) : PrimaryDeviceConfiguration<WebServerDeviceRegistration, WebServerDeviceRegistrationBuilder>()
+{
+    companion object
+    {
+        fun create( url: String, roleName: String ): WebServer = WebServer( url, roleName )
+    }
+
+    object Sensors : DataTypeSamplingSchemeMap()
+    object Tasks : TaskConfigurationList()
+
+    override fun getSupportedDataTypes(): Set<DataType> = Sensors.keys
+
+    override val defaultSamplingConfiguration: Map<DataType, SamplingConfiguration> = emptyMap()
+
+    override fun getDataTypeSamplingSchemes(): DataTypeSamplingSchemeMap = Sensors
+
+    override fun createDeviceRegistrationBuilder(): WebServerDeviceRegistrationBuilder =
+            WebServerDeviceRegistrationBuilder()
+
+    override fun getRegistrationClass(): KClass<WebServerDeviceRegistration> = WebServerDeviceRegistration::class
+
+    override fun isValidRegistration( registration: WebServerDeviceRegistration ): Trilean = Trilean.TRUE
+}
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/infrastructure/serialization/Serialization.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/infrastructure/serialization/Serialization.kt
@@ -63,6 +63,7 @@ val COMMON_SERIAL_MODULE = SerializersModule {
     {
         subclass( CustomProtocolDevice::class )
         subclass( Smartphone::class )
+        subclass( WebServer::class )
 
         subclass( CustomPrimaryDeviceConfiguration::class )
     }
--- a/carp.common/src/commonTest/kotlin/dk/cachet/carp/common/application/TestInstances.kt
+++ b/carp.common/src/commonTest/kotlin/dk/cachet/carp/common/application/TestInstances.kt
@@ -48,6 +48,7 @@ val commonInstances = listOf(
     BLESerialNumberDeviceRegistration( "123456789" ),
     CustomProtocolDevice( "User's phone" ),
     Smartphone( "User's phone" ),
+    WebServer( "https://example.com", "Example server" ),
 
     // Shared device registrations in `devices` namespace.
     DefaultDeviceRegistration(),
--- a/docs/carp-common.md
+++ b/docs/carp-common.md
@@ -70,13 +70,13 @@ _Primary_ devices ([`PrimaryDeviceConfiguration`](../carp.common/src/commonMain/
 in addition to supporting data collection from internal sensors,
 act as a hub to aggregate, synchronize, and upload incoming data received from one or more connected devices. 
 
-| Class                                                                                                                          | Primary | Description                                                                                              |
-|--------------------------------------------------------------------------------------------------------------------------------|:-------:|----------------------------------------------------------------------------------------------------------|
-| [Smartphone](../carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/Smartphone.kt)                     |   Yes   | An internet-connected phone with built-in sensors.                                                       |
-| [AltBeacon](../carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/AltBeacon.kt)                       |         | A beacon meeting the open AltBeacon standard.                                                            |
-| [BLEHeartRateDevice](../carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/BLEHeartRateDevice.kt)     |         | A Bluetooth device which implements a Heart Rate service.                                                |
-| [CustomProtocolDevice](../carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/CustomProtocolDevice.kt) |   Yes   | A primary device which uses a single `CustomProtocolTask` to determine how to run a study on the device. |
-
+| Class                                                                                                                          | Primary | Description                                                                                               |
+|--------------------------------------------------------------------------------------------------------------------------------|:-------:|-----------------------------------------------------------------------------------------------------------|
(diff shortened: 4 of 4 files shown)


Whathecode:
I can dive a bit deeper later, but wanted to give some quick initial feedback already now.

I think two requirements are confounded here:
1. The ability to invite yourself to a study, or in other words, "open studies". In this case, there is no need to send participants a _personalized_ link. You can redirect to a single URL, and the web server can upon request of the participant create a new deployment and proceed from there.
2. The ability to represent web pages which are supposed to be opened by the participant (as opposed to are opened by a study runtime, which we support through `WebTask`) as part of tasks in a study protocol.

We considered these use case during early design, and I'm fairly certain it was on a backlog somewhere. I suspect an iMotions backlog which also contained larger overarching design discussions (crossing core/front-end/back-end).

@ltj can maybe help refresh my memory.

For (1), I think we concluded this is mostly an _infrastructure_ concern, if not, my suspicion would be this should be managed somewhere in the _studies_ subsystem.

For (2), I _think_ we had a similar idea of exposing this as a "web server", or "web service". It makes sense, the "study runtime" in this case is a web server. It will have to know which deployment it is collecting data for, etc., though, so it will have to act as a full "client".

xelahalo:
1. Yeah I remember that this was the conclusion of our previous conversation, but for example in the case of ICAT it is not technically an "open" study, but instead a study where only a set number of participants should have access to it. This is like standard invitations, however, we don't really have access to their e-mail addresses, but instead the researchers would want to send these links to the participants' e-boks or digitalmail or some state maintained inbox in general. We don't really have the integration for e-boks, but even if we had that we would still want to avoid the onboarding process, and for that I somehow have to figure out where to redirect the users as it is not necessarily the case that the same server will initiate the invitation process. To work around this we have few options:

- when generating we take the url as paramereter to the endpoint
- we can lookup what the url is based on some study config ( when I looked at protocol examples, a new device seemed like a reasonable place to put this into )
- we can hack and hardcode the url for the specific study

I kind of want to avoid the last one. We can do the first one as a workaround for now, but it feels a little clunky. 

2. Okay 😃  ( I don't really get the "knowing which deployment it collects data for", don't deployments have connected devices? )
Thanks for the heads up for the schema file. I can also write the tests to check if they exist 😄

Whathecode:
> it is not technically an "open" study, but instead a study where only a set number of participants should have access to it. This is like standard invitations, however, we don't really have access to their e-mail addresses

I probably need to hear more about the exact use case, but this makes me believe this has more to do with an alternative to `EmailAccountIdentity`.

A user can retrieve study invitations based on their account ID. They get access to that by being authenticated using an account identity they have access to. Email is the predominant one CARP was designed for. But, you can see alternatives were considered in the form of `UsernameAccountIdentity`.

Maybe this is also relevant/related. 🤔

Whathecode:
@bardram @xelahalo I [created an issue for one part of what I understood is being asked for in this PR](https://github.com/cph-cachet/carp.core-kotlin/issues/459), in addition with some background of the original design/what is currently in place. Could you provide answers the the questions asked there?

And as per my first response, I think there are two asks here. The second relates to "website devices". I definitely think there is something there, but putting the `url` in the device configuration goes counter CARP's design. The equivalent would be putting a hardcoded BLE address of a specific Bluetooth device to use in a study in the study protocol. That hardly makes it reusable, as it pertains to exactly one _deployment._ You should consider the protocol to only _describe_ the study; suppose somebody else would want to replicate the study, but host it elsewhere.

The [logical place to put a URL would be the `DeviceRegistration`](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/DeviceRegistration.kt). Note that a specific study/backend can _pre-register_ devices. And, device registrations are returned to users [as part of retrieving their invitations](https://github.com/cph-cachet/carp.core-kotlin/blob/e058fc8cf394fbbf5eccd2b3c64439da8cb715cc/carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/application/ParticipationService.kt#L35).

Putting things together, I thus think the real ask here is to:
1. The ICAT application needs to be able to generate a unique URL the researcher can use send to participants, so all the user needs to do is click (username/password can do that, simply by embedding it in URL).
2. When a new URL is requested, ICAT needs to call the authentication service to create a new identity.
3. It should be possible to embed this identity/authentication in the ICAT URL being generated.
4. The ICAT application creates a deployment for an `ICAT protocol` an invites the newly created user [by this identity](https://github.com/cph-cachet/carp.core-kotlin/issues/459) (not email).
5. When the URL is visited, ICAT uses the information embedded in the URL to authenticate and get invitations, finds the ICAT invitation, and get the study deployment information (just like an ordinary CARP client). It can call `registerDevice` (maybe uploading things like browser used if it matters for stimuli presentation, etc.), call `deviceDeployed` and the study can start.

Whathecode:
@xelahalo Regarding JSON schemas, it seems like it is [at least documented in the checklist documenting how to add new device](https://github.com/cph-cachet/carp.core-kotlin/blob/e058fc8cf394fbbf5eccd2b3c64439da8cb715cc/docs/development-checklists.md#add-a-new-connected-device-configuration)s.

> Update JSON schemas for the new type:
Add a new schema in [rpc/schemas/common/devices](https://github.com/cph-cachet/carp.core-kotlin/blob/e058fc8cf394fbbf5eccd2b3c64439da8cb715cc/rpc/schemas/common/devices) corresponding to the class name (e.g., BLEHeartRateDevice.json).
Add the new schema as a subtype in DeviceRegistration.json or DeviceConfiguration.json. The existing examples should guide you, but double-check you specified the right type constant.
Warning: the presence or validity of this schema https://github.com/imotions/carp.core-kotlin/issues/404. It is recommended to serialize an instance of the new data type and [validate the output manually for now](https://www.jsonschemavalidator.net/).

bardram:
> And as per my first response, I think there are two asks here. The second relates to "browser devices". I definitely think there is something there, but putting the `url` in the device configuration goes counter CARP's design. The equivalent would be putting a hardcoded BLE address of a specific Bluetooth device to use in a study in the study protocol. That hardly makes it reusable, as it pertains to exactly one _deployment._ You should consider the protocol to only _describe_ the study; suppose somebody else would want to replicate the study, but host it elsewhere.

This is correct. 

However, you write:

> Putting things together, I thus think the real ask here is to:
> 
> 1. The ICAT application needs to be able to generate a unique URL which the researcher can use to send to a participant, so all the user needs to do is click (username/password can do that, simply by embedding it in URL).

I'm not sure this is what we want @xelahalo ? Generating anonymous (or "magic") links for participants apply for ALL studies, also non-ICAT ones. This is feature we also plan to use in our mobile sensing studies - the user would simply get an QR code and can start uploading data w/o need for the awful login procedure (the largest usability problem we have i all our studies is that users (think: very old woman) can't log in).

So - CAWS should be able to generate such anonymous accounts at "its core".

> 
> If the goal is to make a more generic solution in CAWS, a CAWS front-end could create participant URLs in the same vain. 

This is exactly the goal and we have already designed this "front-end" where researchers can generate a number of anonymous accounts.

xelahalo:
> And as per my first response, I think there are two asks here. The second relates to "browser devices". I definitely think there is something there, but putting the url in the device configuration goes counter CARP's design. The equivalent would be putting a hardcoded BLE address of a specific Bluetooth device to use in a study in the study protocol. That hardly makes it reusable, as it pertains to exactly one deployment. You should consider the protocol to only describe the study; suppose somebody else would want to replicate the study, but host it elsewhere.

@Whathecode I see this point, but then what about [WebTasks](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/tasks/WebTask.kt)? They are also serialized into the protocol, and they also contain a url.

> The [logical place to put a URL would be the DeviceRegistration](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/DeviceRegistration.kt). (Note that a specific study/backend can pre-register devices. And, device registrations are returned to users [as part of retrieving their invitations](https://github.com/cph-cachet/carp.core-kotlin/blob/e058fc8cf394fbbf5eccd2b3c64439da8cb715cc/carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/application/ParticipationService.kt#L35). This could be another path to pass a URL if needed; but it is unclear to me why, since ICAT should manage this.)

In the first mental draft of this pr I also considered this, but then I would only be able to access what I need once I actually have a studyDeployment. For that I need

> 2. When a new URL is requested, ICAT needs to call the authentication service to create a new identity.

before I create the link, so that I have an identity that I can turn into a participant. But if I do that and let ICAT generate the url, then it is technically an open study as anyone would be able to get a link. If we, however, let a client with appropriate access control, like the portal, request the link generation, then what we basically have is my first point:

> - when generating we take the url as parameter to the endpoint

Since we still need to know which client/study to initiate the generation for. That means that I don't actually need a new device, or a new device registration.

Okay so I think my conclusion is that for now we will:

1. take the url as input for generating the link (we don't want ICAT to handle this, I think that would introduce unnecessary complexities @bardram)
2. rely on `UsernameAccountIdentity` for the identity
3. not create a new `DeviceConfiguration` or `DeviceRegistration`
4. take a look at #404 later, as I have a lot of tasks piled up already

@Whathecode as always, thanks for the input, much appreciated 😄

Whathecode:
From the `WebTask` documentation:

> Redirects to a web page which contains the task which needs to be performed.
> ... // The execution of the task is delegated to a web page, so this task uploads no data.

There is thus very limited integration with CARP's domain model, and the full description = of the task _is_ the URL. It was introduced to integrate with external survey websites, like Google Forms.

In contrast, ICAT (at least the impression I have) is intended to become a CARP client. We can move towards modeling the subtasks of ICAT as tasks, or at a minimum have a `ICATTask` with a typed data point to be uploaded on completion.

If you don't do this, there is really no point of integrating ICAT in CARP.

Whathecode:
> Since we still need to know which client/study to initiate the generation for. That means that I don't actually need a new device, or a new device registration.

Not certain I follow. It is definitely clear, however, you expect the CAWS portal to manage ICAT studies (context I was missing). That means that the CAWS portal needs to set up the URL indeed, instead of ICAT. My suggestion for the generic solution was you don't create study-specific URLs. You create CAWS URLs for _participants_, and when visiting the URL (hosted by CAWS webserver) you either present a study overview (theoretically, the participant could be part of multiple studies), or simply redirect to ICAT (if/assuming there is only one study).

You still need to configure the URL _somewhere_, though, and this will need to be done through CAWS then. My suggestion was to use the existing CARP endpoints for this. You can create a `Website` device with a `WebsiteRegistration`, and pre-register this to capture the ICAT url (with the _actual_ parameters for that specific participant) once the participants are invited.

And, ICAT will need to be CARP-aware (unless you want an implementation like `WebTask` which as I argue before, has little benefit). Concretely, ICAT needs to be able to receive the required authentication mechanism, and either assume there is only one deployment for that account, or, also receive a `deploymentId` to operate on. How to pass that is an infrastructure concern.

PR #465, opened 2024-03-07
Add Website as a device

xelahalo:
I also added a convenience URL classes and moved the things that was in WebTask to a common place. I tried to keep things backwards compatible but someone should check it out. I am also unsure about the schemas.


carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/WebBrowser.kt
```diff
@@ -0,0 +1,68 @@
...
+import kotlinx.serialization.Required
+import kotlinx.serialization.Serializable
+import kotlin.js.JsExport
+import kotlin.reflect.KClass
+
+
+@Serializable
+@JsExport
+data class WebBrowser(
+        val possibleTypes: Set<String>,
```
Whathecode, line 22: I had no clue what this was until I read the full code.

Documentation needed, and a rename. I suggest `supportedBrowsers`. But, there is also a design issue. If you have no reason to believe a given browser _wouldn't_ be supported (which should really be the predominant scenario if good web design is assumed), you'd have to provide an exhaustive list here. So in the _majority_ of use cases, you need to provide/know a full list, and even then, a study would fail early if new browsers arise, or they change name. Thus, instead you probably want `unsupportedBrowsers` and a default empty list.

In addition, this (and in `WebBrowserDeviceRegistration`) needs clear instructions on what to pass here. I think I previously suggested this likely relates to HTTP header info (and p.s., if that is the case, `URL` is probably too broad of an abstraction, and you want to limit this to HTTP).

That is reason enough to add a `Browser` type which encapsulates this knowledge for existing browsers (and allows parsing it from the information in the string), so API users don't need to look up what this is for each browser, but which still provides the option to pass `Other`, for forwards compatability. There is instant reuse across `possibleTypes`, and `browser` in the registration, so this isn't premature. And, in front-end representations (this is multiplatform, so also relevant for JS code), it will help showing e.g. pie charts.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/WebBrowser.kt
```diff
@@ -0,0 +1,68 @@
...
+@JsExport
+data class WebBrowserDeviceRegistration(
+        val browser: String,
+        val url: URL,
+        @Required
+        override val deviceDisplayName: String? = null
+) : DeviceRegistration()
+{
+    @Required
+    override val deviceId: String = url.hashCode().toString()
```
Whathecode, line 56: Why not just use the `url`?

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/tasks/WebTask.kt
```diff
@@ -19,45 +19,20 @@ data class WebTask(
...
-         * Identifies the condition, defined by the study protocol, which caused the [WebTask] to be triggered.
-         */
-        TRIGGER_ID( markup( "trigger id" ) )
-    }
-
-
     /**
      * Replace the variables in [url] with the specified runtime values, if the variables are present.
+     *
+     * Propagates construction of [url] to [URL]. Kept for backwards compatibility.
```
Whathecode, line 32: Don't _document_ implementation details. If relevant (I don't think it is here) use _comments_ for that. Consider documentation part of the API, it's supposed to _hide_ (encapsulate) implementation details.

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/application/devices/WebBrowserTest.kt
```diff
@@ -0,0 +1,57 @@
...
+            browser = "CHROME"
+            url = "https://www.example.com"
+        }.build()
+
+        assertEquals("CHROME", registration.browser)
+        assertEquals("https://www.example.com", registration.url.stringRepresentation)
+    }
+
+    @Test
+    fun registration_browser_and_uri_is_serialized()
```
Whathecode, line 26: I think I added tests for some of the early types as regression tests to double-check that base properties were serialized. To a large extent, I think these existing tests suffice, and from this point onwards, we can rely on the serialization framework to work as expected.

I would thus suggest removing this; otherwise we'd have to add this fairly boiler-plate code (which I'm fairly certain provides no extra coverage) to each extending type.

To make that clear, probably the existing type-specific tests should somehow convey they are more "generic" in the nature of what they test. :) You can ignore that for now.

rpc/schemas/common/devices/WebBrowserDeviceRegistration.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "type": "object",
+  "allOf": [ { "$ref": "DeviceRegistration.json#DeviceRegistration" } ],
+  "properties": {
+    "url": { "type": "string", "format": "uri" },
+    "browser": { "type": "string" }
+  },
+  "required": [ "macAddress" ],
```
Whathecode, line 9: Copy/paste error. Also shows you haven't run this through a checker manually. I think the documentation suggests to serialize, and then manually check the serialized form against this schema. There are plenty of online validators you can use.

Whathecode (changes requested):

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/URL.kt
```diff
@@ -0,0 +1,101 @@
...
+         */
+        TRIGGER_ID(markup("trigger id"))
+    }
+}
+
+/**
+ * Regular expression to match any URI according to
+ * [Appendix B of RFC 3986/STD 0066](https://www.rfc-editor.org/rfc/rfc3986#appendix-B)
+ *
+ * TODO: [URIRegex] allows us to write our own validation, but for the most part, what we have now is sufficient.
```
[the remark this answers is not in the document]
Whathecode: And, while looking deeper into this regex, there are two more issues:
1. This is not a regex to _validate_ the URL, but to extract the 5 components. The RFC you link explicitly calls it a "non-validating parser".
2. If it would actually be a regex to validate URLs, it would fail for expected input to this class, namely the URL variables. They are formatted as `{{VAR}}`, and `{` and `}` are not valid characters, which is exactly why they are a good pick for injecting variables. :)

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/URL.kt
```diff
@@ -0,0 +1,101 @@
...
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.Serializable
+import kotlin.js.JsExport
+
+/**
+ * Represents a Uniform Resource Locator (URL) according to RFC 3986. We use [URIRegex] to match
+ * [stringRepresentation], but it cannot do actual validation. Therefore, [DefaultURLValidator] is used to validate
+ * the URL.
+ *
+ * The URL of the web page which contains the task to be performed.
```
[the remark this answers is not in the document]
Whathecode: But, the time when the variable is represented and when it is translated to an actual URL, is different. For a `WebTask`, the variable placeholder is represented as such in the study protocol, and replaced by the client runtime when the task is triggered.

For device registrations, you may want to store a URL on the studies subsystem front-end somewhere, and after each invitation construct a URL to be used in the `WebBrowserDeviceRegistration` when pre-registering the device. But the variable placeholder is never stored in any of the currently designed CARP objects.

So, if you want to extract anything, it would probably be some type of `ParameterizedWebUrl`, which could encapsulate the `markup` functionality, and associated URL + variable validation, based on the "expected" variables.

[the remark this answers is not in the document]
xelahalo: I think I want to refrain from doing something like a `ParameterizedWebUrl`. I guess I will just use `URL` to have a validated url and nothing else. And that will also avoid issues with things like the following.

> If it would actually be a regex to validate URLs, it would fail for expected input to this class, namely the URL variables. They are formatted as {{VAR}}, and { and } are not valid characters, which is exactly why they are a good pick for injecting variables. :)

xelahalo:
I also decided to rename `WebBrowser` to `Website` as it made more sense to me; it felt weird to register a browser at a url, instead it makes more sense that a specific website has supported browsers and a URL at which it is hosted.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/URL.kt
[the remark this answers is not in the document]
Whathecode: If you want to validate, you'd probably still want to restrict it to a WebUrl (http/https) though. I have a hard time imagining what a client running on ftp would look like. :) And, since you link it to web browser requirements.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/WebBrowser.kt
Whathecode: `WebBrowser`. And given the name ambiguity, it also warrants rethinking the name for the device configuration. Sounds like `Website` may be more appropriate. The client runs on a website, which you access with the browser (device registration information).

rpc/schemas/common/devices/WebBrowserDeviceRegistration.json
Whathecode: Not the schema. An instance of the object serialized to JSON. 🙂

I just typically run/modify a unit test, place a breakpoint, and copy the serialized value.

Whathecode: Not the schema. An instance of the object serialized to JSON. 🙂

I just typically run/modify a unit test, place a breakpoint, and copy the serialized value.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/URL.kt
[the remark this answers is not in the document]
xelahalo: I think the requirements are not clear for now, so I dissected this pr into two, and made a draft ( #466 ) of the URL class. Maybe we could turn that into an issue. I make the registration take a string for now.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/Website.kt
```diff
@@ -0,0 +1,105 @@
...
+    override fun isValidRegistration( registration: WebsiteDeviceRegistration ): Trilean =
+        unsupportedBrowsers.contains( registration.browser ).not().toTrilean()
+}
+
+/**
+ * Describes a browser type which can be used to make [WebsiteDeviceRegistration]s.
+ */
+@Serializable
+@JsExport
+enum class Browser
```
Whathecode, line 55: This doesn't store the original value from the header, nor does the mapping.

Whathecode (changes requested):

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/Website.kt
Whathecode: So, this is the core requirement of what we should be storing: https://en.wikipedia.org/wiki/User-Agent_header

I updated the PR accordingly.

Whathecode:
Thank you for adding this @xelahalo !

To have something which requires no further discussion and can be merged now, I adjusted the PR as follows:
- I removed the `Browser` type. This was too naive an implementation. Instead, as suggested before, it is more important to focus on who sets this, and what this _should_ be set to. The [`User-Agent` HTTP header](https://en.wikipedia.org/wiki/User-Agent_header) is the best candidate here. Anything else just forces clients to do more parsing on their end, while at the same time losing information.
- Accordingly, I removed the `unsupportedBrowsers` feature entirely. Doing this properly would require adding a class which parses this header information and derives information, such as which browser it is. Only then does adding this to CARP core provide any real value. But, in addition, I think some additional analysis on which real-world use cases you would want to exclude based on this information is warranted before adding this as a feature (#468).

Whathecode (approved):

PR #466, opened 2024-03-11
Add URL

xelahalo:
I made this draft so that we can continue our discussion on the `URL` or `ParameterizedWebURL`. There are a few to things to discuss:

- validation of a URL (either in-built or caller defined)
- support for custom schemes ( like a deep link to a smartphone based study)
- creating support for parameterization of a url, such as in `WebTask`

Seems like a bigger task, should I create an issue instead?

### DIFF
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/URL.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/URL.kt
@@ -1,0 +1,35 @@
+package dk.cachet.carp.common.application
+
+import dk.cachet.carp.common.infrastructure.serialization.createCarpStringPrimitiveSerializer
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.Serializable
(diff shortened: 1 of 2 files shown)


Whathecode:
Based [on my last analysis](https://github.com/cph-cachet/carp.core-kotlin/pull/465#discussion_r1518565097), it seemed like there was no real potential reuse for `URL` validation: it couldn't be used by `WebTask`, since that needs to take the special "parameterized" syntax into account.

But, there is some potential reuse for a `ParameterizedWebURL`, including validation of the URL, and embedded parameters. Maybe generic, with an enum as type parameter which determines the parameters. And, which is URL-encoding aware, which will become relevant for "device role".

Static typing of URL in a `WebsiteDeviceRegistration` would be a "nice to have", but is different from the above, and the above sounds more relevant to me. I could have a go maybe this weekend. :)

Whathecode:
Closed as it sounds like the other parts of this PR which weren't merged need more clarity in terms of what they hope to achieve.

PR #472, opened 2024-04-12
Failing test for serializing "unknown types" on JS release

jakdan99:
Using the typescript declarations, encoding a `StudyProtocolSnapshot` which has an unknown type throws:
`IllegalStateException: Could not find a matching field for "polymorphicDiscriminator" on the JavaScript runtime.`

Repro steps:
New test in `./typescript-declarations/tests/carp-protocol-test.ts`


Whathecode:
Replaced by #474 which will include a fix.

Issue #473, opened 2024-04-12
Serializing "unknown types" on JS release throws exception


jakdan99:
Using the typescript declarations, encoding a `StudyProtocolSnapshot` which has an unknown type throws:
`IllegalStateException: Could not find a matching field for "polymorphicDiscriminator" on the JavaScript runtime.`

Repro steps:
#472

Whathecode:
Thanks a lot for the repro! 🙏 

Noteworthy: this works, [and is tested](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/carp.protocols.core/src/commonTest/kotlin/dk/cachet/carp/protocols/infrastructure/StudyProtocolSnapshotTest.kt#L81), on the JS target platform.

I suspect the problem is related specifically to _released_ sources, which get minified. Unknown polymorphism on the JS platform [relies on accessing internals of `kotlinx.serialization`](https://github.com/cph-cachet/carp.core-kotlin/blob/2db3c013754ecdf52a06d81068a1d806d7238d25/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/infrastructure/serialization/UnknownPolymorphicSerializer.kt#L69).

```
        // HACK: Modify kotlinx.serialization internals to ensure the encoder is not in polymorphic mode.
        //  Otherwise, `encoder.encodeJsonElement` encodes type information, but this is already represented in the wrapped unknown object.
        AccessInternals.setField( encoder, "polymorphicDiscriminator", null )
```

When the tests run with non-minified sources, however, this `polymorphicDiscriminator` field is renamed to something else.

This code was written quite a long time ago. We should check whether updates to `kotlinx.serialization` allow doing this _without_ the HACK. But, if this is not possible, we will need to ensure the name is fixed [by adding to the `kotlinx.serialization` facade](https://github.com/cph-cachet/carp.core-kotlin/blob/2db3c013754ecdf52a06d81068a1d806d7238d25/typescript-declarations/carp-kotlinx-serialization/index.ts).

PR #474, opened 2024-04-13
Fix JS sources

Whathecode:
The version was set incorrectly in JS sources.

And, a fix for "unknown polymorphic" serialization in JS released sources (#473), including a test which shows it now works.


jakdan99 (approved):
Just tested it on client, it works. Nicely done

Issue #431, opened 2023-02-13
Should `DeviceRegistration` include freeform specification data?


Whathecode:
There [is a current request](https://github.com/cph-cachet/carp.core-kotlin/pull/430) to include the following information for smartphone registrations:
```
        val platform: String? = null,
        val hardware: String? = null,
        val deviceName: String? = null,
        val deviceManufacturer: String? = null,
        val deviceModel: String? = null,
        val operatingSystem: String? = null,
        val sdk: String? = null,
        val release: String? = null,
```

But, as stated previously, such fields weren't added yet to the corresponding registration of [`Smartphone`](https://github.com/cph-cachet/carp.core-kotlin/blob/a0ee82f726d36e6229240efed8629cdf1e96012a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/Smartphone.kt) since [the use cases are unclear](https://github.com/cph-cachet/carp.core-kotlin/issues/410#issuecomment-1276177339):

> So far, it hasn't been included as separate more formalized fields since I did not make time to determine what these should be or where they should be defined. Are these additional fields on the abstract `DeviceRegistration` or on extending classes? E.g., for smartphones we want to know the device model and OS. But, is "device model" generic enough to be applied to the base class, and does it add value over the current `deviceDisplayName`?

Further questions I raised on Slack:

- Is `platform` very different from knowing the `operatingSystem`? Maybe it's sufficient to have a single field representing both, e.g. "Android 13" or "iOS 16.3" are likely equally meaningful to "Android" and "13" / "iOS" and "16.3". Which code is anticipated that would act on these fields separately (really the main line of thinking you need to pursue for anything modeled in core).
- Is `deviceManufacturer` more generic than `SmartphoneDeviceRegistration`? Maybe this is an (optional) field that could be added to the base `DeviceRegistration` class. Yet, again, wouldn't it make sense that `deviceModel` captures both the manufacturer and model of the phone in one field? What's the benefit of separating them?

As is, the current underlying requirement seems **"I want to log any device specifications"**, and this doesn't seem particularly related to something CARP-specific. If there are no CARP use cases acting on this data, I so no real need to define it statically (which also requires supporting backwards compatibility, etc...). To give an example, "I want to configure CARP deployments to only accept iOS phones", would be a CARP-specific requirement warranting such a static field.

[...]

Issue #169, opened 2020-10-08
Do we need to `encodeDefaults` in JSON serialization?


Whathecode:
While upgrading to `kotlinx.serialization` 1.0.0-RC2 the setting `encodeDefaults` for the `JSON` serializer changed from `true` to `false` by default.

This caused unit tests relying on `UnknownPolymorphicSerializer` to start failing. The unknown polymorphic serializer manually parses the JSON to extract expected fields, but these fields are no longer present when they have the default values. Therefore, [I have set `encodeDefaults` to true](https://github.com/cph-cachet/carp.core-kotlin/blob/56eff903d6b246dbf9ba66586e162547ccec7f42/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/serialization/Serialization.kt#L57).

Technically, it should be possible to ignore fields which are not present and presume they are set to default, if they have default values. For now, this did not seem important to worry about so I created this issue instead. Furthermore, maybe this is behavior that is supported by default if we use the new default polymorphic type handler of `kotlinx.serialization` instead of our custom solution: https://github.com/cph-cachet/carp.core-kotlin/issues/161

Whathecode:
Potentially relevant. [Introduced in `kotlinx.serialization` 1.3.0-RC](https://github.com/Kotlin/kotlinx.serialization/releases/tag/v1.3.0-RC), encoding defaults can now be configured on a _per-property_ level using `@EncodeDefaults`.

Whathecode:
Tests are failing for two different reasons:

- Missing `@Required` attributes on fields which should _always_ be serialized but had values assigned by initializers. These fields are needed since the backend does not necessarily have the concrete types available, and these fields are hence _not_ optional. Applying `@Required` solves that.
- The unknown serializers for `TaskDescriptor` and `DeviceDescriptor` did not take into account optional fields. By adding the optional fields to the internal `BaseMembers` classes (used to load the abstract fields), this is now resolved.

Issue #23, opened 2018-10-10
Move common test resources to a central location


Whathecode:
While adding tests for [the new `carp.deployment` submodule](https://github.com/cph-cachet/carp.core-kotlin/tree/deployment), I had to create test classes and functions to initialize common study protocols.

This is very similar to the test classes and objects in `carp.protocols`. Can these be moved to common test resources somehow? This might not be as straightforward due to the multiplatform setup (of course, creating yet another library would be an option, but this seems a bit excessive). Moving them to the existing `carp.test` library would not be possible since then a dependency on `carp.protocols` would need to be added to `carp.test`, resulting in a two-way dependency.

Whathecode:
Possibly related:

While implementing test classes for `carp.client.core` I now also had to implement a `InMemoryDeploymentRepository`. This could ideally be tested through the `DeploymentRepositoryTest` interface provide in `carp.deployment.core` tests.

How can this be referenced?

Whathecode:
Similarly, it should be possible to create a base class for `<ApplicationService>RequestsTest` for each of the application services.

Whathecode:
Being able to depend on test resources of a project within the same build would resolve this, but I am having an extremely hard time figuring out how to do this.

Posted on Stack Overflow: [How to add dependencies to tests of another project in a multi-platform multi-project Kotlin build](https://stackoverflow.com/questions/58956010/how-to-add-dependencies-to-tests-of-another-project-in-a-multi-platform-multi-pr)

Whathecode:
Since being able to depend on test resources of a project does not seem to have high priority, I already went ahead and moved common protocols test sources to the `infrastructure` namespace of carp.protocols: https://github.com/cph-cachet/carp.core-kotlin#stub-classes

There might still be merit in refactoring `<ApplicationService>RequestsTest` to a common base class, which currently is not possible since it relies on `kotlin.test` and aren't just some test stubs which are reused, independent from the test framework used. Therefore, I will leave this open for now.

Whathecode:
> Similarly, it should be possible to create a base class for `<ApplicationService>RequestsTest` for each of the application services.

Ran into this again while adding tests for `ParticipationService`. I think it is time to figure out which dependencies it has and maybe add a `carp.common.test` which depends on `carp.test` submodule to at least remove this duplication across all application services right now. It could test common DDD patterns.

Whathecode:
[...]

PR #484, opened 2024-09-25
Add additionalSpecifications field to DeviceRegistration

yuanchen233:
As discussed in #431, adding a `additionalSpecifications` field to `DeviceRegistration`. This allows to keep additional specification data when register device and access them later on in `StudyDeploymentStatus.deviceStatusList`

### DIFF
--- a/carp.common/src/commonTest/kotlin/dk/cachet/carp/common/infrastructure/serialization/DeviceRegistrationTest.kt
+++ b/carp.common/src/commonTest/kotlin/dk/cachet/carp/common/infrastructure/serialization/DeviceRegistrationTest.kt
@@ -1,9 +1,10 @@
 package dk.cachet.carp.common.infrastructure.serialization
 
+import dk.cachet.carp.common.application.ApplicationData
 import dk.cachet.carp.common.application.devices.DefaultDeviceRegistration
 import dk.cachet.carp.common.application.devices.DeviceRegistration
-import kotlinx.serialization.decodeFromString
-import kotlinx.serialization.encodeToString
+import dk.cachet.carp.common.infrastructure.test.makeUnknown
+import kotlinx.serialization.*
 import kotlin.test.*
 
 
@@ -15,12 +16,12 @@ class DeviceRegistrationTest
     @Test
     fun can_serialize_and_deserialize_device_registration_using_JSON()
     {
-        val default: DeviceRegistration = DefaultDeviceRegistration()
-
-        val serialized = testJson.encodeToString( DeviceRegistration.serializer(), default )
-        val parsed: DeviceRegistration = testJson.decodeFromString( serialized )
-
-        assertEquals( default, parsed )
+        val toSerialize = getTestCases()
+        toSerialize.forEach {
+            val serialized = testJson.encodeToString( DeviceRegistration.serializer(), it )
+            val parsed: DeviceRegistration = testJson.decodeFromString( serialized )
+            assertEquals( it, parsed )
+        }
     }
 
     /**
@@ -29,10 +30,12 @@ class DeviceRegistrationTest
     @Test
     fun unknown_types_are_wrapped_when_deserializing()
     {
-        val unknownRegistration = serializeUnknownDeviceRegistration()
-        val parsed: DeviceRegistration = testJson.decodeFromString( unknownRegistration )
-
-        assertTrue( parsed is CustomDeviceRegistration )
+        val toSerialize = getTestCases()
+        toSerialize.forEach {
+            val unknownRegistration = serializeUnknownDeviceRegistration( it )
+            val parsed: DeviceRegistration = testJson.decodeFromString( unknownRegistration )
+            assertTrue( parsed is CustomDeviceRegistration )
+        }
     }
 
     /**
@@ -41,18 +44,33 @@ class DeviceRegistrationTest
     @Test
     fun serializing_unknown_types_removes_the_wrapper()
     {
-        val unknownRegistration = serializeUnknownDeviceRegistration()
-        val parsed: DeviceRegistration = testJson.decodeFromString( unknownRegistration )
-
-        val serialized = testJson.encodeToString( parsed )
-        assertEquals( unknownRegistration, serialized )
+        val toSerialize = getTestCases()
+        toSerialize.forEach {
+            val unknownRegistration = serializeUnknownDeviceRegistration( it )
+            val parsed: DeviceRegistration = testJson.decodeFromString( unknownRegistration )
+            val serialized = testJson.encodeToString( parsed )
+            assertEquals( unknownRegistration, serialized )
+        }
     }
 
-    private fun serializeUnknownDeviceRegistration(): String
+    private fun getTestCases() = listOf(
+        // With null values.
+        DefaultDeviceRegistration(),
+
+        // With all nullable fields set.
+        DefaultDeviceRegistration(
+            deviceDisplayName = "Device name",
+            additionalSpecifications = ApplicationData( """{"OS":"Android 42"}""" )
+        )
+    )
+
+    /**
+     * Serialize [registration] as an unknown type using JSON.
+     */
+    private fun serializeUnknownDeviceRegistration( registration: DeviceRegistration ): String
     {
-        val registration = DefaultDeviceRegistration()
         var serialized = testJson.encodeToString( DeviceRegistration.serializer(), registration )
-        serialized = serialized.replace( "dk.cachet.carp.common.application.devices.DefaultDeviceRegistration", "com.unknown.CustomRegistration" )
+        serialized = serialized.makeUnknown( registration )
 
         return serialized
     }
--- a/rpc/schemas/common/devices/DeviceRegistration.json
+++ b/rpc/schemas/common/devices/DeviceRegistration.json
@@ -31,7 +31,8 @@
         "__type": true,
         "deviceId": { "type": "string" },
         "deviceDisplayName": { "type": [ "string", "null" ] },
-        "registrationCreatedOn": { "type": "string", "format": "date-time" }
+        "registrationCreatedOn": { "type": "string", "format": "date-time" },
+        "additionalSpecifications": { "$ref":  "../../common/ApplicationData.json" }
       },
       "required": [ "__type", "deviceId", "deviceDisplayName", "registrationCreatedOn" ]
     }
--- a/rpc/src/main/kotlin/dk/cachet/carp/rpc/GenerateExampleRequests.kt
+++ b/rpc/src/main/kotlin/dk/cachet/carp/rpc/GenerateExampleRequests.kt
@@ -187,6 +187,7 @@ private val bikeBeaconPreregistration = bikeBeacon.createRegistration {
     organizationId = UUID( "4e990957-0838-414c-bf25-2d391e2990b5" )
     majorId = 42
     minorId = 42
+    additionalSpecifications = ApplicationData( """{"Model": "AnyBeacon B42"}""" )
 }.setRegistrationCreatedOn( deploymentCreatedOn )
 private val phoneRegistration = phone.createRegistration {
     deviceId = UUID( "fc7b41b0-e9e2-4b5d-8c3d-5119b556a3f0" ).toString()
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/AltBeacon.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/AltBeacon.kt
@@ -2,6 +2,7 @@
 
 package dk.cachet.carp.common.application.devices
 
+import dk.cachet.carp.common.application.ApplicationData
 import dk.cachet.carp.common.application.Trilean
 import dk.cachet.carp.common.application.UUID
 import dk.cachet.carp.common.application.data.CarpDataTypes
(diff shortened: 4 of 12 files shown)


yuanchen233:
For `DeviceRegistrationTest`, original test cases were kept to cover situations when `additionalSpecifications` is null; same for `GenerateExampleRequests` for rpc, additionalSpecifications is only added to one of the deviceRegistration.

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/infrastructure/serialization/DeviceRegistrationTest.kt
```diff
@@ -48,6 +59,32 @@ class DeviceRegistrationTest
...
+    }
+
+    /**
+     * Create a default device registration with display name and additional specifications.
+     */
+    private fun createDeviceRegistrationWithAdditionalInformation(): DeviceRegistration
+    {
+        val deviceDisplayName = "Device name"
+        val additionalSpecifications = "value: some additional specifications"
+        return DefaultDeviceRegistration( deviceDisplayName, additionalSpecifications )
```
Whathecode: There's a bug here. Or, at least, this doesn't do what the code implies. The second parameter is not `additionalSpecifications`, but `deviceId`.

And, `deviceId` was purposefully added as the last parameter, as in the vast majority of cases you want to not specify that yourself. Maintaining it last would have made this less likely to happen. :)

Furthermore, this method is not really common to both places where it is used. The context is different, and in this case, inlining it and making it more specific to the exact thing being tested makes more sense.

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/infrastructure/serialization/DeviceRegistrationTest.kt
```diff
@@ -23,6 +24,16 @@ class DeviceRegistrationTest
         assertEquals( default, parsed )
     }
 
+    @Test
+    fun can_serialize_and_deserialize_not_null_additional_information()
```
Whathecode: Unfortunately we don't have support for parameterized tests. At least, back when I set up this project, there was no multiplatform test framework which provided that.

But, this would be more appropriate here. Check out how I reworked this to a poor man's version of the same. :) The downside is it doesn't report individual test results, but I think that is a small price to pay for maintainability/removing code duplication.

Consider it a potentially useful investigation to look at the current status of parameterized tests in kotlin multiplatform.

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/infrastructure/serialization/DeviceRegistrationTest.kt
```diff
@@ -56,4 +93,16 @@ class DeviceRegistrationTest
...
     }
+
+    /**
+     * Create a serialized device registration with an unknown type and additional information.
+     */
+    private fun serializeUnknownDeviceRegistrationWithAdditionalInformation(): String
+    {
+        val registration = createDeviceRegistrationWithAdditionalInformation()
+        var serialized = testJson.encodeToString( DeviceRegistration.serializer(), registration )
+        serialized = serialized.replace( "dk.cachet.carp.common.application.devices.DefaultDeviceRegistration", "com.unknown.CustomRegistration" )
```
Whathecode: Not that you could have known, but I've created `makeUnknown` test helpers before to help setting this  up as it become a recurring thing.

rpc/schemas/common/devices/DeviceRegistration.json
```diff
@@ -31,7 +31,8 @@
         "__type": true,
         "deviceId": { "type": "string" },
         "deviceDisplayName": { "type": [ "string", "null" ] },
-        "registrationCreatedOn": { "type": "string", "format": "date-time" }
+        "registrationCreatedOn": { "type": "string", "format": "date-time" },
+        "additionalSpecifications": { "type": [ "object", "string", "null" ] }
```
Whathecode: This can reference `ApplicationData.json` for consistency with other places where this is used.

rpc/src/main/kotlin/dk/cachet/carp/rpc/GenerateExampleRequests.kt
```diff
@@ -187,6 +187,7 @@ private val bikeBeaconPreregistration = bikeBeacon.createRegistration {
     organizationId = UUID( "4e990957-0838-414c-bf25-2d391e2990b5" )
     majorId = 42
     minorId = 42
+    additionalSpecifications = "model: 1234"
```
Whathecode: Similar to `applicationData = "{\"uiTheme\": \"black\"}"` and other "application data" fields for which examples are created here, let's make this a representative JSON string.

Whathecode (changes requested):
See comments; I already fixed those (you can resolve them after reading them).

In addition, I squashed three of the commits into one. Each commit ideally should pass all tests/be a cohesive whole. The last commit was a correct separate commit. 👍 I hope/presume you didn't have other changes on this branch, so I force pushed. Apologies if you did (as that can cause conflicts). Make sure to use `git pull --rebase`.

But, I see two other concerns before this can be merged:
- I don't recall 100% whether or not we need `@Required` on the field. I vaguely remember it may be needed for some scenario, maybe "unknown polymorphic serialization". I need to dig up some old issues/documentation.
- You didn't test whether `additionalSpecification` actually gets serialized as a JSON object (instead as a string with escaped `\"` chars, etc.). I put a breakpoint and apparently it doesn't. I suspect that is because the serializer used for that field is not "inherited" as you would expect it. I need to dig in to kotlinx serialization documentation to refresh my memory.

Whathecode:
Regarding `@Required`, I [found something relevant here](https://github.com/cph-cachet/carp.core-kotlin/issues/169#issuecomment-915341765) in context [of a fix I applied](https://github.com/cph-cachet/carp.core-kotlin/commit/cfdccc60a77e2e723d97cd97c00e864f8da52cd5).

> Missing `@Required` attributes on fields which should always be serialized but had values assigned by initializers. These fields are needed since the backend does not necessarily have the concrete types available, and these fields are hence not optional. Applying `@Required` solves that.

That makes sense for fields like `DeviceRegistration.deviceId`, and definitely `PrimaryDeviceConfiguration.isPrimaryDevice`, but it seems less relevant for something like `additionalSpecifications`, which I don't expect to ever have type-specific defaults.

Now I'm actually wondering whether `DeviceRegistration.deviceDisplayName` even ever needed `@Required`. It didn't seem like I considered/questioned that [in the related PR](https://github.com/cph-cachet/carp.core-kotlin/pull/350). I'm thinking _not_ applying `@Required` (as you did here) is correct, and adding it for `deviceDisplayName` may have been erroneous due to a naive mimicking of surrounding code.

And, regarding the `ApplicationDataSerializer` not kicking in, my intuition seems right:
> I suspect that is because the serializer used for that field is not "inherited" as you would expect it.

For this to work as expected, you would have to apply `@Serializable( ApplicationDataSerializer::class )` to all extending classes of `DeviceRegistration`. Try applying it to `DefaultDeviceRegistration` which is used in the test, and see the difference in the serialized output.

```
@Serializable
@JsExport
data class DefaultDeviceRegistration(
    @Required
    override val deviceDisplayName: String? = null,
    @Serializable( ApplicationDataSerializer::class )
    override val additionalSpecifications: String? = null,
    @Required
    override val deviceId: String = UUID.randomUUID().toString(),
) : DeviceRegistration()
```

This gives `... "additionalSpecifications":{"OS":"Android 42"} ...` as expected, instead of the erroneous ` ... "additionalSpecifications":"{\"OS\":\"Android 42\"}" ...`

That would be a bit unfortunate if there is no way around this. It would be worthwhile checking the kotlinx serialization documentation to see whether they have any (new) features which can avoid the need for doing this.

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/infrastructure/serialization/DeviceRegistrationTest.kt
yuanchen233: Now I know the hidden treasure, each time I visit `CreateTestObjects.kt`  I learn something new, as I totally lost track which subsystem I find them.

yuanchen233:
> In addition, I squashed three of the commits into one. Each commit ideally should pass all tests/be a cohesive whole. The last commit was a correct separate commit. 
>
Got it, I was expecting a 'squash and merge' when this branch is done, but saw a TODO warning when pushing. I didn't give much thought if those changes should be applied here after having fun reading that serialization issue


> 
> Now I'm actually wondering whether `DeviceRegistration.deviceDisplayName` even ever needed `@Required`. It didn't seem like I considered/questioned that [in the related PR](https://github.com/cph-cachet/carp.core-kotlin/pull/350). I'm thinking _not_ applying `@Required` (as you did here) is correct, and adding it for `deviceDisplayName` may have been erroneous due to a naive mimicking of surrounding code.
> 
You mentioned in this PR `deviceDisplayName` was initially _not optional_, then the `@Required` annotation makes sense there, initially at least. I tested with and with out the annotation, both passes all the tests we currently have. 

Regarding removing `@Required` from `.deviceDisplayName`, My concern is backward compatibility, especially JS target. As [Defaults are not encoded](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/basic-serialization.md#defaults-are-not-encoded), default null value will become undefined.

> And, regarding the `ApplicationDataSerializer` not kicking in, my intuition seems right:
> I suspect that is because the serializer used for that field is not "inherited" as you would expect it.
> 
> For this to work as expected, you would have to apply `@Serializable( ApplicationDataSerializer::class )` to all extending classes of `DeviceRegistration`. 
...
> That would be a bit unfortunate if there is no way around this. It would be worthwhile checking the kotlinx serialization documentation to see whether they have any (new) features which can avoid the need for doing this.

This actually makes sense, as how kotlinx.serialization works with generic type parameters/abstract class, a serializer is required for each type/subclass. Note we also have `@Serializale` annotation for each subclass of `DeviceRegistration`. If `@Serializable(with=..)` for a field can be inherited, I would expect subclass also inherite `@Serializable` form superclass (which is not the case).

One work around I can think of is creating a new datatype/class to handle `ApplicationData` instead of using `String`, then we can set and enforce serializer for that class. The annotation will then only needed once when we declare the class, and `ApplicationDataSerializer` will be used as the default serializer, instead of `String.serializer()`.

carp.common/src/commonTest/kotlin/dk/cachet/carp/common/infrastructure/serialization/DeviceRegistrationTest.kt
Whathecode: Yeah, there is some redundancy here. For more information, check this issue: https://github.com/cph-cachet/carp.core-kotlin/issues/23 Specifically: https://github.com/cph-cachet/carp.core-kotlin/issues/23#issuecomment-556017519

Whathecode:
> One work around I can think of is creating a new datatype/class to handle ApplicationData instead of using String, then we can set and enforce serializer for that class. The annotation will then only needed once when we declare the class, and ApplicationDataSerializer will be used as the default serializer, instead of String.serializer().

Good idea!

I think I considered it in the past. Maybe I didn't pursue it since without inheritance there was no clear benefit. I shortly looked for past considerations, but couldn't find any, [only the PR where it was introduced](https://github.com/cph-cachet/carp.core-kotlin/pull/263).

So, definitely worthwhile giving this a go! You may want to do that as a separate PR preceding this one. But, also fine as separate commit within this one. Whichever you prefer. :) I won't force push anymore without checking in with you; the branch is yours.

yuanchen233:
> Regarding removing `@Required` from `.deviceDisplayName`, My concern is backward compatibility, especially JS target. As [Defaults are not encoded](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/basic-serialization.md#defaults-are-not-encoded), default null value will become undefined.

I ran some tests and it is the case:

```
carp.deployments.core/src/commonTest/resources/test-requests/DeploymentService/1.0/DeploymentServiceTest/
createStudyDeployment_registers_preregistered_devices.json
Request #3 returned the wrong response. ==> expected: 

<{"deviceConfiguration":{...},"registration":{"__type":"dk.cachet.carp.common.application.devices.DefaultDeviceRegistration","registrationCreatedOn":"2022-04-04T15:03:22.341209Z","deviceDisplayName":null,"deviceId":"17124c6e-c8e0-457e-9b6b-f66860c4b666"},"connectedDevices":[{"__type":"dk.cachet.carp.common.infrastructure.test.StubDeviceConfiguration","roleName":"Connected"}],"connectedDeviceRegistrations":{"Connected":{"__type":"dk.cachet.carp.common.application.devices.DefaultDeviceRegistration","registrationCreatedOn":"2022-04-04T15:03:22.340210600Z","deviceDisplayName":null,"deviceId":"26d56d0e-2d78-4c5a-a428-bfd85d05da77"}}}> 
but was: 
<{"deviceConfiguration":{...},"registration":{"__type":"dk.cachet.carp.common.application.devices.DefaultDeviceRegistration","registrationCreatedOn":"2022-04-04T15:03:22.341209Z","deviceId":"17124c6e-c8e0-457e-9b6b-f66860c4b666"},"connectedDevices":[{"__type":"dk.cachet.carp.common.infrastructure.test.StubDeviceConfiguration","roleName":"Connected"}],"connectedDeviceRegistrations":{"Connected":{"__type":"dk.cachet.carp.common.application.devices.DefaultDeviceRegistration","registrationCreatedOn":"2022-04-04T15:03:22.340210600Z","deviceId":"26d56d0e-2d78-4c5a-a428-bfd85d05da77"}}}>
```

`"deviceDisplayName":null` is missing due to kotlinx serializer not generating default values. If we want to change this, I think it's best to create another issue and plan to fix it for the next major release version? As far as I remember we don't support backward compatibility cross major versions.

Whathecode:
That can easily be resolved with a version migration, so only a minor version bump is needed. 

But, definitely out of scope here. 👍

PR #485, opened 2024-10-04
Add ApplicationData datatype

yuanchen233:
As discussed in #484, introducing a new datatype to hold `applicationData`and set default serializer to `ApplicationDataSerializer`. This way we don't need to specify the serializer used everytime we use `applicationData`.

Changes:
1. New type `ApplicationData` holds a `String` field `data`.
2. In `ApplicationDataSerializer`, a _not null_ String value is expected for field `data`
3. All `applicationData` are now type of `ApplicationData?`
4. Update TypeScript wrapper

### DIFF
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/ApplicationData.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/ApplicationData.kt
@@ -1,0 +1,18 @@
+package dk.cachet.carp.common.application
+
+import dk.cachet.carp.common.infrastructure.serialization.ApplicationDataSerializer
+import kotlinx.serialization.Serializable
+import kotlin.js.JsExport
+
+
+/**
+ * Holds extra [data] which is specific to concrete applications, sensors, or infrastructure, and isn't statically
+ * known to the base infrastructure.
+ *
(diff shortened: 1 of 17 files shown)


Whathecode (changes requested):
The changes look good!

However, I figured this should target main, so this fix is available prior to your other PR (which otherwise still has a bugged commit).

Or, you fix the other PR by adding the custom serializer on the `String?`, which then in this PR gets removed again. But, I tend to avoid that type of multiple modifications if better prerequisites are known. 🙂

I can rebase/move stuff for you if you want. Other than code style and commit order, I think these PRs are good to be merged.

yuanchen233:
Sounds good. 

> I can rebase/move stuff for you if you want. Other than code style and commit order, I think these PRs are good to be merged.

Yes please, and thanks. I try to avoid doing interactive rebase here (at least for now), as there is a chance I mess up everything and creates more work for both of us.  🙂

Whathecode (approved):
I just rebased on develop and removed the use of `ApplicationData` in the commits which aren't yet merged to the develop branch.

I also slightly reworked the `ApplicationData` and `ApplicationDataSerializer` documentation to avoid redundancies. The type now already documents the purpose, so no real need to reiterate on that fully on the serializer.

PR #487, opened 2024-10-09
Make `deviceDisplayName` optional in `DeviceRegistration`

### LINKED ISSUE #486: Make `deviceDisplayName` optional in serialized `DeviceRegistration`
When `DeviceRegistration.deviceDisplayName` was introduced (#350), there was some discussion on whether or not it should be optional. I landed on making it optional, but, for some reason still kept the `@Required` attribute defined on it, which causes it to always be serialized, even if it is `null`.

It sounds like `null` makes a fine default field, which when the name isn't set, shouldn't be serialized.

The fact that it was set to `@Required` caused some confusion in another PR which introduced a new `DeviceRegistration` field (#484). If it's not needed, it should probably be removed. Doing so will require a version bump of APIs and adding an application service migration  to the services which return this type (for backwards compatibility of older clients). This task should thus be bundled with an API upgrade.

yuanchen233:
1. `deviceDisplayName` field is no longer required in `DeviceRegistration`
2. For `CustomDeviceRegistration`, `deviceDisplayName` is default to `null`
3. `ApiVersion` in `DeploymentServiceRequest` bumped to 1.3

Json file under `deployment/src/commonTest/recourses` were copied from `build` folder, but `DeploymentServiceRequest` under `rpc` was updated manually.

Two things to notice
1. The key-value pairs in Json request/respond are ordered. Therefore simply adding the remove field will not guarantee the request/respond to be identical to earlier api version. `JsonObject` in `kotlinx.serailization` is a `Map`, but the added field will by default append to the end. I added a helper function sorting the respond json with a hard-coded string list. Other approach will be appreciated.
2. Consider the following partial respond from [`DeploymentServiceRequest.GetDeviceDeploymentFor`](https://github.com/cph-cachet/carp.core-kotlin/blob/df5affe4a8379c521e01f62db22edd28d08589ad/carp.deployments.core/src/commonTest/resources/test-requests/DeploymentService/1.1/DeploymentServiceTest/createStudyDeployment_registers_preregistered_devices.json#L270)

```
"connectedDevices": [
    {
        "__type": "dk.cachet.carp.common.infrastructure.test.StubDeviceConfiguration",
        "roleName": "Connected"
    }
],
"connectedDeviceRegistrations": {
    "Connected": {
        "__type": "dk.cachet.carp.common.application.devices.DefaultDeviceRegistration",
        "registrationCreatedOn": "2022-04-04T15:03:22.340210600Z",
        "deviceDisplayName": null,
        "deviceId": "26d56d0e-2d78-4c5a-a428-bfd85d05da77"
    }
}
```
`Connected` here is the `roleName`, but it's used as key, which is a bit odd here as most of the keys are reserved keywords. Maybe it could be combined with `connectedDevices` or have its own array with `roleName` as a field. I'm unsure but we might also be using non-keyword keys in other request/respond, for example in `SetParticipantData`.

### DIFF
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/AltBeacon.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/AltBeacon.kt
@@ -78,7 +78,6 @@ data class AltBeaconDeviceRegistration(
      * This value is constrained from -127 to 0.
      */
     val referenceRssi: Short,
-    @Required
     override val deviceDisplayName: String? = null, // TODO: We could map known manufacturerId's to display names.
     override val additionalSpecifications: ApplicationData? = null
 ) : DeviceRegistration()
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/BLESerialNumberDeviceRegistration.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/BLESerialNumberDeviceRegistration.kt
@@ -15,7 +15,6 @@ import kotlin.js.JsExport
 @JsExport
 data class BLESerialNumberDeviceRegistration(
     val serialNumber: String,
-    @Required
     override val deviceDisplayName: String? = null,
     override val additionalSpecifications: ApplicationData? = null
 ) : DeviceRegistration()
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/DefaultDeviceRegistration.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/DefaultDeviceRegistration.kt
@@ -15,7 +15,6 @@ import kotlin.js.JsExport
 @Serializable
 @JsExport
 data class DefaultDeviceRegistration(
-    @Required
     override val deviceDisplayName: String? = null,
     override val additionalSpecifications: ApplicationData? = null,
     @Required
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/DeviceRegistration.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/DeviceRegistration.kt
@@ -38,7 +38,6 @@ abstract class DeviceRegistration
      * An optional concise textual representation for display purposes describing the key specifications of the device.
      * E.g., device manufacturer, name, and operating system version.
      */
-    @Required
     abstract val deviceDisplayName: String?
 
     @Required
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/MACAddressDeviceRegistration.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/MACAddressDeviceRegistration.kt
@@ -14,7 +14,6 @@ import kotlin.js.JsExport
 @JsExport
 data class MACAddressDeviceRegistration(
     val macAddress: MACAddress,
-    @Required
     override val deviceDisplayName: String? = null,
     override val additionalSpecifications: ApplicationData? = null
 ) : DeviceRegistration()
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/Website.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/Website.kt
@@ -51,7 +51,6 @@ data class WebsiteDeviceRegistration(
      * The HTTP User-Agent header of the user agent which made the HTTP request to [url].
      */
     val userAgent: String,
-    @Required
     override val deviceDisplayName: String? = userAgent,
     override val additionalSpecifications: ApplicationData? = null
 ) : DeviceRegistration()
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/infrastructure/serialization/UnknownDeviceSerializers.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/infrastructure/serialization/UnknownDeviceSerializers.kt
@@ -179,7 +179,7 @@ data class CustomDeviceRegistration internal constructor(
     @Serializable
     private data class BaseMembers(
         override val deviceId: String,
-        override val deviceDisplayName: String?,
+        override val deviceDisplayName: String? = null,
         override val additionalSpecifications: ApplicationData? = null
     ) : DeviceRegistration()
 
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/infrastructure/versioning/ApiJsonObjectMigrationBuilder.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/infrastructure/versioning/ApiJsonObjectMigrationBuilder.kt
@@ -58,6 +58,17 @@ class ApiJsonObjectMigrationBuilder(
     }
 
     /**
+     * Retrieve the object identified by [fieldName] in this object, and if present, run the specified [migration].
+     */
+    fun updateOptionalObject( fieldName: String, migration: ApiJsonObjectMigrationBuilder.() -> Unit )
+    {
+        val o = json[ fieldName ] as? JsonObject ?: return
+        val newJson: JsonObject = ApiJsonObjectMigrationBuilder( o, minimumMinorVersion, targetMinorVersion )
+            .apply( migration ).build()
+        json[ fieldName ] = newJson
+    }
+
+    /**
      * Retrieve the array identified by [fieldName] in this object, and run the specified [migration].
      */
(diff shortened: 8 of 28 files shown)


Whathecode:
Seems like you figured most of it out! 👍

Just scrolled through it; no proper review yet. To respond to your comments:

1. I guess the problem is the test fails because it does an exact string match? Even though the ordering in JSON shouldn't matter. So, it's probably more appropriate to update the test instead; that would make a fine separate preceding commit. 
2. That's simply the easiest way to serialize that map. Otherwise you need to introduce other various intermediate objects just for serialization purposes, e.g., a `DeviceConfigurationWithRegistration`. And, that wouldn't even take care of polymorphism yet, so you'd likely need to roll custom serializer. In normal form like this, things are simply easier. 😉

Minor comments: 
- The helper function you placed in CARP common at a glance doesn't belong there. Maybe a more generic solution, decoupled from the specific `deviceDisplayName` (like the other helper methods) can be thought up?
- You named the upgrader "version 2 to 3", and it should be "1 to 3".

Lastly, I need to double-check what I used to do before. It seems a bit annoying all of the JSON request test resources already need to be added now. Any further changes within this release will then need to update those. But, likely that's how I did it before as well...

yuanchen233:
> 1. I guess the problem is the test fails because it does an exact string match? Even though the ordering in JSON shouldn't matter.

Yes, and yes, order of object properties should not be a thing. but since we are targeting JavaScript, and apparently Most Browsers iterate object properties as: [String keys, in insertion order (ES2015 guarantees this and all browsers comply)](https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order#:~:text=String%20keys%2C%20in%20insertion%20order%20(ES2015%20guarantees%20this%20and%20all%20browsers%20comply)), and enforced this even more later on. This will be a rare usecase for us, but send request, and get element by index from the respond would work in JavaScript...

> The helper function you placed in CARP common at a glance doesn't belong there.

Yes, it would be useful everytime a field is removed. However, if we update the test and the order of JSON become irrelavent, this helper function is no longer needed.

Whathecode:
> but since we are targeting JavaScript, and apparently Most Browsers iterate object properties as: [String keys, in insertion order (ES2015 guarantees this and all browsers comply)](https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order#:~:text=String%20keys%2C%20in%20insertion%20order%20(ES2015%20guarantees%20this%20and%20all%20browsers%20comply)), and enforced this even more later on.

Interesting! I wasn't aware about that. However, for the purpose of this discussion, I think that should be irrelevant. Our interface across systems is JSON, and not "JSON as processed by JavaScript". If we allow fields to be arranged in any order there is no point in testing that they _are_ in the right order.

Furthermore, at a glance, `kotlinx.serialization` [can't even be configured to require strict ordering](https://github.com/Kotlin/kotlinx.serialization/blob/master/formats/json/commonMain/src/kotlinx/serialization/json/JsonConfiguration.kt) of contained elements. The only exception, I seem to recall, is the `__type` property which needs to be first to make polymorphism work (don't recall this 100% certain; didn't look it up again). Consequently, the library we use also provides no guarantees in terms of ordering! So, at best, your current implementation ensuring a certain ordering depends on implementation details of the current `kotlinx.serialization` library. Not that we should avoid that at all costs; for unknown polymorphism I do the same, _but_, we'd need proper unit test coverage to make sure this still works on version upgrades.

However, long story short, I think the JSON standard correctly states ordering doesn't matter, so our upgraders shouldn't care. :)

yuanchen233:
I looked into the unit testing and we are comparing `JsonElement` where order should not be relevant for `equals` comparison. Could simply be a mistake I made, I'll convert this into a draft and test a bit more, and also fix other things you mentioned.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/infrastructure/versioning/Major1Minor1To3Migration.kt
```diff
@@ -0,0 +1,23 @@
...
+import kotlinx.serialization.ExperimentalSerializationApi
+import kotlinx.serialization.json.JsonPrimitive
+
+
+/**
+ * - The `deviceDisplayName` field in `DeviceRegistration` is no longer required.
+ *   This also apply to `ConnectedDeviceRegistration`.
+ */
+@Suppress( "MagicNumber" )
+abstract class Major1Minor1To3Migration : ApiMigration( 1, 3 )
```
Whathecode: Why is this abstract? I'm guessing you got a bit mislead by the `Major1Minor0To1Migration` base class ([also living in this namespace](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/infrastructure/versioning/Major1Minor0To1Migration.kt)).

The reason this was abstract, and lived in common, as opposed to a specific subsystem being upgrades, was that it was a cross-cutting concern/upgrade for _all_ subsystems. That is not the case here. So, this should be moved to the respective subsystem this upgrade is related to, and a base class is irrelevant as there is no reuse.

Whathecode (changes requested):

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/infrastructure/versioning/Major1Minor1To3Migration.kt
yuanchen233: I created this abstract class because I believe it's better to have a reference for all the migrations that exist in the entire system. That's also the reason I named the upgrader "version 2 to 3" initially as I was following the release version.

Now I realise that each subsystem is rolling their own ApiVersions, and this naming schema will eventually not work.

Whathecode: Yup. 🙂 The reason they are separate subsystems is so that we can evolve them independently. 😉

That said, trying to align on some type of system-wide versioning, by upgrading from 1 to 3 like you did, does provide some clarity in regards to mapping API versions to carp release versions (since this will be part of a carp 1.3.0 release).

yuanchen233: hmmm..the upgrading should then be 1 to 2, by following the `DeploymentService.ApiVersion`

I think my intuition was naming should follow the release version because, the migration will not be published until the next release. It would made more sense if each subsystem are also released separately. (now I think about it, they do release separately, each api migration bumbs api version. I should really start treating subsystems as separate systems, that caused all the confusion here.)

Regarding system-wide versioning, I think major version upgrade can keep things in place? As there could be multiple api migration in one release, so matching release version cannot be achieved. Any strategy enforcing subsystem apiversion to match release version will make each subsystems' own apiversion meaningless.

yuanchen233: > It seems a bit annoying all of the JSON request test resources already need to be added now. Any further changes within this release will then need to update those. But, likely that's how I did it before as well...

Which means we _are_ limiting to only one api migration per release, per subsystem, never mind with my pervious comment.

I'll mark this conversation as resolved, as the abstract class is already removed. As for naming schema, liner change for each subsystem (1 to 2), or matching the release version (1 to 3) are both reasonable.

Whathecode:
Sorry. I overlooked the update to this PR. Will try to see soon whether it can be merged!

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/infrastructure/versioning/Major1Minor1To3Migration.kt
Whathecode: > Any strategy enforcing subsystem apiversion to match release version will make each subsystems' own apiversion meaningless.

Not really. Each time there is _a_ release of CARP core, a new release for all subsystems can be done, regardless of whether there are any changes in that subsystem. For the time being, that simplifies release management/deployment. At most, it's a minor overhead.

But, we wouldn't want to unnecessarily bump API versions. So at any given release, the API version may run behind if there were no new features added to an endpoint.

That would then mean that from time to time (like potentially now), if new features are added, you'd simply jump a couple of minor versions instead of 1.

> Regarding system-wide versioning, I think major version upgrade can keep things in place?

Major version numbers are reserved for backwards _incompatible_ changes. There won't be migrators for those.

yuanchen233: Thanks for the explanation! I also found the discussion in #467 helpful, which I didn't notice in the first place.

Whathecode:
I made things a bit more concise by adding a helper function which is likely to be useful in future upgraders: https://github.com/cph-cachet/carp.core-kotlin/pull/487/commits/5ce4f589145c1798d61f4ef947c503d0260be543

Whathecode (dismissed):

Whathecode (approved):

PR #489, opened 2024-10-24
488 include `deviceregistration` in `devicedeploymentstatus`

### LINKED ISSUE #488: Include `DeviceRegistration` in `DeviceDeploymentStatus`
In a [previous discussion](https://github.com/cph-cachet/carp.core-kotlin/issues/431#issuecomment-1974898425), we concluded that not including `DeviceRegistration` was an omission, with an intent to not leak private data. However, it shouldn't contain private data which caused the original concern (e.g. passwords), and the other data a researcher would already have access to either way.

> In terms of this issue, I think the solution is thus simply to include `DeviceRegistration` information in `DeviceDeploymentStatus` (which is contained within `StudyDeploymentStatus`). Either way, with the current recommended claim-based access documentation, the researcher who set up the deployment would have access to device registrations by calling `getDeviceDeploymentFor()` (which returns `PrimaryDeviceDeployment` and contains the registrations), since they have access to all deployment endpoints.

This data can thus safely be exposed to anyone with access to the deployment. And, what should and shouldn't go into device registrations [is a separate discussion](https://github.com/cph-cachet/carp.core-kotlin/issues/464).

yuanchen233:
`DeviceRegistration` is added to `DeviceDeploymentStatus`, which is part of `StudyDeploymentStatus`. A new interface `HasDeviceRegistration` is introduced to keep this field since `Unregistered` would not need this field.

A regression test case was added in a separate commit as I'm not sure if it is desired. The test case amis to test `GetStudyDeploymentStatusList` endpoint with registered devices in one of the deployments. This is not covered since `DeploymentServiceTest` tests minimum usage for `GetStudyDeploymentStatusList`, as it is just a list representation of `GetDeploymentStatus`. I added it since json migration for this endpoint works differently.

### DIFF
--- a/carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/application/DeviceDeploymentStatus.kt
+++ b/carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/application/DeviceDeploymentStatus.kt
@@ -3,6 +3,7 @@
 package dk.cachet.carp.deployments.application
 
 import dk.cachet.carp.common.application.devices.AnyDeviceConfiguration
+import dk.cachet.carp.common.application.devices.DeviceRegistration
 import kotlinx.serialization.*
 import kotlin.js.JsExport
 
@@ -56,6 +57,16 @@ sealed class DeviceDeploymentStatus
         abstract val remainingDevicesToRegisterBeforeDeployment: Set<String>
     }
 
+    /**
+     * A device deployment status which indicates the device has been registered.
+     */
+    sealed interface HasDeviceRegistration
+    {
+        /**
+         * The last [DeviceRegistration] for this device.
+         */
+        val deviceRegistration: DeviceRegistration
+    }
 
     /**
      * Device deployment status for when a device has not been registered.
@@ -74,18 +85,20 @@ sealed class DeviceDeploymentStatus
     @Serializable
     data class Registered(
         override val device: AnyDeviceConfiguration,
+        override val deviceRegistration: DeviceRegistration,
         override val canBeDeployed: Boolean,
         override val remainingDevicesToRegisterToObtainDeployment: Set<String>,
         override val remainingDevicesToRegisterBeforeDeployment: Set<String>
-    ) : NotDeployed()
+    ) : NotDeployed(), HasDeviceRegistration
 
     /**
      * Device deployment status when the device has retrieved its [PrimaryDeviceDeployment] and was able to load all the necessary plugins to execute the study.
      */
     @Serializable
     data class Deployed(
-        override val device: AnyDeviceConfiguration
-    ) : DeviceDeploymentStatus()
+        override val device: AnyDeviceConfiguration,
+        override val deviceRegistration: DeviceRegistration
+    ) : DeviceDeploymentStatus(), HasDeviceRegistration
     {
         // All devices that have been deployed necessarily can be deployed.
         override val canBeDeployed = true
@@ -97,9 +110,10 @@ sealed class DeviceDeploymentStatus
     @Serializable
     data class NeedsRedeployment(
         override val device: AnyDeviceConfiguration,
+        override val deviceRegistration: DeviceRegistration,
         override val remainingDevicesToRegisterToObtainDeployment: Set<String>,
         override val remainingDevicesToRegisterBeforeDeployment: Set<String>
-    ) : NotDeployed()
+    ) : NotDeployed(), HasDeviceRegistration
     {
         // All devices that have been deployed necessarily can be deployed.
         override val canBeDeployed = true
--- a/carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/infrastructure/versioning/DeploymentServiceApiMigrator.kt
+++ b/carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/infrastructure/versioning/DeploymentServiceApiMigrator.kt
@@ -38,13 +38,32 @@ private val major1Minor1To3Migration =
             {
                 "dk.cachet.carp.deployments.infrastructure.DeploymentServiceRequest.GetDeviceDeploymentFor" ->
                 {
-                    val responseObject = (response.response as? JsonObject)?.migrate {
+                    val responseObject = response.response?.jsonObject?.migrate {
                         requiredDeviceDisplayName( "registration" )
 
                         updateOptionalObject( "connectedDeviceRegistrations" )
                         {
                             for ( connectedDevice in json.keys ) requiredDeviceDisplayName( connectedDevice )
                         }
+                    }
+                    ApiResponse( responseObject, response.ex )
+                }
+                "dk.cachet.carp.deployments.infrastructure.DeploymentServiceRequest.CreateStudyDeployment",
+                "dk.cachet.carp.deployments.infrastructure.DeploymentServiceRequest.GetStudyDeploymentStatus",
+                "dk.cachet.carp.deployments.infrastructure.DeploymentServiceRequest.RegisterDevice",
+                "dk.cachet.carp.deployments.infrastructure.DeploymentServiceRequest.UnregisterDevice",
+                "dk.cachet.carp.deployments.infrastructure.DeploymentServiceRequest.DeviceDeployed",
+                "dk.cachet.carp.deployments.infrastructure.DeploymentServiceRequest.Stop" ->
+                {
+                    val responseObject = response.response?.jsonObject?.migrate {
+                        removeDeviceRegistration()
+                    }
+                    ApiResponse( responseObject, response.ex )
+                }
+                "dk.cachet.carp.deployments.infrastructure.DeploymentServiceRequest.GetStudyDeploymentStatusList" ->
+                {
+                    val responseObject = response.response?.jsonArray?.migrate {
+                        objects { removeDeviceRegistration() }
                     }
                     ApiResponse( responseObject, response.ex )
                 }
@@ -58,6 +77,15 @@ private val major1Minor1To3Migration =
         private fun ApiJsonObjectMigrationBuilder.requiredDeviceDisplayName( fieldName: String ) =
             updateObject( fieldName ) {
                 json.getOrPut( "deviceDisplayName" ) { JsonNull }
+            }
+
+        /**
+         * Remove the newly added `deviceRegistration` field from `DeviceDeploymentStatus` objects contained within
+         * `StudyDeploymentStatus`.
+         */
+        private fun ApiJsonObjectMigrationBuilder.removeDeviceRegistration() =
+            updateArray( "deviceStatusList" ) {
+                objects { json.remove( "deviceRegistration" ) }
             }
 
         override fun migrateEvent( event: JsonObject ) = event
--- a/rpc/schemas/deployments/DeviceDeploymentStatus.json
+++ b/rpc/schemas/deployments/DeviceDeploymentStatus.json
@@ -30,6 +30,14 @@
       },
       "required": [ "remainingDevicesToRegisterToObtainDeployment", "remainingDevicesToRegisterBeforeDeployment" ]
     },
+    "HasRegistration": {
+        "$anchor": "HasRegistration",
+        "type": "object",
+        "properties": {
+            "deviceRegistration": { "$ref": "../common/devices/DeviceRegistration.json" }
+        },
+        "required": [ "deviceRegistration" ]
+    },
     "Unregistered": {
       "$anchor": "Unregistered",
       "allOf": [
@@ -47,7 +55,8 @@
       "$anchor": "Registered",
       "allOf": [
         { "$ref": "#/$defs/DeviceDeploymentStatus" },
-        { "$ref": "#NotDeployed" }
+        { "$ref": "#NotDeployed" },
+        { "$ref": "#HasRegistration" }
       ],
       "properties": {
         "__type": { "const": "dk.cachet.carp.deployments.application.DeviceDeploymentStatus.Registered" },
@@ -59,7 +68,8 @@
     "Deployed": {
       "$anchor": "Deployed",
       "allOf": [
-        { "$ref": "#/$defs/DeviceDeploymentStatus" }
+        { "$ref": "#/$defs/DeviceDeploymentStatus" },
+        { "$ref": "#HasRegistration" }
       ],
       "properties": {
         "__type": { "const": "dk.cachet.carp.deployments.application.DeviceDeploymentStatus.Deployed" }
@@ -70,7 +80,8 @@
       "$anchor": "NeedsRedeployment",
       "allOf": [
         { "$ref": "#/$defs/DeviceDeploymentStatus" },
-        { "$ref": "#NotDeployed" }
+        { "$ref": "#NotDeployed" },
+        { "$ref": "#HasRegistration" }
       ],
       "properties": {
         "__type": { "const": "dk.cachet.carp.deployments.application.DeviceDeploymentStatus.NeedsRedeployment" }
--- a/carp.clients.core/src/commonTest/kotlin/dk/cachet/carp/clients/domain/ClientRepositoryTest.kt
+++ b/carp.clients.core/src/commonTest/kotlin/dk/cachet/carp/clients/domain/ClientRepositoryTest.kt
@@ -97,7 +97,7 @@ interface ClientRepositoryTest
                 Clock.System.now(),
                 deploymentId,
                 listOf(
-                    DeviceDeploymentStatus.Registered( primaryDevice, true, emptySet(), emptySet() )
+                    DeviceDeploymentStatus.Registered( primaryDevice, registration, true, emptySet(), emptySet() )
                 ),
                 emptyList(),
                 null
--- a/carp.clients.core/src/commonTest/kotlin/dk/cachet/carp/clients/domain/study/StudyTest.kt
+++ b/carp.clients.core/src/commonTest/kotlin/dk/cachet/carp/clients/domain/study/StudyTest.kt
@@ -311,6 +311,7 @@ class StudyTest
                 listOf(
                     DeviceDeploymentStatus.Registered(
                         smartphone,
+                        primaryDeviceDeployment.registration,
                         true,
                         emptySet(),
                         connectedDevices.map { it.roleName }.toSet()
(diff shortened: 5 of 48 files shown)


carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/application/DeviceDeploymentStatus.kt
```diff
@@ -56,6 +57,17 @@ sealed class DeviceDeploymentStatus
         abstract val remainingDevicesToRegisterBeforeDeployment: Set<String>
     }
 
+    /**
+     * A device deployment status which indicates the device has been registered.
+     */
+    sealed interface HasDeviceRegistration
+    {
+        /**
+         * The registration information of the device, if available.
```
Whathecode: It's always available. That's what `HasDeviceRegistration` means, no? :)

Whenever writing documentation, it always pays off to think more about what _isn't_ in the name yet. In this case, all I could come up with is:

> The last [DeviceRegistration] for this device.

At least it adds extra info (devices can be re-registered). If it doesn't, it is better to omit the documentation altogether.

carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/infrastructure/versioning/DeploymentServiceApiMigrator.kt
```diff
@@ -48,6 +47,35 @@ private val major1Minor1To3Migration =
                     }
                     ApiResponse( responseObject, response.ex )
                 }
+                "dk.cachet.carp.deployments.infrastructure.DeploymentServiceRequest.CreateStudyDeployment",
+                "dk.cachet.carp.deployments.infrastructure.DeploymentServiceRequest.GetStudyDeploymentStatus",
+                "dk.cachet.carp.deployments.infrastructure.DeploymentServiceRequest.UnregisterDevice",
+                "dk.cachet.carp.deployments.infrastructure.DeploymentServiceRequest.DeployDevice",
```
Whathecode: This should be `DeviceDeployed`. 🤔 Why does it work? I'm not a big TDD fan, but there is something to _seeing_ a test fail before making it pass.

Seems like luck would have it that's _exactly_ where coverage is missing. 😅 Could you add a preceding commit which provides coverage for `DeploymentService.deviceDeployed()`? And, while at it, verify that all the other endpoints of this service are actually called in the current regression tests.

carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/infrastructure/versioning/DeploymentServiceApiMigrator.kt
```diff
@@ -48,6 +47,35 @@ private val major1Minor1To3Migration =
...
+                "dk.cachet.carp.deployments.infrastructure.DeploymentServiceRequest.Stop",
+                "dk.cachet.carp.deployments.infrastructure.DeploymentServiceRequest.RegisterDevice" ->
+                {
+                    val responseObject = response.response?.jsonObject?.migrate {
+                        updateArray( "deviceStatusList" ) {
```
Whathecode: This looks like the same code twice, so why not make this part of the common function?

carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/infrastructure/versioning/DeploymentServiceApiMigrator.kt
```diff
@@ -60,10 +88,15 @@ private val major1Minor1To3Migration =
...
+        /**
+         * The `deviceRegistration` field was added to the `DeviceStatus` object.
+         */
+        private fun ApiJsonObjectMigrationBuilder.removeDeviceRegistration() =
+            json.remove( "deviceRegistration" )
+
         override fun migrateEvent( event: JsonObject ) = event
     }
 
-
```
Whathecode: I do use double new lines to indicate some type of higher-level grouping (so that new lines _within_ groups can be used). Understandably not that clear to everyone, per se. :)

rpc/schemas/deployments/DeviceDeploymentStatus.json
```diff
@@ -29,6 +29,13 @@
       },
       "required": [ "remainingDevicesToRegisterToObtainDeployment", "remainingDevicesToRegisterBeforeDeployment" ]
     },
+    "HasRegistration": {
+        "type": "object",
+        "properties": {
+            "deviceRegistration": { "$ref": "../common/devices/DeviceRegistration.json" }
+        },
+        "required": [ "registration" ]
```
Whathecode: Should be `deviceRegistration`. Odd it's not caught by tests. 🤔 

Likely relevant to debug `JsonSchemasTest` a bit to see why it didn't complain. Also, I see we are on an old version `com.networknt:json-schema-validator:1.0.67` and there are plenty of updates out.

Whathecode (changes requested):
Smaller comments I fixed in a fixup commit.

Some bigger observations:
- I would have expected a migration to be set up for `RecruitmentService` in the studies subsystem as well, which returns `ParticipantGroupStatus`, which contains `StudyDeploymentStatus` once "invited". It's odd there are no failing tests on this. 🤔 
- Extra test cases seem relevant (also see my comment on a bug which wasn't caught due to lacking coverage of `DeploymentService`). But, if you add them for _past_ versions (p.s. you added it to `1.0` and not `1.1`, was that intentional?), make sure they were generated with code for that version. You may have to branch out from an old release, write the test there, and then possibly do a cherry-pick on this branch.

rpc/schemas/deployments/DeviceDeploymentStatus.json
Whathecode: Seems like upgrading to the latest JSON Schema validator at least starts failing, but for other reasons. There is an error in `DeviceDeploymentStatus.json` which wasn't caught by the old (currently used) version of the validator.

```
"allOf": [ "#/$defs/DeviceDeploymentStatus", "#/$defs/NotDeployed", "#/$defs/HasRegistration" ],
```

That is an incorrect reference to other schemas, as it is missing `"$ref"`. It's correct in other places, e.g.:

```
"allOf": [ { "$ref": "Data.json#SensorData" } ],
```

I'll create a pull request to fix this.

Whathecode: https://github.com/cph-cachet/carp.core-kotlin/pull/491

carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/infrastructure/versioning/DeploymentServiceApiMigrator.kt
yuanchen233: We only have coverage for `CreateStudyDeployment`, if I remember correctly that was the only failing test, as [ Actual testing of the status responses should already be covered adequately in StudyDeployment tests.](https://github.com/cph-cachet/carp.core-kotlin/blob/07dbd91377353990db18f1bc50c6efd6665bae60/carp.deployments.core/src/commonTest/kotlin/dk/cachet/carp/deployments/application/DeploymentServiceTest.kt#L101), but the regression test case are generated from `DeploymentServiceTest`.

> Extra test cases seem relevant

Do we only want the generated json request files, or we also want those tests in `DeploymentServiceTest`? I think I'll include them first anyways to agree on the test cases, and they can be removed later if not desired. 

I added it to `1.0` because I felt the test case was needed from the start of this major version, not due to some changes introduced in `1.1`. Then simlly forget to add them `1.1` 😛. As I generated those json files in ver. 1.0 and see a list of diff with origional ones, wandering what I breaked and realized those are the generated UUIDs difference...

carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/infrastructure/versioning/DeploymentServiceApiMigrator.kt
yuanchen233: I did notice that pattern, like [`DeviceDeploymentStatus`](https://github.com/cph-cachet/carp.core-kotlin/blob/10da0ccbc184619a72decabe5ccb6fb1499d1527/carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/application/DeviceDeploymentStatus.kt), double new lines between `vals` and `classes`, and single line between each `val` or `class`, but it seems like I have not adopt to it well enough to notice it here between big code blocks.😅 Also, maybe I should put in some effort into the codestyle xml so it solve future issues. (not sure if it handles this one..)

Whathecode: Nah. It's just the idiosyncrasies of me having authored this codebase for years. Don't worry about it. As long as I have time to review PRs, I'll gladly clean it up to stay consistent. 😂 But, it's definitely not important.

carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/infrastructure/versioning/DeploymentServiceApiMigrator.kt
Whathecode: > Do we only want the generated json request files, or we also want those tests in DeploymentServiceTest?

I'd always keep them in sync, so that it's just a matter of copying the output of all of them over. 

But, obviously the comment I left in code there didn't consider the implications on reduced regression testing of the infrastructure layer (JSON) by not covering all potential input/output. 😅 So albeit valid—there is plenty of functional coverage in StudDeploymentTest—, that doesn't cover serialization and JSON schema definitions.

I think it would make sense to write test cases that try to cover as much as possible, rather than writing them as unit tests. These are more or less integration tests after all. So, I'd focus on longer scenarios with multiple calls reflecting fuller study lifetime to get more coverage for the infrastructure layer.

> As I generated those json files in ver. 1.0 and see a list of diff with origional ones, wandering what I breaked and realized those are the generated UUIDs difference...

I actually spotted that. 🤔. I thought I injected some more deterministic GUID and timestamp generator for the creation of these tests. I saw it in some test resources, but not others. I don't remember whether that's expected/normal, or oversight.

rpc/schemas/deployments/DeviceDeploymentStatus.json
Whathecode: PR is ready for review/merge.

Whathecode:
@yuanchen233 Is this PR still intended to be merged at some point? Please request a new review if there is anything new for me to look at. ;)

yuanchen233:
> @yuanchen233 Is this PR still intended to be merged at some point? Please request a new review if there is anything new for me to look at. ;)

Hi, yes, I got distracted and forget about this one, I'm back on it now, will do so. My bad on the delays.

yuanchen233:
Hi @Whathecode, I realized that `backwardsCompatibilityTest` /` generateExampleRequests` only generates requests from the current subsystem. As a result, any transitions that rely on other subsystems aren't covered. In this case, we don’t have any regression test cases covering `ParticipantGroupStatus.Running`, since it depends on the deployment subsystem to transition into that state.

Do we want to keep this behavior?
If so, the regression test cases for recruitmentService can be removed. (They are the reason for failing tests)
Otherwise, we’ll need a place to maintain cross-subsystem regression tests to ensure each subsystem can still be upgraded independently.

I’ve only rebased onto the develop branch and reordered some commits, but I kept all commits in this branch for reference, as it was started a while ago. I think [getStudyDeploymentStatusList_with_registered_device_succeeds.json](https://github.com/cph-cachet/carp.core-kotlin/pull/489/commits/c2d3ad1e728e7893db75c8149e142cdc639cbf43#diff-c5f3a29fad3582281712bd8b2a04b8662dfdc8bd020fd06dde5602c71cc768cb) is redundant now and can be dropped.

For test-request json files I branched out form "bump version" commits.

Whathecode:
> As a result, any transitions that rely on other subsystems aren't covered.

I was under the impression that each subsystem did integration tests with its neighboring systems (`HostIntegrationTest`). Maybe I'm mistaken.

UPDATE: it seems like those are only for the current version, and these test results aren't serialized.

> For test-request json files I branched out form "bump version" commits.

You'd want to branch out from the release tags (you can find exact commits in the releases). Version bumping can be done pre-emptively.

carp.deployments.core/src/commonTest/resources/test-requests/DeploymentService/1.0/DeploymentServiceTest/getStudyDeploymentStatusList_with_registered_device_succeeds.json
```diff
@@ -0,0 +1,493 @@
+[
+    {
+        "outcome": "Succeeded",
```
Whathecode: Where does this test come from? You add it in the first commit, but I don't see this test existing in the sources. 

You mention it may be obsolete. If this test doesn't exist, and is something you wrote in a previous version of this PR, it most definitely is. 

Please rebase to a clean commit history. This PR is old enough that I have to look at it again with a clean slate either way. 🙂 And, approval will require doing so either way.

yuanchen233: Got it, will do.

carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/infrastructure/versioning/DeploymentServiceApiMigrator.kt
Whathecode: > Could you add a preceding commit which provides coverage for DeploymentService.deviceDeployed()?

There actually is coverage for a _failing_ call to `deviceDeployed()` (`modifications_after_stop_not_allowed`), but there wasn't one for a succeeding call, which is the one that would fail if this migration fails.

yuanchen233:
> I was under the impression that each subsystem did integration tests with its neighboring systems (`HostIntegrationTest`). it seems like those are only for the current version, and these test results aren't serialized.

Should we consider serialize them and add them to 'BackwardsCompatibilityTest'? That would be a better place for the test case I added in `RecuritmentService`. Or can we just rely on `StudyDeploymentTest` and `HostIntegrationTest` and remove the test case I added? Either way there are some changes needs to be done to pass the checks, but I am unsure if we have enough coverage, it's a bit tricky for me since those are information belong to `StudyDeployment` but presents in `RecruitmentService`.

> You'd want to branch out from the release tags (you can find exact commits in the releases). Version bumping can be done pre-emptively.

Got it, I rebased my branch for generating test cases with release tags.

----
Also, might I ask why is `StoppedOn` not part of `StudyDeploymentSnapshot`, instead call `stop` again when loading data from snapshots [here](https://github.com/cph-cachet/carp.core-kotlin/blob/14b26ab4ffcfb3efbe12338df802d761ac53e1b4/carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/domain/StudyDeployment.kt#L125C1-L126C1)? This will results in different `StoppedOn` value when calling `getDeploymentStatus` after the deployment has stopeed.

Whathecode:
> Should we consider serialize them and add them to 'BackwardsCompatibilityTest'? That would be a better place for the test case I added in RecuritmentService. Or can we just rely on StudyDeploymentTest and HostIntegrationTest and remove the test case I added?

I don't think they can be serialized/tested with the current test setup, as they verify the output of two application services. The serialization test setup focuses on a single service. I'll have a look to see what a good spot is for the test case you added; I'm also a bit uncertain at the moment. Maybe we should just try to keep them unit tests, and add coverage for `deviceDeployed` which was missing (which you do). A "full lifecycle" type of test could possibly be useful as a separate test file (still focusing on a single service), however. So that the all tests in e.g. `DeploymentServiceTest` aim to be as small as possible.

> Also, might I ask why is StoppedOn not part of StudyDeploymentSnapshot, instead call stop again when loading data from snapshots [here](https://github.com/cph-cachet/carp.core-kotlin/blob/14b26ab4ffcfb3efbe12338df802d761ac53e1b4/carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/domain/StudyDeployment.kt#L125C1-L126C1)? This will results in different StoppedOn value when calling getDeploymentStatus after the deployment has stopeed.

That must stem from a time when calling `stop()` resulted in the same outcome, and prevented having to directly modify state, somewhat reducing the change of invalid states arising. However, it's allowed to access it, and in this case probably should, otherwise it's a bug. :)

yuanchen233:
> Maybe we should just try to keep them unit tests, and add coverage for `deviceDeployed` which was missing (which you do). A "full lifecycle" type of test could possibly be useful as a separate test file (still focusing on a single service), however. So that the all tests in e.g. `DeploymentServiceTest` aim to be as small as possible.

Yes, I do think following the original design choice makes sense.  
I believe the issue we’re facing is that the result returned from `RecruitmentServiceRequest` includes information that actually belongs to `DeploymentService`.  

While this is fine for general testing, it becomes problematic for backward compatibility tests, as those dependencies are not properly covered.


> That must stem from a time when calling `stop()` resulted in the same outcome, and prevented having to directly modify state, somewhat reducing the change of invalid states arising. However, it's allowed to access it, and in this case probably should, otherwise it's a bug. :)

I checked the running CAWS service, and observed that calling `getStudyDeploymentStatus` after a deployment has been stopped returns a different `stoppedOn` value each time.  
It appears that the value reflects the time of the request, rather than the actual time the deployment was stopped.

Consider this test case which would fail currently due to different `stoppedOn` value.
```
    fun getStudyDeploymentStatus_after_stop_always_returns_same_result() = runTest {
        val (service, _) = createSUT()
        val deviceRoleName = "Primary"
        val (protocol, _, _) = createSinglePrimaryWithConnectedDeviceProtocol( deviceRoleName )
        val protocolSnapshot = protocol.getSnapshot()
        val invitation1 = createParticipantInvitation( AccountIdentity.fromUsername( "User 1" ) )
        val deploymentId1 = UUID.randomUUID()
        service.createStudyDeployment( deploymentId1, protocolSnapshot, listOf( invitation1 ) )
        service.stop( deploymentId1 )

        val status1 = service.getStudyDeploymentStatus( deploymentId1 )
        val status2 = service.getStudyDeploymentStatus( deploymentId1 )
        assertEquals( status1, status2 )
    }
```
The reason behind this behavior is that `stoppedOn` is not part of the snapshot, so its value is not persisted in the database.  

If I understand correctly, when `getStudyDeploymentStatus` is called, a new `StudyDeployment` instance is reconstructed using `fromSnapshot`, and `stop` is invoked again with the current time.  
As a result, the `stoppedOn` value differs each time the status is retrieved.

However, this should be a separate issue.

Whathecode:
> The reason behind this behavior is that stoppedOn is not part of the snapshot, so its value is not persisted in the database.

Ugh, true. :( And fixing that while retaining data from previous versions in a database will require a database migration, as part of which the boolean `isStopped` will need to be converted to some representative `stoppedOn` time, which will have to be semi-arbitrary. :(

Whathecode:
This PR has been reworked and is now replaced by https://github.com/cph-cachet/carp.core-kotlin/pull/513

Issue #353, opened 2022-01-24
Replace JSON pointer URI fragment with anchor references in JSON schemas


Whathecode:
The json-schema-validator library [we want to use for JSON schema validation](https://github.com/cph-cachet/carp.core-kotlin/pull/351) [does not yet support named anchor lookup](https://github.com/networknt/json-schema-validator/issues/227) for the metaschema we are using (draft 2019/09).

As a workaround, we currently use JSON pointer URIs within our schemas to point to these anchors.

The current locations are:

- `TriggeredTask.json` to point to `TaskControl#Control`
- In application service request objects (e.g., `DataStreamServiceRequest.json`) to point to definitions within the same file
- `JsonSchemaTest.kt` to load the reponse schemas of requests

Once json-schema-validator provides support for this, we can upgrade and replace these references with named anchor references.

Whathecode:
This works partially with a bugfix for `json-schema-validator` introduced [in 1.0.67](https://github.com/networknt/json-schema-validator/releases/tag/1.0.67). All external references to anchors now use the anchor notation. However, internal anchor references still fail and currently thus still use the full path notation.

Whathecode:
> However, internal anchor references still fail and currently thus still use the full path notation.

With the latest version of JSON schema validator this now also works for internal references. However, it begs the question whether we _want to_ replace all internal references with anchors (and thus add anchors where there currently aren't any).

Are there conventions for when to use an anchor and when not?

PR #491, opened 2024-10-26
Upgrade JSON schema validator + fixes

Whathecode:
There was an error in the `DeviceDeploymentStatus` JSON schema which wasn't spotted by the JSON schema validator.

Upgrading to the latest version definitely provides better validation, but in addition also seems to introduce a bug in the evaluation of `unevaluatedProperties` (https://github.com/networknt/json-schema-validator/issues/1123). For now, I have removed these where it caused validation to fail, thus resulting in less strict/correct schema definitions. If we decide to merge this PR, we need to create a new bug to fix this once the bug is fixed in the schema validator.

On the upside, with the new version of `networknt
/json-schema-validator` it should be possible to use more concise anchor references (https://github.com/cph-cachet/carp.core-kotlin/issues/353). I'll add this as a separate commit later.

### DIFF
--- a/rpc/build.gradle
+++ b/rpc/build.gradle
@@ -10,7 +10,7 @@ dependencies {
     implementation "org.reflections:reflections:${versions.reflections}"
     implementation "org.jetbrains.kotlinx:kotlinx-serialization-json:${versions.serialization}"
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:${versions.coroutines}"
-    implementation "com.networknt:json-schema-validator:1.0.67"
+    implementation "com.networknt:json-schema-validator:1.5.2"
     implementation "commons-io:commons-io:${versions.apacheCommons}"
 
     def generateForProjects = coreModules + commonModule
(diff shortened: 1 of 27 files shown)


yuanchen233 (dismissed):
Before this bug is fixed, those schemas allows additional properties. (I tried to apply `"additionalProperties": false` resulting the same error message)

Whathecode:
@yuanchen233 This should be a cleaner workaround, to be resolved in #492.

This has the benefit that extra unexpected properties on concrete types will still correctly fail. And existing properties are properly evaluated. Just a messy/redundant schema definition which shouldn't be necessary.

yuanchen233 (approved):
> Are there conventions for when to use an anchor and when not?

Are we using anchor for all subschemas? Could you also leave a note/answer to your own question in #353 for future referance? 😀

Whathecode:
> Could you also leave a note/answer to your own question in #353 for future referance? 😀

Of course! 👍🙂

Check the last commit description of this PR. I can't copy paste it now (I'm on phone).

Issue #496, opened 2025-03-08
Change build setup to publish snapshot builds


Whathecode:
In another library I manage, I ran into issues releasing builds to Maven after upgrading to Kotlin 2.1.0. I [had to change the way snapshot configuration works in builds](https://github.com/Whathecode/kotlinx.interval/pull/70/commits/d854dcd016db518dec582e4685dc3405cceb3fdd).

Since this library uses the same mechanism, this should likely be changed accordingly as part of upgrading to Kotlin 2.1.0.

PR #501, opened 2025-05-03
Upgrade to Kotlin 2.x

### LINKED ISSUE #495: Use official Kotlin list/map/set integration for Javascript
Currently, wrappers to convert from/to Kotlin `List`/`Map`/`Set` and Javascript equivalents are maintained manually through a hackish wrapper which exposes internals of generated JS code.

The lack of support for this has been known for a while, [but it seems starting from the Kotlin 2.0.0 release things have improved](https://youtrack.jetbrains.com/issue/KT-34995/JS-List-Map-and-Set-types-are-hard-to-use-from-JS-because-of-mangled-member-names#focus=Comments-27-11089648.0-0). Verify whether this can be used so the custom wrappers no longer need to be maintained.

### LINKED ISSUE #499: Upgrade to Kotlin 2.x

Whathecode:
I got JS/JVM tests passing on the latest version of Kotlin (2.1.20) (closes #499).
As expected, this required quite extensive changes to the TypeScript wrappers (including adopting KtSet/KtList/KtMap; closes #495). With a bit of hacking, I got this to work as well!

What is left is:
- [x] Upgrading kotlinx.datetime to 0.6.2
- [x] Upgrading kotlinx.serialization to 1.8.1
- [x] ... and while we are at it, nexusPublishPlugin to '2.0.0' (it's a major version, so we need to double-check whether further changes are needed not to break automated publishing of artifacts)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,6 +70,10 @@ fabric.properties
 
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
+replay_pid*
+
+# Missing from gitignore.io template
+.kotlin/
 
 ### Node ###
 # Logs
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ buildscript {
 
         versions = [
             // Kotlin multiplatform versions.
-            kotlin:'1.9.23',
+            kotlin:'2.1.20',
             serialization:'1.6.3',
             coroutines:'1.8.0',
             datetime:'0.5.0',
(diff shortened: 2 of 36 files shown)


Whathecode:
Mmmm ... without #496, this PR will fail to publish a snapshot once merged. I can have a look at this as well.

And, it's a bit annoying you need to use both the auto-generated `KtSet` and custom `List`, due to `Collection` not being supported yet. That shouldn't block the release (and only impacts TypeScript consumers), but at least a bug/request should be reported to JetBrains, and maybe we can try to find a workaround with some more JS/TypeScript post-compilation mutations.

NGrech (approved):
Had a look at the Kotlin code mainly, everything there looked good. 
Checks are passing. 
I will assume correctness for type script code.

Whathecode:
Great! Let's merge it, and I'll see whether I can pick up #496 after. So, don't be surprised to see failing snapshot publishing being reported on main. :)

PR #506, opened 2025-05-30
Fix publication of snapshots

### LINKED ISSUE #496: Change build setup to publish snapshot builds

Whathecode:
Fixes #496 

In addition, I already bumped the version to 1.3.0, so that snapshots are published as `1.3.0-SNAPSHOT` and not `1.2.1-SNAPSHOT`. And, updated some GitHub action dependencies. On another repo, I saw occasional failures of the gradle wrapper verification, and the version used here is a pretty old.

### DIFF
--- a/.github/workflows/publish-snapshots.yml
+++ b/.github/workflows/publish-snapshots.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
         with:
           fetch-depth: 1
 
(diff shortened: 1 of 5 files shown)


yuanchen233 (approved):
:shipit:

Issue #478, opened 2024-06-11
Need more endpoints for getting data in the Data Service


bardram:
Currently there is only one endpoint for getting data which is based on sequence IDs. This is clearly no useful for getting any data since nobody knows about such IDs.

We need getter endpoints for getting things like:

* a general query getter to query for data
* get data for a deployment
* for data types
* time
* etc.

PR #509, opened 2025-06-25
Carp analytics core data

NGrech:


carp.data.core/build.gradle
```diff
@@ -18,5 +18,17 @@ kotlin {
...
         }
+
+        jvmMain {
+            dependencies {
+                implementation "org.xerial:sqlite-jdbc:3.39.3.0"
+            }
+        }
+        jvmTest {
+            dependencies {
+                implementation("org.xerial:sqlite-jdbc:3.39.3.0")
```
Whathecode, line 29: Main dependencies are automatically pulled into test. So this can be removed.

However, I'm skeptical of pulling in this dependency; is it really needed? For testing, any reason an in memory database doesn't suffice? Remember that core is supposed to be infrastructure agnostic. The infrastructure that is available is still very technology-independent and doesn't pull in external dependencies.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataPoint.kt
```diff
@@ -0,0 +1,17 @@
...
+import kotlinx.datetime.Instant
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+/**
+ * A single typed data point from a CARP data stream, annotated with its stream origin and timestamp.
+ */
+@Serializable
+@SerialName("dk.cachet.carp.analytics.AnalyticsDataPoint")
+data class CollectedDataPoint(
```
Whathecode, line 13: We need to better disambiguate this from `DataStreamPoint`. (similar comment for all `Collected...` types).

The difference seems to be this doesn't carry information to resync (no sequence id/sync points, so I suppose this represents synchronized data), provenance (information on why data was collected), and can only represent points in time, not intervals. `Collected` doesn't do that justice at all. In fact the opposite, as what we collected was more rich data.

Out of context it is unclear why that is needed in the data subsystem. Consider that it may suffice to define this in the analytics subsystem, reducing the scope within which you need to worry about disambiguation.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataQuery.kt
```diff
@@ -0,0 +1,15 @@
+package dk.cachet.carp.data.application
+
+import dk.cachet.carp.common.application.UUID
+import kotlinx.datetime.Instant
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class CollectedDataQuery(
```
Whathecode, line 8: Missing documentation. E.g. what is offsetdays? What if any/all of these values are null? Are you certain they have to be nullable (otherwise it just adds unnecessary complexity).

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataSet.kt
```diff
@@ -0,0 +1,33 @@
...
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+/**
+ * A simple wrapper around a list of [CollectedDataPoint]s.
+ * Used to represent a collection of typed data for processing within a workflow step.
+ */
+@Serializable
+@SerialName("dk.cachet.carp.analytics.AnalyticsDataCollection")
+data class CollectedDataSet( val points: List<CollectedDataPoint> = emptyList() )
```
Whathecode, line 14: Why a different serial name than class name? 

And, it's a bit unclear why this class exists. You can support the same as extension functions without the overhead.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataSet.kt
```diff
@@ -0,0 +1,33 @@
...
+{
+    /**
+     * Filter by stream ID.
+     */
+    fun filterByStream( streamId: DataStreamId ): CollectedDataSet =
+        CollectedDataSet(points.filter { it.streamId == streamId })
+
+    /**
+     * Filter by time range.
```
Whathecode, line 23: Documentation that repeats the method name is a code smell. Instead, focus on what isn't clear yet, such as, boundary conditions (are data points on the boundary included or excluded?)

And, disambiguate "filter", and do so consistently. In one method you use `Filter`, in the other `Extract`. Same behavior should be documented consistently (I sometimes even copy over phrasing from the standard library in cases like this).

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/StudyDataService.kt
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Provides researcher-optimized access to collected data for studies, devices, and streams.
+ */
+interface StudyDataService : ApplicationService<StudyDataService, StudyDataService.Event>
+{
+
+    companion object
+    {
+        val API_VERSION = ApiVersion(1, 0)
```
Whathecode, line 18: I need to double check what I did before, but I think it would make sense to align the version with the version of carp when the endpoints are introduced.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/StudyDataService.kt
```diff
@@ -0,0 +1,46 @@
...
+ *   - deviceRoleNames: List of device role names (e.g., `phone`) to include.
+ *   - from: Absolute start time for filtering.
+ *   - to: Absolute end time for filtering.
+ *   - offsetDays: Relative offset window (e.g., first 30 days after subject enrollment).
+ *
+ * @return A [CollectedDataSet] containing the matching data points.
+ */
+    suspend fun getCollectedData(
+        studyId: UUID,
+        query: CollectedDataQuery
```
Whathecode, line 44: The documentation of this method shows you would be better off just having these as separate parameters instead of an object wrapping them, then you can document them in place and reference them statically (`[param]`).

If there is a good reason to keep the query object then that is where all of this should be documented. Keep documentation close to what it documents, and when using abstractions elsewhere, don't document their internal specifics; defer to the abstraction. It reduces risk of things going out of sync, and also forces you to reason more about any abstraction you introduce.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/StudyDataService.kt
```diff
@@ -0,0 +1,46 @@
...
+ *   - fields: List of CARP Data Type fields (e.g., `step_count`) to include.
+ *   - deviceRoleNames: List of device role names (e.g., `phone`) to include.
+ *   - from: Absolute start time for filtering.
+ *   - to: Absolute end time for filtering.
+ *   - offsetDays: Relative offset window (e.g., first 30 days after subject enrollment).
+ *
+ * @return A [CollectedDataSet] containing the matching data points.
+ */
+    suspend fun getCollectedData(
+        studyId: UUID,
```
Whathecode, line 43: This should not live in the data subsystem. It should be agnostic to studies. 

It's also odd you need to pass studyId and can filter further by deployment IDs. What happens when I pass a deployment ID which isn't part of that study?

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/StudyDataService.kt
```diff
@@ -0,0 +1,46 @@
...
+        override val apiVersion: ApiVersion = API_VERSION
+    }
+
+/**
+ * Retrieves collected data points for a study, filtered by optional parameters.
+ *
+ * @param studyId The unique identifier of the study to query. (Required)
+ * @param query Optional filters for the data query, including:
+ *   - studyDeploymentIds: List of subject deployment IDs to filter by.
+ *   - fields: List of CARP Data Type fields (e.g., `step_count`) to include.
```
Whathecode, line 34: Why strings, and not `DataType`? The documentation also implies the namespace is ignored/can't be specified.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/StudyDataService.kt
```diff
@@ -0,0 +1,46 @@
...
+ *   - deviceRoleNames: List of device role names (e.g., `phone`) to include.
+ *   - from: Absolute start time for filtering.
```
Whathecode, line 36: Inclusive of exclusive?

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/StudyDataService.kt
```diff
@@ -0,0 +1,46 @@
...
+ *   - to: Absolute end time for filtering.
+ *   - offsetDays: Relative offset window (e.g., first 30 days after subject enrollment).
```
Whathecode, line 38: Use type safety, and in addition providing more flexibility (not just days), instead, by using `Duration`.

But, this feels problematic again, since this subsystem should have no notion of study start times. You could use time since the first synchronized data point in any of the filtered out data streams, but that complexity indicates why it doesn't really belong here.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/db/StudyDataRepository.kt
```diff
@@ -0,0 +1,23 @@
...
+ */
+interface StudyDataRepository
+{
+/**
+ * Query collected data points based on study and optional filters.
+ *
+ * @param studyId The ID of the study to query (required).
+ * @param query The query object containing optional filters such as subjectDeploymentIds, deviceRoleNames, dataTypeNames, from, to, and offsetDays.
+ * @return A list of matching [CollectedDataPoint]s.
+ */
```
Whathecode, line 18: A repository interface in which you can only query is suspect. It means it relies on implicit side effects to populate it.

carp.data.core/src/commonTest/kotlin/dk/cachet/carp/data/infrastructure/StudyDataServiceDecoratorTest.kt
```diff
@@ -0,0 +1,52 @@
...
+        {
+            calledWith = StudyDataServiceRequest.GetCollectedData(
+                studyId, query
+            )
+            return CollectedDataSet() // Return empty for now
+        }
+    }
+
+    @Test
+    fun testGetCollectedDataIsDelegated() = runTest {
```
Whathecode, line 34: Use underscores in test names (see all other tests).

And, no point starting every test with "test". Follow the structure "method_....." so that at a glance you can identify the SUT.

carp.data.core/src/commonTest/kotlin/dk/cachet/carp/data/infrastructure/StudyDataServiceRequestTest.kt
```diff
@@ -0,0 +1,59 @@
+package dk.cachet.carp.data.infrastructure
+
+import dk.cachet.carp.common.application.UUID
+import dk.cachet.carp.data.application.CollectedDataQuery
+import kotlinx.datetime.Instant
+import kotlinx.serialization.json.Json
+import kotlin.test.*
+
+class StudyDataServiceRequestTest
```
Whathecode, line 9: I don't remember how exactly I tested the other services, but look at those to get a feel for the right level of coverage and test setups. 

Align them/be consistent, unless there is a specific reason to do something different. 

I seem to recall adding more generic test setups through the use of base classes to more quickly provide coverage.

carp.data.core/src/jvmTest/kotlin/dk/cachet/carp/data/infrastructure/db/SQLiteStudyDataRepositoryTest.kt
```diff
@@ -0,0 +1,73 @@
+package dk.cachet.carp.data.infrastructure.db
+
+import dk.cachet.carp.common.application.UUID
+import dk.cachet.carp.data.application.CollectedDataQuery
+import kotlinx.coroutines.test.runTest
+import kotlinx.datetime.Instant
+import kotlin.test.*
+
+class SQLiteStudyDataRepositoryTest
```
Whathecode, line 9: Look at how I use interfaces to test against for my in memory implementations. The same tests should be able to run for different concrete instances. Core's primary task is clearly defining the contract, not providing the implementation.

rpc/src/main/kotlin/dk/cachet/carp/rpc/GenerateExampleRequests.kt
```diff
@@ -490,5 +501,16 @@ private val exampleRequests: Map<KFunction<*>, LoggedRequest.Succeeded<*>> = map
     DataStreamService::removeDataStreams to example(
         request = DataStreamServiceRequest.RemoveDataStreams( deploymentIds ),
         response = deploymentIds
+    ),
+    // StudyDataService
```
Whathecode, line 505: Missing newline above (similar to the other code blocks for other services).

rpc/src/main/kotlin/dk/cachet/carp/rpc/GenerateExampleRequests.kt
```diff
@@ -490,5 +501,16 @@ private val exampleRequests: Map<KFunction<*>, LoggedRequest.Succeeded<*>> = map
...
+    StudyDataService::getCollectedData to example(
+        request = StudyDataServiceRequest.GetCollectedData(
+            studyId = studyId,
+            query = CollectedDataQuery()
+        ),
+        response = exampleCollectedData
+    ),
+
+
```
Whathecode, line 514: excessive newlines

Whathecode (changes requested):
Overall, I'm skeptical this belongs in the data subsystem. The main problem is this subsystem should have no business thinking about "studies". Introducing this starts entangling the two subsystems and thus deviates from the intended separation of concerns.

What _is_ likely in scope is providing access to synchronized data, and thus enabling time-based queries, which seems to be part of what you are hoping to achieve here. 

I think that is this issue: https://github.com/cph-cachet/carp.core-kotlin/issues/478

rpc/src/test/kotlin/dk/cachet/carp/rpc/JsonSchemasTest.kt
```diff
@@ -33,6 +33,14 @@ class JsonSchemasTest
                 // Validate request.
                 val requestJson = mapper.readTree( r.requestObject.json )
                 val requestErrors = requestSchema.validate( requestJson )
+                if (requestErrors.isNotEmpty())
+                {
+                    println("Schema mismatch for ${r.requestObject.klass}")
```
Whathecode, line 38: If you want this, we can easily make that a separate commit/PR. Do note copilot's comment. 

I'd have to run this and see example test output to better understand why this is better. 🙂

PR #511, opened 2025-07-02
Add `protocolVersion` field in StudyDetails

### LINKED ISSUE #446: Need to know the version tag for a StudyProtocolSnapshot
In order to show it to the user in the web portal, we also need the version tag - which currently is not part of the snapshot. 

See >> https://github.com/cph-cachet/carp.core-kotlin/blob/develop/carp.protocols.core/src/commonMain/kotlin/dk/cachet/carp/protocols/application/StudyProtocolSnapshot.kt

yuanchen233:
Closes #446

### DIFF
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/infrastructure/versioning/ApplicationServiceApiMigrator.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/infrastructure/versioning/ApplicationServiceApiMigrator.kt
@@ -63,7 +63,7 @@ class ApplicationServiceApiMigrator<
         // Defer applying response migrations.
         fun downgradeResponse( response: JsonElement?, ex: Exception? ): JsonElement
         {
-            val updatedResponse = toApply.fold( ApiResponse( response, ex ) ) { curResponse, migration ->
+            val updatedResponse = toApply.reversed().fold( ApiResponse( response, ex ) ) { curResponse, migration ->
                 migration.migrateResponse( request, curResponse, requestVersion )
             }
 
--- a/carp.protocols.core/src/commonMain/kotlin/dk/cachet/carp/protocols/application/VersionedStudyProtocolSnapshot.kt
+++ b/carp.protocols.core/src/commonMain/kotlin/dk/cachet/carp/protocols/application/VersionedStudyProtocolSnapshot.kt
@@ -1,0 +1,22 @@
+package dk.cachet.carp.protocols.application
+
+import dk.cachet.carp.protocols.domain.StudyProtocol
+import kotlinx.serialization.Serializable
+import kotlin.js.JsExport
+
+/**
+    * A serializable snapshot of a [StudyProtocol] and its version, used in a study.
+ */
+@Serializable
+@JsExport
+data class VersionedStudyProtocolSnapshot (
+    /**
+     * A snapshot of the protocol to use in this study, or null when not yet defined.
+     */
+    val protocolSnapshot: StudyProtocolSnapshot,
+    /**
+     * The version of the protocol to use in this study, or null when not yet defined.
+     */
+    val protocolVersion: ProtocolVersion
+)
+
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyDetails.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyDetails.kt
@@ -4,6 +4,7 @@ package dk.cachet.carp.studies.application
 
 import dk.cachet.carp.common.application.UUID
 import dk.cachet.carp.deployments.application.users.StudyInvitation
+import dk.cachet.carp.protocols.application.ProtocolVersion
 import dk.cachet.carp.protocols.application.StudyProtocolSnapshot
 import kotlinx.datetime.Instant
 import kotlinx.serialization.*
@@ -40,5 +41,9 @@ data class StudyDetails(
     /**
      * A snapshot of the protocol to use in this study, or null when not yet defined.
      */
-    val protocolSnapshot: StudyProtocolSnapshot?
+    val protocolSnapshot: StudyProtocolSnapshot?,
+    /**
+     * The version of the protocol to use in this study, or null when not yet defined.
+     */
+    val protocolVersion: ProtocolVersion? = null
 )
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyDetails.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyDetails.kt
@@ -4,10 +4,9 @@ package dk.cachet.carp.studies.application
 
 import dk.cachet.carp.common.application.UUID
 import dk.cachet.carp.deployments.application.users.StudyInvitation
-import dk.cachet.carp.protocols.application.ProtocolVersion
-import dk.cachet.carp.protocols.application.StudyProtocolSnapshot
+import dk.cachet.carp.protocols.application.VersionedStudyProtocolSnapshot
 import kotlinx.datetime.Instant
-import kotlinx.serialization.*
+import kotlinx.serialization.Serializable
 import kotlin.js.JsExport
 
 
@@ -39,11 +38,7 @@ data class StudyDetails(
      */
     val invitation: StudyInvitation,
     /**
-     * A snapshot of the protocol to use in this study, or null when not yet defined.
+     * A versioned snapshot of the protocol to use in this study, or null when not yet defined.
      */
-    val protocolSnapshot: StudyProtocolSnapshot?,
-    /**
-     * The version of the protocol to use in this study, or null when not yet defined.
-     */
-    val protocolVersion: ProtocolVersion? = null
+    val versionedProtocolSnapshot: VersionedStudyProtocolSnapshot?,
 )
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/versioning/StudyServiceApiMigrator.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/versioning/StudyServiceApiMigrator.kt
@@ -1,13 +1,13 @@
 package dk.cachet.carp.studies.infrastructure.versioning
 
 import dk.cachet.carp.common.application.services.ApiVersion
-import dk.cachet.carp.common.infrastructure.versioning.ApiResponse
-import dk.cachet.carp.common.infrastructure.versioning.ApplicationServiceApiMigrator
-import dk.cachet.carp.common.infrastructure.versioning.Major1Minor0To1Migration
-import dk.cachet.carp.common.infrastructure.versioning.getType
+import dk.cachet.carp.common.infrastructure.serialization.JSON
+import dk.cachet.carp.common.infrastructure.versioning.*
+import dk.cachet.carp.protocols.application.ProtocolVersion
 import dk.cachet.carp.studies.application.StudyService
 import dk.cachet.carp.studies.infrastructure.StudyServiceInvoker
 import dk.cachet.carp.studies.infrastructure.StudyServiceRequest
+import kotlinx.datetime.Clock
 import kotlinx.serialization.json.*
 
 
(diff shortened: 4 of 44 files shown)


carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyDetails.kt
```diff
@@ -40,5 +41,9 @@ data class StudyDetails(
     /**
      * A snapshot of the protocol to use in this study, or null when not yet defined.
      */
-    val protocolSnapshot: StudyProtocolSnapshot?
+    val protocolSnapshot: StudyProtocolSnapshot?,
+    /**
+     * The version of the protocol to use in this study, or null when not yet defined.
+     */
+    val protocolVersion: ProtocolVersion? = null
```
Whathecode: If both are either `null` or not `null` at the same time, make it a single property to prevent illegal state where one is `null` and the other one isn't. We'd have to spar a bit on the name of that, maybe `VersionedStudyProtocolSnapshot`.

However, that also shows there may be a backwards compatibility problem. If the version isn't allowed to be `null` when the protocol is set, you can't migrate existing data without querying the protocols subsystem and finding an exact match among the different versions for the given `protocolId`.

For API migration, I think we are okay, as this type only seems to be _returned_ from `StudyService` and doesn't act as input, so it can simply be removed.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/versioning/StudyServiceApiMigrator.kt
```diff
@@ -51,11 +48,35 @@ private val major1Minor0To1Migration =
...
+        override fun migrateEvent( event: JsonObject ): JsonObject = event
+    }
 
 val StudyServiceApiMigrator = ApplicationServiceApiMigrator(
     StudyService.API_VERSION,
     StudyServiceInvoker,
     StudyServiceRequest.Serializer,
-    StudyService.Event.serializer(),
-    listOf( major1Minor0To1Migration )
+    StudyService.Event.serializer( ),
```
Whathecode: Remove unnecessarily added space.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/versioning/StudyServiceApiMigrator.kt
```diff
@@ -51,11 +48,35 @@ private val major1Minor0To1Migration =
...
+                {
+                    val responseObject = ( response.response as? JsonObject )
+                        ?.migrate {
+                            json.remove( "protocolVersion" )
+                        }
+                    ApiResponse( responseObject, response.ex )
+                }
+                else -> response
+        }
+        override fun migrateEvent( event: JsonObject ): JsonObject = event
```
Whathecode: `StudyCreated` and `StudyGoneLive` are also impacted, it seems. In previous migrations this seems to have _added_ new data where needed (I was missing some documentation here whether this was up or downwards migration :) ). It is used in `BackwardsCompatibilityTest` to migrate old recorded events to new events to verify whether everything passes on the latest version.

Without that migration, I would actually expect tests to fail, which interestingly, they don't. 🤔 Maybe due to current nullability, and it allowed to be null, and no check being in place that the protocol and version both need to be null, or not null.

Theoretically, that migration should be done if we want to run an older `StudyService` with a newer `RecruitmentService`. In practice, that will never happen (this is the same subsystem, so would be deployed in one go). So, I think this migration is primarily there for the `BackwardsCompatibilityTest`. The upgrader can thus provide stubbed data to please that test, as the event consumer doesn't rely on it other than just storing it for presentation purposes. Do make a comment of that in the migrator for clarity!

Whathecode (changes requested):

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyDetails.kt
Whathecode: > you can't migrate existing data without querying the protocols subsystem and finding an exact match among the different versions for the given protocolId.

So this would need to be migration info going into the release notes. Do we have current production data at DTU which needs to survive an upgrade from CARP 1.2 to CARP 1.3?

yuanchen233: > If both are either `null` or not `null` at the same time, make it a single property to prevent illegal state where one is `null` and the other one isn't. We'd have to spar a bit on the name of that, maybe `VersionedStudyProtocolSnapshot`.
> 

Yes, since it only make sense to set a protocol version when there is a protocol. 
We can set the new  `VersionedStudyProtocolSnapshot` as 
```
class VersionedStudyProtocolSnapshot {
  val studyProtocolSnapshot: studyProtocolSnapshot
  val protocolVersion: ProtocolVersion?
}
```
To allow the version to be `null`.
We can also make `StudyProtocolSnapshot` and `VersionedStudyProtocolSnapshot` co-exist and deprecate the former for a easy migration, if we want to make sure version is not null when protocol is set.

> So this would need to be migration info going into the release notes. Do we have current production data at DTU which needs to survive an upgrade from CARP 1.2 to CARP 1.3?

We do have production data, but I can make some temporary hacks on CAWS for compatibility. (although we already have hacks there, hopefully they can be removed after this patch) The `protocolVersion` is not used or relied on much since it was not exposed.

Whathecode: > We can also make StudyProtocolSnapshot and VersionedStudyProtocolSnapshot co-exist and deprecate the former for a easy migration

The point of version migration is that you don't have to do that. 

And, instead of adding in hacks, I suggest just doing data migration as I suggest, i.e., fill in the missing data based on data which should be retrievable from the protocols DB. 🤷🏼‍♂️

That's probably as much work as a hack, and doubt any related DB downtime (if any) will bother many.

yuanchen233: That makes sense, I'll do data migration when doing upgrade.

Whathecode: You can already write the script. But, be aware that some of the other PRs planned for 1.3 may also require migration. 😉

yuanchen233: Thanks, I'll definitely test it well before applying them :)
I made a separate commit for this, as there were more changes than i expected, feel free to squash them.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/infrastructure/versioning/ApplicationServiceApiMigrator.kt
```diff
@@ -63,7 +63,7 @@ class ApplicationServiceApiMigrator<
         // Defer applying response migrations.
         fun downgradeResponse( response: JsonElement?, ex: Exception? ): JsonElement
         {
-            val updatedResponse = toApply.fold( ApiResponse( response, ex ) ) { curResponse, migration ->
+            val updatedResponse = toApply.reversed().fold( ApiResponse( response, ex ) ) { curResponse, migration ->
```
Whathecode, line 66: This should  be a separate bugfix commit, explaining in the description what was wrong here. And, the current two commits, excluding this change, should be squashed.

carp.protocols.core/src/commonMain/kotlin/dk/cachet/carp/protocols/application/VersionedStudyProtocolSnapshot.kt
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+    * A serializable snapshot of a [StudyProtocol] and its version, used in a study.
+ */
+@Serializable
+@JsExport
+data class VersionedStudyProtocolSnapshot (
+    /**
+     * A snapshot of the protocol to use in this study, or null when not yet defined.
+     */
```
Whathecode, line 15: Wrong documentation. This is the behavior you just changed!  Also for version.

carp.protocols.core/src/commonMain/kotlin/dk/cachet/carp/protocols/application/VersionedStudyProtocolSnapshot.kt
```diff
@@ -0,0 +1,22 @@
+package dk.cachet.carp.protocols.application
+
+import dk.cachet.carp.protocols.domain.StudyProtocol
+import kotlinx.serialization.Serializable
+import kotlin.js.JsExport
+
+/**
+    * A serializable snapshot of a [StudyProtocol] and its version, used in a study.
```
Whathecode, line 8: Don't refer to things which should be hidden from this subsystem in documentation, i.e., "used in a study". Let the caller (which could be many other things than the studies subsystem) decide how to use this.

Documentation should be as encapsulated as the code it documents. 

This also begs the question: _should_ this be defined here, as opposed to in the studies subsystem? Do you anticipate it will be used here (plausible; I'm just wondering whether you gave it thought)?

There is also weird indenting here, at least on GitHub. And, try to refer to members in documentation using the [] syntax; it helps with good naming. E.g., given the type name, just `protocol` and `version` probably suffices, which can more easily be used in a documentation sentence as well.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyDetails.kt
```diff
@@ -38,7 +38,7 @@ data class StudyDetails(
      */
     val invitation: StudyInvitation,
     /**
-     * A snapshot of the protocol to use in this study, or null when not yet defined.
```
Whathecode, line 41: The snapshot isn't versioned. The protocol has a version.

Whathecode (changes requested):
This now clearly shows that this is a backwards incompatible change. 

You can't just "make up" data for old requests. The migrator assumes the version is "Initial", which you assume since you know an implementation detail of an optional default defined elsewhere. But it can literally be anything else. 

I overlooked the "setting" part of this request. I believe we need to go back to the drawing board. The fact that so many files had to be changed also shows it may not be the right place to do this. These files shouldn't change just because of a fairly simple added requirement to show something extra in a single place. 

Instead, a backwards compatible change would be a new request in the protocols subsystem to retrieve the version for a given protocol. 🤷🏼‍♂️ Probably a much much smaller change, which can be called directly by the studies subsystem, and furthermore would be validated, in contrast to the current solution which simply takes in data and forwards it again. In addition, it even recognizes the current decoupled nature between studies and the protocols subsystem: you can pass _any_ protocol to a study, it **doesn't even need to be in the protocols subsystem!** So the UI will also need to handle the case if the protocol isn't found in that subsystem.

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/infrastructure/versioning/ApplicationServiceApiMigrator.kt
Whathecode: Given the final verdict for this PR, which will involve rolling back almost everything _except_ this change, you can see the usefulness of keeping distinct changes separate. 😉

yuanchen233:
> You can't just "make up" data for old requests. The migrator assumes the version is "Initial", which you assume since you know an implementation detail of an optional default defined elsewhere. But it can literally be anything else.
> 
How about making this `snapshotVersion` nullable? Or say "Version Unknown"?

> I overlooked the "setting" part of this request. I believe we need to go back to the drawing board. The fact that so many files had to be changed also shows it may not be the right place to do this. These files shouldn't change just because of a fairly simple added requirement to show something extra in a single place.
> 
That's fair, and also why I keep this as a separated commit, I felt I was going into the wrong direction. It was meant for a draft for discussion, my bad for marking it as ready. (thanks for the quick, late night comments by the way. )
 
> Instead, a backwards compatible change would be a new request in the protocols subsystem to retrieve the version for a given protocol. 🤷🏼‍♂️ Probably a much much smaller change, which can be called directly by the studies subsystem, and furthermore would be validated, in contrast to the current solution which simply takes in data and forwards it again. In addition, it even recognizes the current decoupled nature between studies and the protocols subsystem: you can pass _any_ protocol to a study, it **doesn't even need to be in the protocols subsystem!** So the UI will also need to handle the case if the protocol isn't found in that subsystem.

I'm not following this well. As per my understanding, a `protocol` have different versions sharing the same `id`. Therefore `id` and `versionTag` together are compound keys to uniquely identify a "versionedProtocol". The problem we are currently facing is, we don't keep `versionTag` information inside a study. I can't think of a easy way to get the `versionTag` from protocol subsystem for a given protocol, besides comparing the snapshots one by one.

What is the purpose of `versionTag`? If it is for helping researchers managing different versions of same protocol through out a study, then the researcher should have access to it through study. If it is only there for keeping a history, it then should not be exposed, and we should rely less on it in the protocol subsystem.

PR #513, opened 2025-08-06
Include `DeviceRegistration` in `DeviceDeploymentStatus` (slight rework)

Whathecode:
I'm taking @yuanchen233's [PR](https://github.com/cph-cachet/carp.core-kotlin/pull/489) and reworking it a bit based on what we learned, and maintaining a bit cleaner commit history.

In addition to Yuan's original PR, this fixes `stoppedOn` not being stored on `StudyDeploymentSnapshot` (a bug observed in the other PR), in order to make newly added tests which increase coverage for `DeploymentService`'s `BackwardsCompatibilityTest` pass. **Important: this will require a database migration!** The `isStopped` field in `StudyDeploymentSnapshot` needs to be replaced with an `Instant` (or `null` if not yet stopped). Unfortunately, the actual time when the deployment stopped can't be retrieved, so a heuristic will need to be used, such as stopping it a fixed amount of time after it has started. Or, if it truly matters, a more complex solution would be verifying when the last data point was received. **We should remember to document this in the release notes!**

Lastly, I updated the sonatype publication endpoints and updated the secrets in the repo, as the old endpoints have been deprecated since June. That should mean that snapshot (and Maven) publications should start working again after this PR is merged.

Missing:
- there is no JSON schema coverage validation for `DeviceDeploymentStatus.NeedsRedeployment`. However, adding that now is non-trivial. We could maybe run validation on the output of test requests for the backwards compatibility tests in the future (#514).
- Yuan [tried to add coverage](https://github.com/cph-cachet/carp.core-kotlin/pull/489/commits/ac8c4a1ec99be5b281c33fe5555a31442b4df3e4) for `StudyDeploymentStatus` which is output as part of `ParticipantGroupStatus` in `RecruitmentService`. But, as he noted, that doesn't work since to add coverage for `DeviceDeploymentStatus.Registered`, separate calls need to be done to `DeploymentService` which aren't logged by the current setup of `BackwardsCompatibilityTest`. That's a bigger shortcoming of the current setup (#515), but out of scope for this PR.


yuanchen233 (approved):

Whathecode:
@yuanchen233 The snapshot publication failed. But, it failed with a 403, which is progress from the previous 401. On another repo I successfully managed to publish a release, so I think all should be configured correctly. Seemingly, on Sonatype Central you now need to "enable snapshot" publishing per namespace, which I just tried and re-triggered the publication. 🤞

Issue #319, opened 2021-10-08
Endpoints to edit staged participant groups


Whathecode:
The backend now has the notion of "staged" participant groups: sets of participants with assigned master devices _before_ they are deployed. https://github.com/cph-cachet/carp.core-kotlin/issues/255

But, we are still missing endpoints to add/remove/change device assignments in the application service.

Whathecode:
Once these requests are introduced, `RecruitmentService.inviteNewParticipantGroup` should be marked as obsolete. The concurrent implementation can lead to errors when the dependent `DeploymentService` is down. Eventual consistency should be used instead (#295).

yuanchen233:
I'm not very clear with the plan for staged participation group, it seems to me this notion have 3 purpose:
1.  keep track of all deployments of the current study: `Recruitmetn.participantGroups`
2. manage participant groups before they are invited, and keep data like `AssignedParticipantRoles` or `connectedDevicePreregistrations`
3. hold personal information: #479 

For (2), those information should be removed after the group is invited, since Deployment is then responsible for role/device assignments.
For (3), it sounds redundant with `Deployments.ParticipantGroup`, since it contains consent and participant data, but we also have issue on #464 

Other things to consider:
- combine participant group from study and deployment, and Introduce State Machine like participant group status. right now `Deployment.ParticipantGroup` in created and invited when `Deployment` is created through `ParticipantGroupService.createAndInviteParticipantGroup`. It can go other-way around, we can create participant group, then later on invite it, and create study deployment when the group is invited.
- separate repository/snapshot for staged participant group, otherwise it will result in very frequent update to `Recruitment Repository` as the number of deployments increase.
- not expose `DeploymentServiceRequest.CreateStudyDeployment`, study/recruitment can not track deployments created directly through this method. same thing might also applies to `Stop`.

yuanchen233:
`Recruitment` subsystem now has new endpoints:

- `CreateParticipantGroup`
- `UpdateParticipantGroup`
- `InviteParticipantGroup`

and `InviteNewParticipantGroup` is now deprecated.

> separate repository/snapshot for staged participant group, otherwise it will result in very frequent update to Recruitment Repository as the number of deployments increase.

This is tracked #525 

> not expose DeploymentServiceRequest.CreateStudyDeployment, study/recruitment can not track deployments created directly through this method. same thing might also applies to Stop.

No, `Deployment` should function as a stand alone system.

PR #516, opened 2025-08-14
Add `name` field to StagedParticipantGroup and update related methods

### LINKED ISSUE #479: Need a name for a participant group
Right now a participant group / deployment only has a UUID. This is not very userfriendly, so we need to be able to have a name to the group. In the portal we right now just merge the participants names like this

image

But - we should be able to add a prober name like "Family Bardram".

yuanchen233:
I am not confident on where to set default values, as this field is nullable. Currently I have set default value for `StagedParticipantGroup` constructor and `InviteNewParticipantGroup` request.

### DIFF
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
@@ -62,10 +62,10 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
 
     /**
      * Create a new participant [group] of previously added participants and instantly send out invitations
-     * to participate in the study with the given [studyId].
+     * to participate in the study with the given [studyId] and an optional [name] representing this group.
      *
-     * In case a group with the same participants has already been deployed and is still running (not stopped),
-     * the latest status for this group is simply returned.
+     * In case a group with the same participants and [name] has already been deployed and is still
+     * running (not stopped), the latest status for this group is simply returned.
      *
      * @throws IllegalArgumentException when:
      *  - a study with [studyId] does not exist
@@ -74,7 +74,11 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
      *  - not all necessary participant roles part of the study have been assigned a participant
      * @throws IllegalStateException when the study is not yet ready for deployment.
      */
-    suspend fun inviteNewParticipantGroup( studyId: UUID, group: Set<AssignedParticipantRoles> ): ParticipantGroupStatus
+    suspend fun inviteNewParticipantGroup(
+        studyId: UUID,
+        group: Set<AssignedParticipantRoles>,
+        name: String? = null
+    ): ParticipantGroupStatus
 
     /**
      * Get the status of all deployed participant groups in the study with the specified [studyId].
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/users/ParticipantGroupStatus.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/users/ParticipantGroupStatus.kt
@@ -26,6 +26,11 @@ sealed class ParticipantGroupStatus
     abstract val id: UUID
 
     /**
+     * An optional name to represent the group of participants.
+     */
+    abstract val name: String?
+
+    /**
      * The participants that are part of this group.
      */
     abstract val participants: Set<Participant>
@@ -37,7 +42,8 @@ sealed class ParticipantGroupStatus
     @Serializable
     data class Staged(
         override val id: UUID,
-        override val participants: Set<Participant>
+        override val participants: Set<Participant>,
+        override val name: String? = null
     ) : ParticipantGroupStatus()
 
 
@@ -54,7 +60,8 @@ sealed class ParticipantGroupStatus
              */
             fun fromDeploymentStatus(
                 participants: Set<Participant>,
-                deploymentStatus: StudyDeploymentStatus
+                deploymentStatus: StudyDeploymentStatus,
+                name: String?
             ): InDeployment
             {
                 val id = deploymentStatus.studyDeploymentId
@@ -66,12 +73,20 @@ sealed class ParticipantGroupStatus
                     is StudyDeploymentStatus.Invited,
                     is StudyDeploymentStatus.DeployingDevices ->
                         // If deployment was ready at one point (`startedOn`), consider the study 'Running'.
-                        if ( startedOn == null ) Invited( id, participants, createdOn, deploymentStatus )
-                        else Running( id, participants, createdOn, deploymentStatus, startedOn )
+                        if ( startedOn == null ) Invited( id, participants, createdOn, deploymentStatus, name )
+                        else Running( id, participants, createdOn, deploymentStatus, startedOn, name )
                     is StudyDeploymentStatus.Running ->
-                        Running( id, participants, createdOn, deploymentStatus, checkNotNull( startedOn ) )
+                        Running( id, participants, createdOn, deploymentStatus, checkNotNull( startedOn ), name )
                     is StudyDeploymentStatus.Stopped ->
-                        Stopped( id, participants, createdOn, deploymentStatus, startedOn, deploymentStatus.stoppedOn )
+                        Stopped(
+                            id,
+                            participants,
+                            createdOn,
+                            deploymentStatus,
+                            startedOn,
+                            deploymentStatus.stoppedOn,
+                            name
+                        )
                 }
             }
         }
@@ -97,7 +112,8 @@ sealed class ParticipantGroupStatus
         override val id: UUID,
         override val participants: Set<Participant>,
         override val invitedOn: Instant,
-        override val studyDeploymentStatus: StudyDeploymentStatus
+        override val studyDeploymentStatus: StudyDeploymentStatus,
+        override val name: String? = null
     ) : InDeployment()
 
     /**
@@ -113,7 +129,8 @@ sealed class ParticipantGroupStatus
         /**
          * The time when the study deployment started running, i.e., when all devices were deployed for the first time.
          */
-        val startedOn: Instant
+        val startedOn: Instant,
+        override val name: String? = null
     ) : InDeployment()
 
     /**
@@ -134,6 +151,7 @@ sealed class ParticipantGroupStatus
         /**
          * The time when the study deployment was stopped.
          */
-        val stoppedOn: Instant
+        val stoppedOn: Instant,
+        override val name: String? = null
     ) : InDeployment()
 }
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
@@ -174,18 +174,23 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
     private val _participantGroups: MutableMap<UUID, StagedParticipantGroup> = mutableMapOf()
 
     /**
-     * Create and add the participants identified by [participantIds] as a participant group.
+     * Create and add the participants identified by [participantIds] as a participant group, optionally with a [name]
+     * representing this group.
      *
      * @throws IllegalArgumentException when one or more of the participants aren't in this recruitment.
      * @throws IllegalStateException when the study is not yet ready for deployment.
      */
-    fun addParticipantGroup( participantIds: Set<UUID>, id: UUID = UUID.randomUUID() ): StagedParticipantGroup
+    fun addParticipantGroup(
+        participantIds: Set<UUID>,
+        name: String? = null,
+        id: UUID = UUID.randomUUID()
+    ): StagedParticipantGroup
     {
         require( participants.map { it.id }.containsAll( participantIds ) )
             { "One of the participants for which to create a participant group isn't part of this recruitment." }
         check( getStatus() is RecruitmentStatus.ReadyForDeployment ) { "The study is not yet ready for deployment." }
 
-        val group = StagedParticipantGroup( id )
+        val group = StagedParticipantGroup( id, name )
         group.addParticipants( participantIds )
 
         _participantGroups[ group.id ] = group
@@ -206,7 +211,11 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
             { "A study deployment with ID \"$deploymentId\" is not part of this recruitment." }
 
         val participants = group.participantIds.map { id -> _participants.first { it.id == id } }
-        return ParticipantGroupStatus.InDeployment.fromDeploymentStatus( participants.toSet(), studyDeploymentStatus )
+        return ParticipantGroupStatus.InDeployment.fromDeploymentStatus(
+            participants.toSet(),
+            studyDeploymentStatus,
+            group.name
+        )
     }
 
     /**
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/RecruitmentServiceRequest.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/RecruitmentServiceRequest.kt
@@ -65,7 +65,11 @@ sealed class RecruitmentServiceRequest<out TReturn> : ApplicationServiceRequest<
     }
 
     @Serializable
-    data class InviteNewParticipantGroup( val studyId: UUID, val group: Set<AssignedParticipantRoles> ) :
+    data class InviteNewParticipantGroup(
+        val studyId: UUID,
+        val group: Set<AssignedParticipantRoles>,
+        val name: String? = null
+    ) :
         RecruitmentServiceRequest<ParticipantGroupStatus>()
     {
         override fun getResponseSerializer() = serializer<ParticipantGroupStatus>()
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/versioning/RecruitmentServiceApiMigrator.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/versioning/RecruitmentServiceApiMigrator.kt
@@ -43,30 +43,36 @@ private val major1Minor2To3Migration =
                 "dk.cachet.carp.studies.infrastructure.RecruitmentServiceRequest.StopParticipantGroup" ->
                 {
                     val responseObject = response.response?.jsonObject?.migrate {
-                        removeDeviceRegistration()
+                        updateParticipantGroupStatus()
                     }
                     ApiResponse( responseObject, response.ex )
                 }
                 "dk.cachet.carp.studies.infrastructure.RecruitmentServiceRequest.GetParticipantGroupStatusList" ->
                 {
                     val responseObject = response.response?.jsonArray?.migrate {
-                        objects { removeDeviceRegistration() }
+                        objects {
+                            updateParticipantGroupStatus()
+                        }
                     }
                     ApiResponse( responseObject, response.ex )
                 }
                 else -> response
             }
 
-        /**
-         * Remove the newly added `deviceRegistration` field from `DeviceDeploymentStatus` objects contained within
-         * `StudyDeploymentStatus`.
-         */
-        private fun ApiJsonObjectMigrationBuilder.removeDeviceRegistration() =
+
+        private fun ApiJsonObjectMigrationBuilder.updateParticipantGroupStatus()
+        {
+            // Remove the newly added `deviceRegistration` field from `DeviceDeploymentStatus` objects
+            // contained within `StudyDeploymentStatus`.
             updateObject( "studyDeploymentStatus" ) {
                 updateArray( "deviceStatusList" ) {
                     objects { json.remove( "deviceRegistration" ) }
                 }
             }
+
+            // Remove newly added 'name' field from 'ParticipantGroupStatus'.
+            json.remove( "name" )
+        }
 
         override fun migrateEvent( event: JsonObject ) = event
     }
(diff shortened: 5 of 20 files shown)


Whathecode:
> I am not confident on where to set default values, as this field is nullable.

I don't think we want to set a default value in the domain layer. On a UI level, we want to know whether it is set or not, and if not, the UI can revert to e.g., showing a list of participant names. So, in my head, the default was a UI concern only.

That also reflects [what I wrote in the ticket before](https://github.com/cph-cachet/carp.core-kotlin/issues/479#issuecomment-2173036610):
> The front-end could render a default name concatenating the currently assigned participants to the group in case the field is null. Making it a required field seems unnecessary.

**UPDATE:** after reviewing I now understand you were referring to setting a default of `null`. As stated in review comments, I would do it in more places so that you don't have to update so many tests, just to pass an irrelevant argument.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/users/ParticipantGroupStatus.kt
```diff
@@ -25,6 +25,11 @@ sealed class ParticipantGroupStatus
      */
     abstract val id: UUID
 
+    /**
+     * The name of this participant group.
```
Whathecode: Document optionality, and what the name refers to. E.g.:

> An optional name to represent the group of participants.

Also for `StagedParticipantGroup`.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/users/ParticipantGroupStatus.kt
```diff
@@ -37,6 +42,7 @@ sealed class ParticipantGroupStatus
     @Serializable
     data class Staged(
         override val id: UUID,
+        override val name: String?,
```
Whathecode: Since it's optional, it's better to have it as the last parameter, and make it an optional parameter (assigning null by default).
Same comment for the other constructors, factory methods, and application service calls.
That would also prevent you from having to pass a name in a whole lot of places where the name doesn't matter, e.g. `ParticipantGroupStatusTest` and `HostsIntegrationTest`.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
```diff
@@ -74,7 +74,11 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
      *  - not all necessary participant roles part of the study have been assigned a participant
      * @throws IllegalStateException when the study is not yet ready for deployment.
      */
-    suspend fun inviteNewParticipantGroup( studyId: UUID, group: Set<AssignedParticipantRoles> ): ParticipantGroupStatus
+    suspend fun inviteNewParticipantGroup(
+        studyId: UUID,
+        name: String?,
```
Whathecode: From the documentation:

> In case a group with the same participants has already been deployed and is still running (not stopped), the latest status for this group is simply returned.

By adding name in the mix, we need to decide what the contract here is. I guess you could create multiple participant groups with the same participants but differing name now.
If so, we should update the documentation here to clarify. Currently, it is ambiguous.

Likewise, I suppose the `name` doesn't need to be unique. You can have the same name for differing participant groups. That may be a bit odd, though, from a UI perspective, but I guess there is no reason to prevent it here in core.

This also warrants additional test cases in `RecruitmentServiceTest`.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -179,13 +179,17 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
      * @throws IllegalArgumentException when one or more of the participants aren't in this recruitment.
      * @throws IllegalStateException when the study is not yet ready for deployment.
      */
-    fun addParticipantGroup( participantIds: Set<UUID>, id: UUID = UUID.randomUUID() ): StagedParticipantGroup
+    fun addParticipantGroup(
+        participantIds: Set<UUID>,
+        id: UUID = UUID.randomUUID(),
```
Whathecode: It is bad style to have optional arguments in the middle of argument lists. Also mentioned elsewhere, name should probably be `null` by default, solving this issue here.

And, given that it's more likely to set `name` than `id` manually, probably you want to have `name` before `id`.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/RecruitmentServiceRequest.kt
```diff
@@ -65,7 +65,11 @@ sealed class RecruitmentServiceRequest<out TReturn> : ApplicationServiceRequest<
     }
 
     @Serializable
-    data class InviteNewParticipantGroup( val studyId: UUID, val group: Set<AssignedParticipantRoles> ) :
+    data class InviteNewParticipantGroup(
+        val studyId: UUID,
+        val name: String? = null,
```
Whathecode: It's bad style to let mandatory parameters follow optional arguments.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceTest.kt
```diff
@@ -106,7 +106,7 @@ interface RecruitmentServiceTest
         val participant = recruitmentService.addParticipant( studyId, EmailAddress( "test@test.com" ) )
 
         val assignParticipant = AssignedParticipantRoles( participant.id, AssignedTo.All )
-        val groupStatus = recruitmentService.inviteNewParticipantGroup( studyId, setOf( assignParticipant ) )
+        val groupStatus = recruitmentService.inviteNewParticipantGroup( studyId, null, setOf( assignParticipant ) )
```
Whathecode: I suggest to only set a group name for this "happy path" and use null for failure scenarios and where inviting the participant group is not the key behavior being tested. That provides backwards compatibility test coverage for both.
And, here on the happy path, you could add an assertion to see whether the value is set, which seems on par with the current participant assertion.

Note how this removes irrelevant code in the other tests, where setting or not setting the participant name doesn't make a difference.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentTest.kt
```diff
@@ -105,7 +105,7 @@ class RecruitmentTest
         assertTrue( recruitment.getStatus() is RecruitmentStatus.ReadyForDeployment )
 
         val participantIds = setOf( participant.id )
-        val group = recruitment.addParticipantGroup( participantIds )
+        val group = recruitment.addParticipantGroup(participantIds, name = "Test Group")
```
Whathecode: Maybe only pass a name for this happy path, and then also assert it.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/CreateTestObjects.kt
```diff
@@ -34,8 +34,9 @@ fun createComplexRecruitment(): Recruitment
     val studyId = UUID.randomUUID()
     val recruitment = Recruitment( studyId ).apply {
         val participant = addParticipant( EmailAddress( "test@test.com" ) )
+        val name = "Test Group"
         lockInStudy( createComplexProtocol().getSnapshot(), StudyInvitation( "Test" ) )
-        addParticipantGroup( setOf( participant.id ) )
+        addParticipantGroup(setOf( participant.id ), name = name)
```
Whathecode: spacing

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/StudiesCodeSamples.kt
```diff
@@ -61,7 +62,12 @@ class StudiesCodeSamples
             val participation = AssignedParticipantRoles( participant.id, AssignedTo.All )
             val participantGroup = setOf( participation )
 
-            val groupStatus: ParticipantGroupStatus = recruitmentService.inviteNewParticipantGroup( studyId, participantGroup )
+            val groupStatus: ParticipantGroupStatus =
+                recruitmentService.inviteNewParticipantGroup(
+                    studyId,
+                    participantGroupName,
+                    participantGroup
+                )
```
Whathecode: This should match the same code in the `README.md`. These samples are here to make sure they compile. :) However, setting a name is optional and detracts from the main documentation there, so I don't think we need this.

rpc/src/main/kotlin/dk/cachet/carp/rpc/GenerateExampleRequests.kt
```diff
@@ -155,6 +155,7 @@ private val studyLiveStatus = StudyStatus.Live( studyId, studyName, studyCreated
 
 // Deployment data matching the example protocol.
 private val deploymentId = UUID( "c9cc5317-48da-45f2-958e-58bc07f34681" )
+private val deploymentName = "Transportation deployment"
```
Whathecode: The goal for these examples is to be cohesive/representative. This deployment maps to a single participant with account `boaty@mcboatface.com`, so maybe use "Boaty McBoatface" here instead. :)

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/versioning/RecruitmentServiceApiMigrator.kt
```diff
@@ -44,13 +44,17 @@ private val major1Minor2To3Migration =
...
                     }
                     ApiResponse( responseObject, response.ex )
                 }
                 "dk.cachet.carp.studies.infrastructure.RecruitmentServiceRequest.GetParticipantGroupStatusList" ->
                 {
                     val responseObject = response.response?.jsonArray?.migrate {
-                        objects { removeDeviceRegistration() }
+                        objects {
+                            removeDeviceRegistration()
+                            removeName()
```
Whathecode: Duplicate code. Instead, have a single method related to updating participant group status objects and use comments within to indicate the same that is now a method doc.

Whathecode (changes requested):

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/users/ParticipantGroupStatus.kt
yuanchen233: Got it, that's what I did before force-pushing, gladly Github keeps that version as well. I put the `name` filed in the middle as I wanted to match argument order across different method calls, and then updated the test cases as they start failing. I had the idea that only making the `name` field optional on request, but now I see I am over complicating things.

Whathecode: > I had the idea that only making the name field optional on request, but now I see I am over complicating things.

Not at all. :) Worthy considerations.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
yuanchen233: > In case a group with the same participants has already been deployed and is still running (not stopped), the latest status for this group is simply returned.

I added a new test case `inviteNewParticipantGroup_for_already_deployed_group_returns_existing_group`, to take advantage of the new `name` field to test this as well. This might be redundant, as we already `inviteNewParticipantGroup_multiple_times_returns_same_group`, but including a `group name` ensures the status of running group is returned. I kept both test case, as the reply message is different.

> Likewise, I suppose the name doesn't need to be unique. You can have the same name for differing participant groups. That may be a bit odd, though, from a UI perspective, but I guess there is no reason to prevent it here in core.

Yes, this can be limited on infrastructure/implementation level (CAWS). It is nullable here and strictness seems unnecessary. I added a test case `inviteNewParticipantGroup_for_multiple_groups_with_same_name_succeeds`, is this what you are referring to on 'additional test cases' ?

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
```diff
@@ -62,10 +62,10 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
 
     /**
      * Create a new participant [group] of previously added participants and instantly send out invitations
-     * to participate in the study with the given [studyId].
+     * to participate in the study with the given [studyId] and an optional [name] representing this group.
      *
      * In case a group with the same participants has already been deployed and is still running (not stopped),
-     * the latest status for this group is simply returned.
+     * the latest status for this group is simply returned, [name] is ignored in this case.
```
Whathecode: That is different from what I suggested.

>  I guess you could create multiple participant groups with the same participants but differing name now.

I find "ignoring" `name` a highly unintuitive contract. (Why allow passing something which is ignored?) Instead, just add `name` to what makes the group unique. So, it should be possible to create two participant groups with the same participants but differing name.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -179,13 +179,17 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
      * @throws IllegalArgumentException when one or more of the participants aren't in this recruitment.
      * @throws IllegalStateException when the study is not yet ready for deployment.
      */
-    fun addParticipantGroup( participantIds: Set<UUID>, id: UUID = UUID.randomUUID() ): StagedParticipantGroup
+    fun addParticipantGroup(
+        participantIds: Set<UUID>,
+        name: String? = null,
```
Whathecode, line 185: Document `name`.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/versioning/RecruitmentServiceApiMigrator.kt
```diff
@@ -43,30 +43,35 @@ private val major1Minor2To3Migration =
...
+        private fun ApiJsonObjectMigrationBuilder.updateParticipantGroupStatus()
+        {
+            // Remove the newly added `deviceRegistration` field from `DeviceDeploymentStatus` objects
+            // contained within `StudyDeploymentStatus`.
             updateObject( "studyDeploymentStatus" ) {
                 updateArray( "deviceStatusList" ) {
                     objects { json.remove( "deviceRegistration" ) }
                 }
             }
+            // Remove newly added 'name' field from 'ParticipantGroupStatus'.
```
Whathecode, line 73: Newline before to make this a "code paragraph". :)

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/users/ParticipantGroupStatus.kt
```diff
@@ -54,7 +60,8 @@ sealed class ParticipantGroupStatus
              */
             fun fromDeploymentStatus(
                 participants: Set<Participant>,
-                deploymentStatus: StudyDeploymentStatus
+                deploymentStatus: StudyDeploymentStatus,
+                name: String?
```
Whathecode, line 64: I think even this can be given a default `null` parameter. As stated in the first review, then you can prevent modifying `ParticipantGroupStatusTest`.

_If_ you are interested in testing this simple setter, add a new test. (But it's arguably not that interesting.)

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceTest.kt
```diff
@@ -171,6 +173,26 @@ interface RecruitmentServiceTest
...
+        val assignedParticipant = AssignedParticipantRoles( participant.id, AssignedTo.All )
+        val groupStatus1 = recruitmentService.inviteNewParticipantGroup(
+            studyId,
+            setOf( assignedParticipant ),
+            "Group 1"
+        )
+        val groupStatus2 = recruitmentService.inviteNewParticipantGroup(
+            studyId,
+            setOf( assignedParticipant ),
+            "Group 2"
```
Whathecode, line 191: Yeah; I really don't like that you pass a different name and you get the same group back. :) It's a bad/unintuitive contract.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceTest.kt
```diff
@@ -198,6 +220,24 @@ interface RecruitmentServiceTest
...
     }
 
+    @Test
+    fun inviteNewParticipantGroup_for_multiple_groups_with_same_name_succeeds() = runTest {
+        val (recruitmentService, studyService) = createSUT()
+        val (studyId, _) = createLiveStudy( studyService )
+
+        val p1 = recruitmentService.addParticipant( studyId, EmailAddress( "test@test.com" ) )
+        val assignedP1 = AssignedParticipantRoles( p1.id, AssignedTo.All )
+        recruitmentService.inviteNewParticipantGroup( studyId, setOf( assignedP1 ), "Test Group" )
```
Whathecode: Minor: you could rename "Test group" to "Same name" to emphasize this is the same name you are testing here. Or, assign it to a `sameName` variable.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/infrastructure/ParticipantGroupStatusTest.kt
```diff
@@ -21,7 +21,8 @@ class ParticipantGroupStatusTest
         val studyDeploymentId = UUID.randomUUID()
         val deploymentStatus = StudyDeploymentStatus.Invited( Clock.System.now(), studyDeploymentId, listOf(), emptyList(), null )
         val participants = setOf( Participant( AccountIdentity.fromEmailAddress( "test@test.com" ) ) )
-        val groupStatus = ParticipantGroupStatus.InDeployment.fromDeploymentStatus( participants, deploymentStatus )
+        val name = "Test Group"
+        val groupStatus = ParticipantGroupStatus.InDeployment.fromDeploymentStatus(participants, deploymentStatus, name)
```
Whathecode: Missing spaces.

Whathecode (changes requested):

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
Whathecode: The new test looks fine. But, as the new comment on the modified test indicates, I don't like that you disallow the same participants with differing group name. I'll resolve this one and we can follow-up on the other comment.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
yuanchen233: I see what you mean now, I thought that `participants` are the actual identity, and `name` is just a label/metadata. This way, you can't create two participant groups with the same participants but different name, where `name` can be changed later. We do assign an `id` to each group, but `participants` are what keep each deployed groups unique.

> Instead, just add name to what makes the group unique.

I didn't do this for a few reasons:
- As said above, I treated `participants` as identity and I didn't want to change this behaviour.
- `name` can be changed after deployment as suggested in the original issue, then it should not be used as what make groups unique.
- `name` is a nullable field, I think uniqueness around null is a bit unclear/messy to handle.

> I find "ignoring" name a highly unintuitive contract. (Why allow passing something which is ignored?)

This makes a lot of sense, it does feel odd. How about we keep this endpoint unchanged, and setting `name` through another endpoint, like #319

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/users/ParticipantGroupStatus.kt
yuanchen233: Makes sense—changing the test isn’t ideal. I didn’t add a default `null` here because this method builds a `ParticipantGroupStatus` from a `StudyDeploymentStatus`. `name` exists on the former, not the latter, and we only have it in Recruitment. If we default to `null`, a missed argument would silently drop the label and give a status that doesn’t match the actual group. Since groups are always created with either a name or an explicit `null`, I’d rather keep the parameter required so the call is intentional.

If we still want to test the tiny setter, I can add a small focused test instead of touching `ParticipantGroupStatusTest`.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
Whathecode: > name is a nullable field, I think uniqueness around null is a bit unclear/messy to handle.

That's easy/not a concern. 🙂

> name can be changed after deployment as suggested in the original issue, then it should not be used as what make groups unique.

That's a very valid concern. Especially when considering async requests. It's bound to end up in odd edge cases. It's not really identity, though, but used here to make the endpoint idempotent. 

> setting name through another endpoint,

That may make sense! Also considering that the plan is to make the invite endpoint obsolete eventually. I think I linked to that issue as a potential prerequisite before. 🙂

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/users/ParticipantGroupStatus.kt
Whathecode: Without looking at details, your reasoning here makes sense. 👌

And, no need to add test. You can resolve this.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
yuanchen233: > Also considering that the plan is to make the invite endpoint obsolete eventually. I think I linked to that issue as a potential prerequisite before. 

Yes, I decided to go for this one first because I didn't want to add a new endpoint then change it right after. Also I though this endpoint needed to be updated anyway as we are not deprecating it now. However, now I do realise that I can just leave it if new endpoints are ready.

> It's not really identity, though, but used here to make the endpoint idempotent.

Thanks for clarifying, agreed, I'll go ahead and make the change.

Whathecode (approved):
There were just two more missing spaces, which I fixed. Great work 🚀 !

PR #517, opened 2025-08-29
feat(data): add getBatchForStudyDeployments endpoint with filtering

NGrech:
This PR adds a new `getBatchForStudyDeployments()` endpoint to `DataStreamService` that enables efficient retrieval of data from multiple study deployments with optional filtering capabilities.

### New Endpoint
- **`getBatchForStudyDeployments()`**: Retrieve data for multiple deployments in a single call
  - Filter by device role names 
  - Filter by data types
  - Filter by time range (inclusive start, exclusive end)
  - Returns aggregated `DataStreamBatch` with non-overlapping sequences per stream

### Implementation Details
- Time-range filtering using inclusive lower and exclusive upper bound (`[from, to)`)
- Adjusts `firstSequenceId` when measurements are filtered from the beginning
- Maintains sequence ordering and non-overlap guarantees per data stream

### Bug Fixes
- Fixed time filtering to properly adjust sequence IDs when clipping measurements
- Updated RPC example requests to match `ParticipantGroupStatus` constructor changes

### Testing
- Added comprehensive test suite in `InMemoryDataStreamServiceBatchRetrievalTest`
- Tests cover filtering, aggregation, edge cases, and non-monotonic timestamps

### DIFF
--- a/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamBatch.kt
+++ b/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamBatch.kt
@@ -3,6 +3,7 @@
 package dk.cachet.carp.data.application
 
 import dk.cachet.carp.common.application.data.Data
+import dk.cachet.carp.common.application.data.DataType
 import kotlinx.serialization.*
 import kotlinx.serialization.builtins.*
 import kotlinx.serialization.descriptors.*
@@ -90,7 +91,7 @@ class MutableDataStreamBatch : DataStreamBatch
     /**
      * Append all data stream sequences contained in [batch] to this batch.
      *
-     * @throws IllegalArgumentException when the start of any of the sequences contained in [batch]
+     * @throws IllegalArgumentException when the start of the sequences contained in [batch]
      *   precede the end of a previously appended sequence to the same data stream.
      */
     fun appendBatch( batch: DataStreamBatch )
--- a/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamService.kt
+++ b/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamService.kt
@@ -2,10 +2,13 @@ package dk.cachet.carp.data.application
 
 
 import dk.cachet.carp.common.application.UUID
+import dk.cachet.carp.common.application.data.DataType
 import dk.cachet.carp.common.application.services.ApiVersion
 import dk.cachet.carp.common.application.services.ApplicationService
 import dk.cachet.carp.common.application.services.IntegrationEvent
-import kotlinx.serialization.*
+import kotlinx.datetime.Instant
+import kotlinx.serialization.Required
+import kotlinx.serialization.Serializable
 
 
 /**
@@ -59,6 +62,33 @@ interface DataStreamService : ApplicationService<DataStreamService, DataStreamSe
         fromSequenceId: Long,
         toSequenceIdInclusive: Long? = null
     ): DataStreamBatch
+
+    /**
+     * Retrieve data across multiple study deployments, optionally filtered by device role names,
+     * data types, and time range.
+     *
+     * The response is a canonical [DataStreamBatch]: for each [DataStreamId], sequences are
+     * ordered by start time and non-overlapping (contract preserved). No derived/secondary
+     * indexing is applied in this API; analytics-specific projections are out of scope here.
+     *
+     * Time range semantics: if [from] or [to] are specified, sequences are clipped to the
+     * half-open interval [from, to) (inclusive start, exclusive end).
+     *
+     * @param studyDeploymentIds Study deployments to query. Must not be empty.
+     * @param deviceRoleNames Optional device role name filter (e.g., "phone"). If null or empty, all are included.
+     * @param dataTypes Optional data type filter. If null or empty, all are included.
+     * @param from Optional absolute start time (inclusive). If null, no lower bound.
+     * @param to Optional absolute end time (exclusive). If null, no upper bound.
+     * @return A [DataStreamBatch] containing matching data sequences, preserving per-stream invariants.
+     */
+    suspend fun getBatchForStudyDeployments(
+        studyDeploymentIds: Set<UUID>,
+        deviceRoleNames: Set<String>? = null,
+        dataTypes: Set<DataType>? = null,
+        from: Instant? = null,
+        to: Instant? = null
+    ): DataStreamBatch
+
 
     /**
      * Stop accepting incoming data for all data streams for each of the [studyDeploymentIds].
--- a/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/DataStreamServiceDecorator.kt
+++ b/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/DataStreamServiceDecorator.kt
@@ -1,6 +1,7 @@
 package dk.cachet.carp.data.infrastructure
 
 import dk.cachet.carp.common.application.UUID
+import dk.cachet.carp.common.application.data.DataType
 import dk.cachet.carp.common.infrastructure.services.ApplicationServiceDecorator
 import dk.cachet.carp.common.infrastructure.services.ApplicationServiceInvoker
 import dk.cachet.carp.common.infrastructure.services.Command
@@ -8,6 +9,7 @@ import dk.cachet.carp.data.application.DataStreamBatch
 import dk.cachet.carp.data.application.DataStreamId
 import dk.cachet.carp.data.application.DataStreamService
 import dk.cachet.carp.data.application.DataStreamsConfiguration
+import kotlinx.datetime.Instant
 
 
 class DataStreamServiceDecorator(
@@ -32,6 +34,22 @@ class DataStreamServiceDecorator(
         toSequenceIdInclusive: Long?
     ) = invoke( DataStreamServiceRequest.GetDataStream( dataStream, fromSequenceId, toSequenceIdInclusive ) )
 
+    override suspend fun getBatchForStudyDeployments(
+        studyDeploymentIds: Set<UUID>,
+        deviceRoleNames: Set<String>?,
+        dataTypes: Set<DataType>?,
+        from: Instant?,
+        to: Instant?
+    ) = invoke(
+        DataStreamServiceRequest.GetBatchForStudyDeployments(
+        studyDeploymentIds,
+        deviceRoleNames,
+        dataTypes,
+        from,
+        to
+    )
+    )
+
     override suspend fun closeDataStreams( studyDeploymentIds: Set<UUID> ) =
         invoke( DataStreamServiceRequest.CloseDataStreams( studyDeploymentIds ) )
 
@@ -51,5 +69,12 @@ object DataStreamServiceInvoker : ApplicationServiceInvoker<DataStreamService, D
                 service.getDataStream( dataStream, fromSequenceId, toSequenceIdInclusive )
             is DataStreamServiceRequest.CloseDataStreams -> service.closeDataStreams( studyDeploymentIds )
             is DataStreamServiceRequest.RemoveDataStreams -> service.removeDataStreams( studyDeploymentIds )
+            is DataStreamServiceRequest.GetBatchForStudyDeployments -> service.getBatchForStudyDeployments(
+                studyDeploymentIds,
+                deviceRoleNames,
+                dataTypes,
+                from,
+                to
+            )
         }
 }
--- a/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/DataStreamServiceRequest.kt
+++ b/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/DataStreamServiceRequest.kt
@@ -1,6 +1,7 @@
 package dk.cachet.carp.data.infrastructure
 
 import dk.cachet.carp.common.application.UUID
+import dk.cachet.carp.common.application.data.DataType
 import dk.cachet.carp.common.application.services.ApiVersion
 import dk.cachet.carp.common.infrastructure.serialization.ignoreTypeParameters
 import dk.cachet.carp.common.infrastructure.services.ApplicationServiceRequest
@@ -9,6 +10,7 @@ import dk.cachet.carp.data.application.DataStreamBatchSerializer
 import dk.cachet.carp.data.application.DataStreamId
 import dk.cachet.carp.data.application.DataStreamService
 import dk.cachet.carp.data.application.DataStreamsConfiguration
+import kotlinx.datetime.Instant
 import kotlinx.serialization.*
 import kotlin.js.JsExport
 
@@ -54,15 +56,25 @@ sealed class DataStreamServiceRequest<out TReturn> : ApplicationServiceRequest<D
     }
 
     @Serializable
-    data class CloseDataStreams( val studyDeploymentIds: Set<UUID> ) :
-        DataStreamServiceRequest<Unit>()
+    data class GetBatchForStudyDeployments(
+        val studyDeploymentIds: Set<UUID>,
+        val deviceRoleNames: Set<String>? = null,
+        val dataTypes: Set<DataType>? = null,
+        val from: Instant? = null,
+        val to: Instant? = null
+    ) : DataStreamServiceRequest<DataStreamBatch>()
+    {
+        override fun getResponseSerializer() = DataStreamBatchSerializer
+    }
+
+    @Serializable
+    data class CloseDataStreams( val studyDeploymentIds: Set<UUID> ) : DataStreamServiceRequest<Unit>()
     {
         override fun getResponseSerializer() = serializer<Unit>()
     }
 
     @Serializable
-    data class RemoveDataStreams( val studyDeploymentIds: Set<UUID> ) :
-        DataStreamServiceRequest<Set<UUID>>()
+    data class RemoveDataStreams( val studyDeploymentIds: Set<UUID> ) : DataStreamServiceRequest<Set<UUID>>()
     {
         override fun getResponseSerializer() = serializer<Set<UUID>>()
     }
--- a/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/InMemoryDataStreamService.kt
+++ b/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/InMemoryDataStreamService.kt
@@ -2,14 +2,19 @@ package dk.cachet.carp.data.infrastructure
 
 import dk.cachet.carp.common.application.UUID
 import dk.cachet.carp.common.application.data.Data
+import dk.cachet.carp.common.application.data.DataType
 import dk.cachet.carp.common.application.intersect
 import dk.cachet.carp.common.domain.ExtractUniqueKeyMap
 import dk.cachet.carp.data.application.DataStreamBatch
 import dk.cachet.carp.data.application.DataStreamId
+import dk.cachet.carp.data.application.DataStreamSequence
 import dk.cachet.carp.data.application.DataStreamService
 import dk.cachet.carp.data.application.DataStreamsConfiguration
+import dk.cachet.carp.data.application.Measurement
 import dk.cachet.carp.data.application.MutableDataStreamBatch
 import dk.cachet.carp.data.application.MutableDataStreamSequence
+import dk.cachet.carp.data.application.applyToTimestamp
+import kotlinx.datetime.Instant
 
 
 /**
@@ -20,14 +25,18 @@ class InMemoryDataStreamService : DataStreamService
     private val configuredDataStreams: ExtractUniqueKeyMap<UUID, DataStreamsConfiguration> =
         ExtractUniqueKeyMap( { configuration -> configuration.studyDeploymentId } )
         {
-            studyDeploymentId ->
-                IllegalStateException(
-                    "Data streams for deployment with \"$studyDeploymentId\" have already been configured."
-                )
+                studyDeploymentId ->
+            IllegalStateException(
+                "Data streams for deployment with \"$studyDeploymentId\" have already been configured."
+            )
         }
     private val stoppedStudyDeploymentIds: MutableSet<UUID> = mutableSetOf()
     private val dataStreams: MutableDataStreamBatch = MutableDataStreamBatch()
 
+    companion object
+    {
+        private const val MICROSECONDS_TO_MILLISECONDS = 1000L
+    }
 
     /**
      * Start accepting data for a study deployment for data streams configured in [configuration].
@@ -52,15 +61,15 @@ class InMemoryDataStreamService : DataStreamService
     override suspend fun appendToDataStreams( studyDeploymentId: UUID, batch: DataStreamBatch )
     {
         require( batch.sequences.all { it.dataStream.studyDeploymentId == studyDeploymentId } )
-            { "The study deployment ID of one or more sequences in `batch` doesn't match `studyDeploymentId`." }
+        { "The study deployment ID of one or more sequences in `batch` doesn't match `studyDeploymentId`." }
 
         val configuration = configuredDataStreams[ studyDeploymentId ]
         requireNotNull( configuration ) { "No data streams configured for this study deployment." }
         require( batch.sequences.all { it.dataStream in configuration.expectedDataStreamIds } )
-            { "The batch contains a sequence with a data stream which wasn't configured for this study deployment." }
+        { "The batch contains a sequence with a data stream which wasn't configured for this study deployment." }
 
         check( studyDeploymentId !in stoppedStudyDeploymentIds )
-            { "Data streams for this study deployment have been closed." }
+        { "Data streams for this study deployment have been closed." }
 
         dataStreams.appendBatch( batch )
     }
@@ -86,9 +95,9 @@ class InMemoryDataStreamService : DataStreamService
         val configuration = configuredDataStreams[ dataStream.studyDeploymentId ]
         requireNotNull( configuration ) { "No data streams configured for this study deployment." }
         require( dataStream in configuration.expectedDataStreamIds )
-            { "The batch contains a sequence with a data stream which wasn't configured for this study deployment." }
+        { "The batch contains a sequence with a data stream which wasn't configured for this study deployment." }
         require( fromSequenceId >= 0 && (toSequenceIdInclusive == null || toSequenceIdInclusive >= fromSequenceId) )
-            { "The starting sequence ID is negative or the end sequence ID is smaller than the starting ID." }
+        { "The starting sequence ID is negative or the end sequence ID is smaller than the starting ID." }
 
         return dataStreams.sequences
             .filter { it.dataStream == dataStream }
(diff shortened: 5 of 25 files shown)


Whathecode:
@NGrech Can you clean up the commit history a bit? At least the fixes can be squashed. There is no point in me reviewing a commit which later gets undone. ;) Also, the generated test files should probably be added simultaneously with the commit which would otherwise fail if you don't do so. It should be a goal for each commit to compile.

Possibly, that means all of this should live in a single commit. I'll review it as such if I find some time.

NGrech:
@Whathecode I squashed the new fixes into the original one. 
Note that the reason the generated test files were not in the original commit was that there is no mention of that requirement in the [CONTRIBUTING.md](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/CONTRIBUTING.md). 
I think we should add fix this:

> You can also run detekt separately through `gradle detekt `

to `gradle detektPasses`, since that is the command run in the code analysis check when committing and (at least on windows) `gradle detekt ` will build successfully when there are issues that `gradle detektPasses` will fail on.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataPoint.kt
```diff
@@ -0,0 +1,17 @@
...
+import kotlinx.serialization.Serializable
+
+/**
+ * A single typed data point from a CARP data stream, annotated with its stream origin and timestamp.
+ */
+@Serializable
+@SerialName("dk.cachet.carp.analytics.AnalyticsDataPoint")
+data class CollectedDataPoint(
+    val streamId: DataStreamId,
+    val timestamp: Instant,
```
Whathecode: Given that a CARP `Measurement` has a `sensorStartTime` and `sensorEndTime`, what is this?

E.g., how do you store an [`InterbeatInterval`](https://github.com/cph-cachet/carp.core-kotlin/blob/develop/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/data/InterbeatInterval.kt) using this data type?

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataPoint.kt
```diff
@@ -0,0 +1,17 @@
...
+    val data: Data // Polymorphic, must be registered in serializers module
```
Whathecode: There is no real point documenting this everywhere it gets used. It's not documented in the other data types either, so let's be consistent.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataPoint.kt
```diff
@@ -0,0 +1,17 @@
...
+import kotlinx.datetime.Instant
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+/**
+ * A single typed data point from a CARP data stream, annotated with its stream origin and timestamp.
+ */
+@Serializable
+@SerialName("dk.cachet.carp.analytics.AnalyticsDataPoint")
+data class CollectedDataPoint(
```
Whathecode: Inconsistent namespace and serial name from class name. This should likely simply be removed.

And, I'm still wondering about the name of this class. It seems to be a `DataStreamPoint` but minus the provenance information. Didn't we talk about making this potentially a `SynchronizedDataPoint` (if that is what it is)? That would explain dropping the synchronization info. And, if you would still link the `sequenceId`, you'd be able to link back fully to the raw data/origin, so you can retrieve the provenance information from this data point if needed.

The documentation doesn't really clarify intent here. But, the above would make sense to me.

Whathecode (changes requested):
First review round: questioning what a `CollectedDataPoint` is. More to follow as I continue looking, but that seems pretty fundamental. :)

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataSet.kt
```diff
@@ -0,0 +1,35 @@
+package dk.cachet.carp.data.application
+
+import dk.cachet.carp.common.application.data.Data
+import kotlinx.datetime.Instant
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+/**
+ * A simple wrapper around a list of [CollectedDataPoint]s.
```
Whathecode: No wrapper is needed to achieve this. You can just use [extension functions](https://kotlinlang.org/docs/extensions.html) for `Iterable<CollectedDataPoint>`, with the added benefit that these functions will work on _any_  list of that type. And, with the added benefit you don't have to iterate over all elements and instantiate a new object at end of iteration each time before processing the first element.

You can define them in the same file as `CollectedDataPoint`.

That said, it also shows the minimal benefit of this class: is `.filterByStream( id )` really better than `.filter { it.streamId == id }`? When you filter, do you expect to _only_ filter on data stream ID, time range, or type? Making one richer `filter` abstraction allowing to filter all three, with sensible defaults, may make more sense.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataPoint.kt
NGrech: Yes you are right, this was more a note to my self that did not get removed.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataPoint.kt
NGrech: I used the `CollectedData*` prefix to show that these objects represent data already collected and exiting in a repository, and not being part of a stream from device. The idea was to reflect the lifecycle stage of the data. That said, I see the risk of confusion with existing stream-related classes. I’m open to alternative names. 

And in regards tot he ``SerialName`` I will remove that.

NGrech: With respect to the `sequenceID` I linked each ``CollectedDataPoint`` back to its origin via DataStreamId, and my assumption was that synchronization had already taken place, at this point, so sequence information wasn’t critical. Currently, synchronization happens in InMemoryDataStreamService, but perhaps it would make more sense to integrate provenance handling into the collection class itself.

NGrech: With respect to the `sequenceID` I linked each ``CollectedDataPoint`` back to its origin via DataStreamId, and my assumption was that synchronization had already taken place, at this point, so sequence information wasn’t critical. Currently, synchronization happens in InMemoryDataStreamService, but perhaps it would make more sense to integrate provenance handling into the collection class itself.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataPoint.kt
NGrech: This was a bit of an oversight on my part. Right now I treat everything as points, but from an analytics perspective, intervals (aggregates or continuous ranges) need different handling. It might make sense to introduce a separate class, e.g., CollectedDataRange, to make that distinction explicit. What do you think about that vs trying to put them together in the same structre?

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataSet.kt
NGrech: Yes, the current implementation only provides minimal benefit over extension functions.

The reality is that this class (and the point class) are sitting in an awkward middle ground, currently it is a compromise to preserve some of the structure of the stream concept, but ideal it should be more of a analytics-ready format.

My main concern/question here is should this class redone to a tidy data table structure (possible using the Kotlin DataFrames) optimized for analytics. This would mean/require:

- Flattening data objects into rows with columns (timestamp, stream_id, data_type, flattened_properties)
- Supporting time-aligned queries across different data streams
- Enabling basic analytical operations/table manipulation (grouping, filtering, aggregation)
- Handling the synching step so that the data is all synchronized (At this point, when the data is in the analytics phase, it should be synced)

The draw back and what held me from taking this step initially is that DataFrame structures can be problematic for APIs, more complexity for the serialization to JSON, the API clarity will be compromised without a clear and typed schema, and there will be an increased data overhead with a lot of repeating metadata in the rows.

I am interested on your take on this. On the one hand I want the data when retrieved to be ready for analytics, but on the other hand still want to make the transmission of data easier.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataPoint.kt
Whathecode: That would deviate from the existing data point, leading to an architectural inconsistency which would need a pretty good reason to be warranted.

I'd have to revisit why the original data point didn't type them differently, but I can imagine it has to do with the _actual_ interpretation of the type already being encoded in `DataType` in this data structure, so trying to do the same, yet incomplete, using polymorphism felt like it added little to nothing extra.

NGrech:
@Whathecode I have looked at the comments you have made and I replied to your feedback, I had some questions and comments so would be interested in seeing your views.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataPoint.kt
Whathecode: > The idea was to reflect the lifecycle stage of the data. 

Arguably, all data represented in these classes is "collected" (by the sensor). I'm not certain what "uncollected" data would be. 😉

But, perhaps you mean something akin to "ingested", referring to "data ingestion" as uploading data to a backend is often called. 

That's still problematic, however, since the non-synchronized data is as much "ingested" as this data type (recall we want to store all raw incoming data to guarantee no data loss and allow re-synchronization of data).

Still, the only thing you have stated so far that seems to make this more unique is that the timestamps have been synchronized.

Whathecode: > synchronization had already taken place, at this point, so sequence information wasn’t critical.

I wasn't arguing it was critical, but if you want to interpret the provenance of _why_ the data was collected, i.e. trigger ids, you would either need to look it up (and that would be possible through sequenceId) or duplicate it. And, for debugging purposes, it could be nice to know the _original_ timestamps before synchronization, which again sequenceId would allow looking up, rather than duplicating that data. That's why I'm saying they may be merit to store it, so that a reverse lookup is possible if it is ever needed. 

The `DataStreamId` is of course also needed. It's part of the same composite key.


> Currently, synchronization happens in InMemoryDataStreamService, but perhaps it would make more sense to integrate provenance handling into the collection class itself.

Not certain what you mean by this. But, I'd imagine raw data ingestion and synchronized data ingestion to be handled separately. At iMotions, we stored one in a simple postgres DB, and the other in a time series database optimized for querying data (TimeScale). New data points were placed on an event bus, I seem to recall. Of course, this isn't necessarily needed for CAWS, but just stating this for perspective/inspiration.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataSet.kt
Whathecode: Most of this sounds specific to the analysis subsystem, and potentially even "infrastructure", not core domain. I don't see why/how an additional intermediate data representation in the data subsystem helps with this.

However, time synchronization, as we discussed before _can_ fit in the data subsystem. We'd also need it for triggers based on incoming data from other devices, etc.

Whathecode:
I see you added this to the 2.0.0 milestone instead of 1.3. Any reason you expect this to be a breaking change, i.e., warranting a new major release?

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataSet.kt
NGrech: So after reviewing everything again, and with a clearer understanding of the roles of both `Data `and `Measurement`, I realize my initial design was flawed in trying to use the `Data `class as the core container for retrieving data.

We already agreed that the current `CollectedData`* classes are not a good fit. Building on that, my updated view is:

- Drop the idea of using `Data `as the container and instead rely on `Measurement`. 
- Rather than introducing custom `CollectedData`* classes, reuse the existing `DataStreamPoint`, which already provides the provenance and metadata we need.
- As a consequence, add am immutable implementations of DataStreamSequence and DataStreamBatch. These will serve consumer-facing use cases, while the mutable remaining for producers.

I’ll make these changes and amend the PR.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataPoint.kt
NGrech: I think the issues with this class are now addressed in the comment https://github.com/carp-dk/carp.core-kotlin/pull/517#discussion_r2330428500. I will make those changes and push them.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/CollectedDataPoint.kt
NGrech: I think the issues with this are now addressed in the comment https://github.com/carp-dk/carp.core-kotlin/pull/517#discussion_r2330428500. I will make those changes and push them.

NGrech:
@Whathecode I have updated the the code based on the last discussion.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamSequence.kt
```diff
@@ -149,6 +155,126 @@ class MutableDataStreamSequence<TData : Data>(
...
 
+/**
+ * An immutable sequence of consecutive [measurements] for a [dataStream] starting from [firstSequenceId]
+ * which all share the same [triggerIds] and [syncPoint].
+ *
+ * This implementation is designed for data retrieval scenarios where immutability provides safety
+ * and cleaner concepts.
+ */
+@JsExport
+class ImmutableDataStreamSequence<TData : Data> private constructor(
```
Whathecode: Is this really needed? 🤔 You don't use it here. How do you expect to use it?

There already is the read-only `DataStreamSequence` interface, and as part of serialization `DataStreamSequenceSnapshot` is used (which is immutable). So, unless you want to safeguard against explicit casts that should get you there already from an encapsulation perspective,.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamBatch.kt
```diff
@@ -134,6 +141,149 @@ class MutableDataStreamBatch : DataStreamBatch
 }
 
 
+/**
+ * An immutable collection of data stream [sequences] optimized for data retrieval and analytics.
+ *
+ * Unlike [MutableDataStreamBatch] which enforces non-overlapping sequences for incremental building,
+ * this implementation is designed for cross-device, cross-sensor data analysis where temporal
+ * overlap between different devices/data types is expected and valid.
```
Whathecode: The contract of `DataStreamBatch` is:

> A collection of non-overlapping, ordered, data stream [sequences].

Thus, at a glance this is a violation of the Liskov Substitution Principle. You break the base contract which is expected to be upheld.

However, `DataStreamBatch` _does_ allow for "cross-device, cross-sensor data". The documentation on the types should probably be improved, but **the non-overlap/order constraint is per data stream**.  The `appendSequence` and `appendBatch` methods clarify this better. You can see that sequences are stored per `DataStreamId`, which includes `deviceRoleName`.

Possibly there is something wrong with the implementation which prompted you to create this class?

If you were missing an `appendSequences` method, you could add that to `MutableDataStreamBatch`. Furthermore, that class currently is overly strict, in that it doesn't allow adding a preceding sequence even if doing so wouldn't cause any overlap. In case that is what you need, you could support that by updating the internal data structure and re-order elements. I simply didn't need this before.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamBatch.kt
```diff
@@ -134,6 +141,149 @@ class MutableDataStreamBatch : DataStreamBatch
...
+ *
+ * The internal structure is optimized for analytics queries:
+ * Device Role Name → Data Type → Synchronized Data Sequences
```
Whathecode: Why would `device role name -> data type -> sequences` be quicker than a composite key lookup of `device role name/data type -> sequences`?

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamBatch.kt
```diff
@@ -134,6 +141,149 @@ class MutableDataStreamBatch : DataStreamBatch
...
+
+    companion object
+    {
+        /**
+         * Create an [ImmutableDataStreamBatch] from the provided sequences.
+         *
+         * @throws IllegalArgumentException when sequences within the same DataStreamId overlap.
+         *
+         * note: Sequences across different devices/data types may overlap (this is expected for analytics).
+         * note: private constructor + factory methods provides validation and optimal internal structure.
```
Whathecode: Design rationale doesn't belong in API documentation. Furthermore, that's what factory methods are usually for, so no need to comment on this.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamBatch.kt
```diff
@@ -134,6 +141,149 @@ class MutableDataStreamBatch : DataStreamBatch
...
+                        "Sequences within data stream $dataStreamId overlap: ${previous.range} and ${current.range}"
+                    }
+                }
+            }
+
+            // Build hierarchical structure: Device → DataType → Sequences
+            val deviceDataMap = sequences
+                .groupBy { it.dataStream.deviceRoleName }
+                .mapValues { (_, deviceSequences) ->
+                    deviceSequences.groupBy { it.dataStream.dataType }
```
Whathecode: These aren't ordered, thus breaking the `DataStreamBatch` contract. Is that intentional?

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamBatch.kt
```diff
@@ -134,6 +141,149 @@ class MutableDataStreamBatch : DataStreamBatch
...
+     * @param deviceRoleName if specified, only return data types for this device role name
+     */
+    fun getDataTypes( deviceRoleName: String? = null ): Set<DataType> =
+        if (deviceRoleName != null)
+        {
+            deviceDataMap[deviceRoleName]?.keys.orEmpty()
+        } else
+        {
+            deviceDataMap.values.flatMap { it.keys }.toSet()
+        }
```
Whathecode: If these are useful, they are likely useful on the interface as well.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamBatch.kt
```diff
@@ -134,6 +141,149 @@ class MutableDataStreamBatch : DataStreamBatch
...
+
+    /**
+     * Get all sequences for a specific device role name.
+     *
+     * @param synchronize if true, convert all sequences to synchronized data points and flatten
+     */
+    fun getForDevice( deviceRoleName: String, synchronize: Boolean = false ): Any =
```
Whathecode: Synchronization is an operation. And, _how_ to synchronize is open for change. Directly linking that to the data structure is thus undesirable. Instead, have an operation perform synchronization _on_ the data structure.

Why is the return type `Any`?

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamBatch.kt
```diff
@@ -134,6 +141,149 @@ class MutableDataStreamBatch : DataStreamBatch
...
+        if (synchronize)
+        {
+            // Return synchronized data points grouped by data type
+            deviceDataMap[deviceRoleName]?.mapValues { (_, sequences) ->
+                sequences.flatMap { it }.map { it.synchronize() }.sortedBy { it.syncPoint.synchronizedOn }
```
Whathecode: Remember that `it.synchronize()` can cause data points to "jump back in time". So, the previous point may end up laying before the next one in the sequence.

I consider it part of synchronization to make sure that doesn't happen. In essence, it means "synchronization" didn't happen correctly (how should you interpret this data?). A quick pragmatic solution is to drop such data points. Another solution would be some type of compression/interpolation. A perfect solution would look at data outside of the stream to see whether correlations can be found with other data streams in order to update the sync points.

Doing the first pragmatic solution, and documenting this, is probably just fine.

Whathecode (commented):
Still an incomplete review, but I started looking at why you added `ImmutableDataStreamBatch` and `...Sequence`. The PR description is missing some clarification in regards to _why_ you are adding this. Have a look at some of the questions I asked and see whether you can clarify things.

I also have the impression that adding these changes can easily be done as a separate commit (and even PR). You don't _need_ those for your updates to `DataStreamService`, and as far as I can tell, the existing data structures would work just fine.

While looking at changes, I noticed some incorrect code style whitespaces. I added a commit which you can squash.

NGrech:
@Whathecode & @yuanchen233 I have updated the PR, I think it is in a good state now and ready for revie. 
Main thing to note from the last version is that I dropped immutable class approach to not violate LSP, and added specific tests to ensure that the returns are sequential and non overlapping.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamService.kt
```diff
@@ -2,10 +2,13 @@ package dk.cachet.carp.data.application
...
 
 import dk.cachet.carp.common.application.UUID
+import dk.cachet.carp.common.application.data.DataType
 import dk.cachet.carp.common.application.services.ApiVersion
 import dk.cachet.carp.common.application.services.ApplicationService
 import dk.cachet.carp.common.application.services.IntegrationEvent
-import kotlinx.serialization.*
+import kotlinx.datetime.Instant
+import kotlinx.serialization.Required
+import kotlinx.serialization.Serializable
```
Whathecode, line 11: I don't think it's configured, but wildcard is used for serialization imports across the codebase.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamService.kt
```diff
@@ -60,6 +63,33 @@ interface DataStreamService : ApplicationService<DataStreamService, DataStreamSe
...
+     * The response is a canonical [DataStreamBatch]: for each [DataStreamId], sequences are
+     * ordered by start time and non-overlapping (contract preserved). No derived/secondary
+     * indexing is applied in this API; analytics-specific projections are out of scope here.
+     *
+     * Time range semantics: if [from] or [to] are specified, sequences are clipped to the
+     * half-open interval [from, to) (inclusive start, exclusive end).
+     *
+     * @param studyDeploymentIds Study deployments to query. Must not be empty.
+     * @param deviceRoleNames Optional device role name filter (e.g., "phone"). If null or empty, all are included.
+     * @param dataTypes Optional data type filter. If null or empty, all are included.
```
Whathecode, line 79: > If null or empty, all are included.

That's counter intuitive. If empty, just filter out everything. A good API doesn't give two ways to do the same thing.

On why it matters: suppose a caller sets up a dynamic filter determining the set of device role names they are interested in, which ends up being empty. Now the caller will get _all_ data, instead of no data, as expected.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamService.kt
```diff
@@ -60,6 +63,33 @@ interface DataStreamService : ApplicationService<DataStreamService, DataStreamSe
         toSequenceIdInclusive: Long? = null
     ): DataStreamBatch
 
+    /**
+     * Retrieve data across multiple study deployments, optionally filtered by device role names,
+     * data types, and time range.
+     *
+     * The response is a canonical [DataStreamBatch]: for each [DataStreamId], sequences are
+     * ordered by start time and non-overlapping (contract preserved). No derived/secondary
+     * indexing is applied in this API; analytics-specific projections are out of scope here.
```
Whathecode, line 72: Drop this; all of this are implementation/design details. Not API documentation. The contract (API) of `DataStreamBatch` is documented already on `DataStreamBatch`.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamService.kt
```diff
@@ -60,6 +63,33 @@ interface DataStreamService : ApplicationService<DataStreamService, DataStreamSe
...
+     *
+     * Time range semantics: if [from] or [to] are specified, sequences are clipped to the
+     * half-open interval [from, to) (inclusive start, exclusive end).
```
Whathecode, line 75: That only works if from and to are specified. But, it looks like you can omit this and instead just document inclusive/exclusive nature in the corresponding from/to parameters. As is, this causes more confusion than it answers edge cases.

Instead, I'm more surprised about how `Instant` comes into the picture here. The data subsystem only has `Long`'s for `sensorStartTime` and `sensorEndTime`. So ... what is happening here? How do I know what to pass?

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamService.kt
```diff
@@ -60,6 +63,33 @@ interface DataStreamService : ApplicationService<DataStreamService, DataStreamSe
...
+     *
+     * @param studyDeploymentIds Study deployments to query. Must not be empty.
```
Whathecode, line 77: > Must not be empty.

Why? It seems like an overly strict contract. You can easily return nothing if you pass nothing, which would cause less additional handling for this edge case by the caller if they don't care about optimization/saving a roundtrip.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamService.kt
```diff
@@ -60,6 +63,33 @@ interface DataStreamService : ApplicationService<DataStreamService, DataStreamSe
...
+     * @param deviceRoleNames Optional device role name filter (e.g., "phone"). If null or empty, all are included.
+     * @param dataTypes Optional data type filter. If null or empty, all are included.
+     * @param from Optional absolute start time (inclusive). If null, no lower bound.
+     * @param to Optional absolute end time (exclusive). If null, no upper bound.
+     * @return A [DataStreamBatch] containing matching data sequences, preserving per-stream invariants.
+     */
+    suspend fun getBatchForStudyDeployments(
```
Whathecode, line 84: I'm not certain about the naming of this. Maybe simply `getData`? But, it will depend a bit on what actually comes out. It still seems like some synchronization is bound to happen (which would need documentation!), given the `from` and `to` `Instant` parameters, in which case `getSynchronizedData` or similar could be more appropriate.

Whathecode (changes requested):
Only reviewed the `DataStreamService.getBatchForStudyDeployments()` contract for now.

Issue #295, opened 2021-08-20
Integrate RPC calls between application services through events using eventual consistency


Whathecode:
Some application services call other application services. E.g., `RecruitmenServiceHost` calls `DeploymentService.createStudyDeployment` for each call to `deployParticipantGroup`.

Currently, this is done synchronously. We have always known this is undesirable, as now what is supposed to be a single transaction (a call to the application service) may only succeed partially (in case the call to the other application service fails).

At some point, we want to integrate these services through _eventual consistency_. Technically, the service which wants to call another service will need to emit an event, which is subsequently listened to by some integration service, which then calls the target service. Conceptually, however, this also changes the semantics of the original call, as the promise no longer is to complete the full request immediately. It may take some time, and may even fail.

I don't think resolving this currently has high priority; I don't think it has to be part of the [1.0.0 milestone](https://github.com/cph-cachet/carp.core-kotlin/milestone/1). But, it is an important issue to pick up later.

Whathecode:
In relation to the `inviteNewParticipantGroup` call specifically, creating and inviting participant groups should really be separate calls, and `inviteNewParticipantGroup` should be made obsolete. This can likely be made part of #319.

This would be a breaking change of how the current `inviteNewParticipantGroup` call works, so I marked this for milestone 2.0.

PR #518, opened 2025-10-01
Persist participant `roleAssignment` in `StagedParticipantGroup`

yuanchen233:
In preparation for [editing `StagedParticipantGroup`](#319 ), creation and invitation process for new participant groups should be separated. Persist participants' `roleAssignment` provides the ability to modify `StagedParticipantGroup` before they are invited.

Note: 

- This change requires data migration: currently, role assignments are persisted in deployment subsystem.

- For implementations with event-driven databases, the `ParticipantGroupAdded` event now takes new parameters consisting of a `Set<AssignedParticipantRoles>` and an optional `name: String?`.

### DIFF
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
@@ -133,13 +133,11 @@ class RecruitmentServiceHost(
         val recruitment = getRecruitmentOrThrow( studyId )
         val (protocol, invitations) = recruitment.createInvitations( group )
 
-        // In case the same participants have been invited before,
+        // In case the same participants with the same roles have been invited with the same group name before,
         // and that deployment is still running, return the existing group.
-        // TODO: The same participants might be invited for different role names, which we currently cannot differentiate between.
-        val toDeployParticipantIds = group.map { it.participantId }.toSet()
         val deployedStatus = recruitment.participantGroups.entries
-            .firstOrNull { (_, group) ->
-                group.participantIds == toDeployParticipantIds && group.name == name
+            .firstOrNull { (_, existingGroup) ->
+                existingGroup.roleAssignments == group && existingGroup.name == name
             }
             ?.let { deploymentService.getStudyDeploymentStatus( it.key ) }
         if ( deployedStatus != null && deployedStatus !is StudyDeploymentStatus.Stopped )
@@ -148,7 +146,7 @@ class RecruitmentServiceHost(
         }
 
         // Create participant group and mark as deployed.
-        val participantGroup = recruitment.addParticipantGroup( toDeployParticipantIds, name, uuidFactory.randomUUID() )
+        val participantGroup = recruitment.addParticipantGroup( group, name, uuidFactory.randomUUID() )
         participantGroup.markAsDeployed()
         participantRepository.updateRecruitment( recruitment )
 
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/users/ParticipantGroupStatus.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/users/ParticipantGroupStatus.kt
@@ -35,6 +35,11 @@ sealed class ParticipantGroupStatus
      */
     abstract val participants: Set<Participant>
 
+    /**
+     * The participant role assignments in this group.
+     */
+    abstract val assignedParticipantRoles: Set<AssignedParticipantRoles>
+
 
     /**
      * The [participants] have not yet been invited. The list of participants can still be modified.
@@ -43,6 +48,7 @@ sealed class ParticipantGroupStatus
     data class Staged(
         override val id: UUID,
         override val participants: Set<Participant>,
+        override val assignedParticipantRoles: Set<AssignedParticipantRoles>,
         override val name: String? = null
     ) : ParticipantGroupStatus()
 
@@ -60,6 +66,7 @@ sealed class ParticipantGroupStatus
              */
             fun fromDeploymentStatus(
                 participants: Set<Participant>,
+                roleAssignment: Set<AssignedParticipantRoles>,
                 deploymentStatus: StudyDeploymentStatus,
                 name: String?
             ): InDeployment
@@ -73,14 +80,29 @@ sealed class ParticipantGroupStatus
                     is StudyDeploymentStatus.Invited,
                     is StudyDeploymentStatus.DeployingDevices ->
                         // If deployment was ready at one point (`startedOn`), consider the study 'Running'.
-                        if ( startedOn == null ) Invited( id, participants, createdOn, deploymentStatus, name )
-                        else Running( id, participants, createdOn, deploymentStatus, startedOn, name )
+                        if ( startedOn == null )
+                        {
+                            Invited( id, participants, roleAssignment, createdOn, deploymentStatus, name )
+                        }
+                        else
+                        {
+                            Running( id, participants, roleAssignment, createdOn, deploymentStatus, startedOn, name )
+                        }
                     is StudyDeploymentStatus.Running ->
-                        Running( id, participants, createdOn, deploymentStatus, checkNotNull( startedOn ), name )
+                        Running(
+                            id,
+                            participants,
+                            roleAssignment,
+                            createdOn,
+                            deploymentStatus,
+                            checkNotNull( startedOn ),
+                            name
+                        )
                     is StudyDeploymentStatus.Stopped ->
                         Stopped(
                             id,
                             participants,
+                            roleAssignment,
                             createdOn,
                             deploymentStatus,
                             startedOn,
@@ -111,6 +133,7 @@ sealed class ParticipantGroupStatus
     data class Invited(
         override val id: UUID,
         override val participants: Set<Participant>,
+        override val assignedParticipantRoles: Set<AssignedParticipantRoles>,
         override val invitedOn: Instant,
         override val studyDeploymentStatus: StudyDeploymentStatus,
         override val name: String? = null
@@ -124,6 +147,7 @@ sealed class ParticipantGroupStatus
     data class Running(
         override val id: UUID,
         override val participants: Set<Participant>,
+        override val assignedParticipantRoles: Set<AssignedParticipantRoles>,
         override val invitedOn: Instant,
         override val studyDeploymentStatus: StudyDeploymentStatus,
         /**
@@ -141,6 +165,7 @@ sealed class ParticipantGroupStatus
     data class Stopped(
         override val id: UUID,
         override val participants: Set<Participant>,
+        override val assignedParticipantRoles: Set<AssignedParticipantRoles>,
         override val invitedOn: Instant,
         override val studyDeploymentStatus: StudyDeploymentStatus,
         /**
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
@@ -3,6 +3,7 @@ package dk.cachet.carp.studies.domain.users
 import dk.cachet.carp.common.application.EmailAddress
 import dk.cachet.carp.common.application.UUID
 import dk.cachet.carp.common.application.users.AccountIdentity
+import dk.cachet.carp.common.application.users.AssignedTo
 import dk.cachet.carp.common.application.users.EmailAccountIdentity
 import dk.cachet.carp.common.application.users.UsernameAccountIdentity
 import dk.cachet.carp.common.domain.AggregateRoot
@@ -29,7 +30,10 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
     sealed class Event : DomainEvent
     {
         data class ParticipantAdded( val participant: Participant ) : Event()
-        data class ParticipantGroupAdded( val participantIds: Set<UUID> ) : Event()
+        data class ParticipantGroupAdded(
+            val participants: Set<AssignedParticipantRoles>,
+            val name: String? = null
+        ) : Event()
     }
 
 
@@ -174,27 +178,29 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
     private val _participantGroups: MutableMap<UUID, StagedParticipantGroup> = mutableMapOf()
 
     /**
-     * Create and add the participants identified by [participantIds] as a participant group, optionally with a [name]
-     * representing this group.
+     * Create and add the [participants] with assigned roles to a participant group, and optionally give it a [name].
      *
-     * @throws IllegalArgumentException when one or more of the participants aren't in this recruitment.
+     * @throws IllegalArgumentException when:
+     *  - one or more of the participants aren't in this recruitment.
+     *  - any of the participant roles specified in [participants] are not part of the configured study protocol.
      * @throws IllegalStateException when the study is not yet ready for deployment.
      */
     fun addParticipantGroup(
-        participantIds: Set<UUID>,
+        participants: Set<AssignedParticipantRoles>,
         name: String? = null,
         id: UUID = UUID.randomUUID()
     ): StagedParticipantGroup
     {
-        require( participants.map { it.id }.containsAll( participantIds ) )
-            { "One of the participants for which to create a participant group isn't part of this recruitment." }
-        check( getStatus() is RecruitmentStatus.ReadyForDeployment ) { "The study is not yet ready for deployment." }
+        val status = getStatus()
+        check( status is RecruitmentStatus.ReadyForDeployment ) { "The study is not yet ready for deployment." }
+
+        validateRoleAssignments( status.studyProtocol, participants )
 
         val group = StagedParticipantGroup( id, name )
-        group.addParticipants( participantIds )
+        group.addParticipants( participants )
 
         _participantGroups[ group.id ] = group
-        event( Event.ParticipantGroupAdded( participantIds ) )
+        event( Event.ParticipantGroupAdded( participants, null ) )
 
         return group
     }
@@ -213,6 +219,7 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
         val participants = group.participantIds.map { id -> _participants.first { it.id == id } }
         return ParticipantGroupStatus.InDeployment.fromDeploymentStatus(
             participants.toSet(),
+            group.roleAssignments,
             studyDeploymentStatus,
             group.name
         )
@@ -223,4 +230,29 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
      */
     override fun getSnapshot( version: Int ): RecruitmentSnapshot =
         RecruitmentSnapshot.fromParticipantRecruitment( this, version )
+
+    /**
+     * Validate that all participants exist in this recruitment and that all assigned roles are part of the protocol.
+     *
+     * @throws IllegalArgumentException when:
+     *  - one or more of the participants aren't in this recruitment.
+     *  - any of the participant roles specified in [participants] are not part of the configured study protocol.
+     */
+    private fun validateRoleAssignments( protocol: StudyProtocolSnapshot, participants: Set<AssignedParticipantRoles> )
+    {
+        require( this.participants.map { it.id }.containsAll( participants.participantIds() ) )
+            { "One of the participants for which to create a participant group isn't part of this recruitment." }
+
+        val assignedParticipantRoles = participants
+            .map { it.assignedRoles }
+            .filterIsInstance<AssignedTo.Roles>()
+            .flatMap { it.roleNames }
+            .toSet()
+        val availableRoles = protocol.participantRoles.map { it.role }.toSet()
+
+        assignedParticipantRoles.forEach { assigned ->
+            require( assigned in availableRoles )
+                { "The assigned participant role \"$assigned\" is not part of the study protocol." }
+        }
+    }
 }
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/StagedParticipantGroup.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/StagedParticipantGroup.kt
@@ -1,6 +1,8 @@
 package dk.cachet.carp.studies.domain.users
 
 import dk.cachet.carp.common.application.UUID
+import dk.cachet.carp.studies.application.users.AssignedParticipantRoles
+import dk.cachet.carp.studies.application.users.participantIds
 import kotlinx.serialization.*
 
 
@@ -17,12 +19,17 @@ data class StagedParticipantGroup(
     /**
      * An optional name to represent the group of participants.
      */
-    val name: String? = null,
+    var name: String? = null,
 )
 {
-    private val _participantIds: MutableSet<UUID> = mutableSetOf()
+    private val _roleAssignments: MutableSet<AssignedParticipantRoles> = mutableSetOf()
+    /**
+     * The roles assigned to participants in this group.
+     */
+    val roleAssignments: Set<AssignedParticipantRoles>
+        get() = _roleAssignments
     val participantIds: Set<UUID>
-        get() = _participantIds
+        get() = _roleAssignments.participantIds()
 
     /**
      * Determines whether this participant group has been deployed.
@@ -30,19 +37,17 @@ data class StagedParticipantGroup(
     var isDeployed: Boolean = false
         private set
 
-
-
     /**
-     * Add participants with [participantIds] to this group.
+     * Add [participants] with assigned roles to this group.
      * This is only allowed when the group hasn't been deployed yet.
      *
      * @throws IllegalStateException when this participant group is already deployed.
      */
-    fun addParticipants( participantIds: Set<UUID> )
+    fun addParticipants( participants: Set<AssignedParticipantRoles> )
     {
-        check( !isDeployed ) { "Can't add participant after a participant group has been deployed." }
+        check( !isDeployed ) { "Can't add participants after a participant group has been deployed." }
 
-        _participantIds.addAll( participantIds )
+        _roleAssignments.addAll( participants )
     }
 
     /**
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/versioning/RecruitmentServiceApiMigrator.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/versioning/RecruitmentServiceApiMigrator.kt
@@ -70,8 +70,9 @@ private val major1Minor2To3Migration =
                 }
             }
 
-            // Remove newly added 'name' field from 'ParticipantGroupStatus'.
+            // Remove newly added fields from `ParticipantGroupStatus`.
             json.remove( "name" )
+            json.remove( "assignedParticipantRoles" )
         }
 
         override fun migrateEvent( event: JsonObject ) = event
--- a/carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceTest.kt
+++ b/carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceTest.kt
@@ -185,6 +185,8 @@ interface RecruitmentServiceTest
             setOf( assignedParticipant ),
             "Group 1"
         )
+
+        // Deploy the same participants in a group with a different name.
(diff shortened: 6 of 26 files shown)


carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/StagedParticipantGroup.kt
```diff
@@ -30,19 +32,28 @@ data class StagedParticipantGroup(
...
 
-
+    /**
+     * The roles assigned to participants in this group.
+     *
+     * Present only before deployment; must be null after
+     */
+    private var _roleAssignments: Set<AssignedParticipantRoles>? = mutableSetOf()
+    val roleAssignments: Set<AssignedParticipantRoles>?
+        get() = _roleAssignments
```
Whathecode: This change is made in the commit:
> Refactor: change participant group creation to use role assignments

This isn't refactoring: it adds new functionality and modifies persisted data. The change to persistence also means that guidance needs to be provided on how to upgrade prior data as part of the next release. The (final) assigned roles are currently only persisted in the deployments subsystem.

Of course, if the contract is that role assignments are null after deployment, data migration can just set this to null. However, it's not immediately clear to me why that would make sense as a value, more than, e.g., the _last_ assigned value, just like `name` is set to the last assigned value. This is inconsistent, and thus confusing.

The commit message is missing rationale (the "why"). But, after looking at the changes, I believe I can infer something along the lines of:
> `StagedParticipantGroup` currently doesn't persist assigned participant roles. There was no need to do so before, since these were only used in the invitations sent out to participants, and creating and inviting a participant group is done in a single operation. However, now there is a need to be able to modify staged participant groups before inviting them, which means that this state needs to be persisted.

Whathecode (commented):
Slowly starting to look at this. I've been busy lately, so I'll just post comments as I go along.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -148,7 +148,7 @@ class RecruitmentServiceHost(
         }
 
         // Create participant group and mark as deployed.
-        val participantGroup = recruitment.addParticipantGroup( toDeployParticipantIds, name, uuidFactory.randomUUID() )
+        val participantGroup = recruitment.addParticipantGroup( group, name, uuidFactory.randomUUID() )
```
Whathecode, line 149: Since you now store assignments in recruitment, you can in fact resolve the `TODO` above:

> // TODO: The same participants might be invited for different role names, which we currently cannot differentiate between.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/users/ParticipantGroupStatus.kt
```diff
@@ -43,6 +43,7 @@ sealed class ParticipantGroupStatus
     data class Staged(
         override val id: UUID,
         override val participants: Set<Participant>,
+        val assignedParticipantRoles: Set<AssignedParticipantRoles>,
```
Whathecode: I contemplated whether adding something like `AssignedParticipant` wouldn't be easier to work with, rather than storing it in normalized form like this. But, `Participant` [is slated to be expanded with participant metadata at some point](https://github.com/carp-dk/carp.core-kotlin/issues/273). So then you'd have to update both `AssignedParticipant` and `Participant` if you want to expose this data in both.

Keeping them separate like this may be fine, and may even prove to be preferable over time. :)

However, why is this only defined on `Staged`? I'd expect this to be defined on the abstract `ParticipantGroupStatus` (thus for all statuses). It simply can't be modified after deployment.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -174,24 +174,25 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
...
-     * Create and add the participants identified by [participantIds] as a participant group, optionally with a [name]
+     * Create and add the participants identified by [roleAssignments] as a participant group, optionally with a [name]
      * representing this group.
      *
      * @throws IllegalArgumentException when one or more of the participants aren't in this recruitment.
      * @throws IllegalStateException when the study is not yet ready for deployment.
      */
     fun addParticipantGroup(
-        participantIds: Set<UUID>,
+        roleAssignments: Set<AssignedParticipantRoles>,
```
Whathecode: Isn't `participants` a more meaningful name here? Then you can also update the documentation:
> Create and add the [participants] with assigned roles to a participant group, and optionally give it a [name].

In addition, this new parameter requires you to check for constraints (whether the participant roles are valid), just like `createInvitations` does. You can probably lift it out into a private method. Don't forget to update the documentation accordingly.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -174,24 +174,25 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
...
     ): StagedParticipantGroup
     {
+        val participantIds = roleAssignments.participantIds()
         require( participants.map { it.id }.containsAll( participantIds ) )
             { "One of the participants for which to create a participant group isn't part of this recruitment." }
         check( getStatus() is RecruitmentStatus.ReadyForDeployment ) { "The study is not yet ready for deployment." }
 
-        val group = StagedParticipantGroup( id, name )
-        group.addParticipants( participantIds )
+        val group = StagedParticipantGroup(id, name)
```
Whathecode: Spacing.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -174,24 +174,25 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
...
+        group.setRoleAssignments( roleAssignments )
```
Whathecode: I think I prefer the previous `.addParticipants` name. There is nothing inherently wrong with calling participant ID with assigned roles "participants" imho.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/StagedParticipantGroup.kt
```diff
@@ -17,7 +19,7 @@ data class StagedParticipantGroup(
     /**
      * An optional name to represent the group of participants.
      */
-    val name: String? = null,
+    var name: String? = null,
```
Whathecode, line 22: At first I thought this was wrong. But, unlike changing participants or their assigned roles, I guess there is nothing wrong with modifying the group name after the participants have been invited. 👍

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/StagedParticipantGroup.kt
Whathecode: Wouldn't it be better if this simply _replaces_ `participantIds`? Why have it as a separate field?

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentTest.kt
```diff
@@ -106,13 +109,14 @@ class RecruitmentTest
...
-        val group = recruitment.addParticipantGroup( participantIds, groupName )
+        val roleAssignment = setOf( AssignedParticipantRoles( participant.id, AssignedTo.All ) )
+        val group = recruitment.addParticipantGroup( roleAssignment, groupName )
         assertEquals( Recruitment.Event.ParticipantGroupAdded( participantIds ), recruitment.consumeEvents().last() )
         assertEquals(
             participant.id,
             recruitment.participantGroups[ group.id ]?.participantIds?.singleOrNull()
         )
-        assertEquals( groupName, group.name )
+        assertEquals( groupName, group.name)
```
Whathecode: Spacing.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/StagedParticipantGroupTest.kt
```diff
@@ -17,36 +16,53 @@ class StagedParticipantGroupTest
     {
         val group = StagedParticipantGroup()
         val participantId = UUID.randomUUID()
-        group.addParticipants( setOf( participantId ) )
+        val roleAssignment = AssignedParticipantRoles( participantId, AssignedTo.All )
+        group.setRoleAssignments( setOf( roleAssignment ) )
 
         assertEquals( participantId, group.participantIds.singleOrNull() )
+        assertEquals( roleAssignment, group.roleAssignments!!.singleOrNull())
```
Whathecode: Spacing.

Whathecode (changes requested):
Full review of the first commit is done. Let me know if this feedback makes sense.

And, already quickly looking at the second commit: I think we can mark the `inviteNewParticipantGroup` obsolete as part of that work.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/users/ParticipantGroupStatus.kt
yuanchen233: > However, why is this only defined on Staged? I'd expect this to be defined on the abstract ParticipantGroupStatus (thus for all statuses). It simply can't be modified after deployment.

The question I had along the way is : Should participant role assignments persiste in `Recruitment`, currently it's in `ParticipantInvitation` which is under `Deployment`.

I was following the idea of `StagedParticipantGroup` only holds those data temporarely, then they should be removed once the group is invited. This is why role assignment is set to `null` after deployments. This make `assignedParticipantRoles` only relavent in `Staged`, since it will be `null` in all other states.

Whathecode: But what do you gain by doing so? That data is still relevant after deployment, just not mutable. Otherwise you need to query the deployment subsystem to get that information, even if you don't need any deployment data.

Whathecode: Oh, I see we internally query the deployment subsystem to see whether participant group status is retrieved either way, which also includes assignment. 

So I can see how that would be a bit inconsistent. However, it's worthwhile considering whether down the line that information shouldn't be sent via the event bus instead of a chained sync RPC call. 🤔

yuanchen233: > But what do you gain by doing so? That data is still relevant after deployment, just not mutable.

Yes, I think it's better to keep the role assignment data even after deployment. This is related to your other comments as well, I'll go ahead and make the change.

For consistency considerations, would you think it should be a separate issue? or it should be considered along with #295, instead of solving it here?

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/StagedParticipantGroup.kt
yuanchen233: I was trapped in the idea of set `roleAssignment` to null after deployment, therefore keep it separated from `participantIds` was needed. This will be changed.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentTest.kt
yuanchen233: When I review my own code here I look for those missing spaces (or at least I try to), but they still shows up every time :( 

Maybe I should set detekt to not exclude '*/test/*' and let it check for me.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/StagedParticipantGroup.kt
yuanchen233: Yep, the only thing bothered me was: we now check for a unique combination of participants + name when create/invite new participant groups. Allowing group name to be modified after the group has been invited, means same participants could be invited to more than one group.  There is nothing wrong with it by design, just felt like removing a constrain without mentioning it. 

Anyway, I do believe it should be up to implementation/infrastructure layer to limit if same participant can be invited to more than one group.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
yuanchen233: > Isn't participants a more meaningful name here? 

Yes, however, `Recruitment` has `val participants: Set<Participant>`, and within this method it is used to check if all `participant` was added to this recruitment before. I didn't want to reuse variable name here.

---
Nvm, it is reused in other places as well.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/users/ParticipantGroupStatus.kt
Whathecode: P.s. for context, there is this issue on the problem with [synchronous RPC calls between our application services](https://github.com/carp-dk/carp.core-kotlin/issues/295). It still shouldn't be high priority, _but_, it does validate the design decision here to keep role assignments stored. That's more in line with where we want to be headed.

UPDATE: I see that's the one you link to as well! 👍 But, as I just said here, I don't think that should have priority while developing this feature right now.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/StagedParticipantGroup.kt
Whathecode: > Allowing group name to be modified after the group has been invited, means same participants could be invited to more than one group. There is nothing wrong with it by design, just felt like removing a constrain without mentioning it.

Well-spotted! This seems like a potentially desirable feature to me, and not a bug. 🤷‍♂️ I think the previous constraint was probably accidental complexity.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentTest.kt
Whathecode: If the checker doesn't find these, we should probably fix the configuration and/or checker, yes. :) But as usual, probably not a priority.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/StagedParticipantGroup.kt
yuanchen233: I removed private `_participantIds` but kept the accessor as: 
`val participantIds: Set<UUID>`
        `get() = _roleAssignments.participantIds()`

because I saw there are several places only using `participantIds`, I thought calling `roleAssignments.participantIds()` in different places is redundant and misleading.

Would this be the desired approach?

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -133,13 +133,11 @@ class RecruitmentServiceHost(
...
-        // In case the same participants have been invited before,
+        // In case the same participants with the same roles have been invited with the same group name before,
         // and that deployment is still running, return the existing group.
-        // TODO: The same participants might be invited for different role names, which we currently cannot differentiate between.
-        val toDeployParticipantIds = group.map { it.participantId }.toSet()
         val deployedStatus = recruitment.participantGroups.entries
-            .firstOrNull { (_, group) ->
-                group.participantIds == toDeployParticipantIds && group.name == name
+            .firstOrNull { (_, existGroup) ->
+                existGroup.roleAssignments == group && existGroup.name == name
```
Whathecode: Minor: `existingGroup` is more grammatical.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/StagedParticipantGroup.kt
```diff
@@ -17,32 +19,35 @@ data class StagedParticipantGroup(
...
+     * Add [participants] with assigned roles to this group.
      * This is only allowed when the group hasn't been deployed yet.
      *
      * @throws IllegalStateException when this participant group is already deployed.
      */
-    fun addParticipants( participantIds: Set<UUID> )
+    fun addParticipants( participants: Set<AssignedParticipantRoles> )
     {
-        check( !isDeployed ) { "Can't add participant after a participant group has been deployed." }
+        check( !isDeployed ) { "Can't set role assignments after a participant group has been deployed." }
```
Whathecode: "add participants" is more logical here, given that we didn't end up changing the method name.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -223,4 +228,29 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
...
+     * Validate that all participants exist in this recruitment and that all assigned roles are part of the protocol.
+     *
+     * @throws IllegalArgumentException when:
+     *  - one or more of the participants aren't in this recruitment.
+     *  - any of the participant roles specified in [participants] are not part of the configured study protocol.
+     */
+    private fun validateRoleAssignments( protocol: StudyProtocolSnapshot, participants: Set<AssignedParticipantRoles> )
+    {
+        require( this.participants.map { it.id }.containsAll( participants.participantIds() ) )
+        { "One of the participants for which to create a participant group isn't part of this recruitment." }
```
Whathecode: Missing indentation.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -223,4 +228,29 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
...
+        val assignedParticipantRoles = participants
+            .map { it.assignedRoles }
+            .filterIsInstance<AssignedTo.Roles>()
+            .flatMap { it.roleNames }
+            .toSet()
+        val availableRoles = protocol.participantRoles.map { it.role }.toSet()
+
+        assignedParticipantRoles.forEach { assigned ->
+            require( assigned in availableRoles )
+            { "The assigned participant role \"$assigned\" is not part of the study protocol." }
```
Whathecode: Missing indentation.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -174,27 +176,29 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
...
+
+        validateRoleAssignments( status.studyProtocol, participants )
 
         val group = StagedParticipantGroup( id, name )
-        group.addParticipants( participantIds )
+        group.addParticipants( participants )
 
         _participantGroups[ group.id ] = group
-        event( Event.ParticipantGroupAdded( participantIds ) )
+        event( Event.ParticipantGroupAdded( participants.participantIds() ) )
```
Whathecode: The domain event needs to be updated as well, since [otherwise not all changes to the aggregate root are reflected as events](https://github.com/carp-dk/carp.core-kotlin/blob/a22826d48d32686675afe669ebfdf63fefae53d4/docs/development-checklists.md).

> Domain events keep track of changes to the aggregate root. Currently, they are stored as a simple list of events. These can be handled by calling consumeEvents, after which they are erased. Note: depending on the chosen persistence model in an infrastructure, events may go unused, e.g., in case a document-oriented database is used and any update is written by updating the full document.

This is to enable event sourcing implementations, which is a bit theoretical since we never used that either at iMotions or DTU (both use a document store so events go unused as mentioned in the note), but, no reason to deviate from the architecture at this point. :) In the version migrator you will have to downgrade the event.

I actually almost forgot about this requirement. I wonder whether we overlooked it in some of the previous PRs. 🤔 Could you have a look and file bugs if we missed it in previous spots (no need to address right now unless it's quick; let's prevent scope creep)?

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/StagedParticipantGroup.kt
```diff
@@ -17,32 +19,35 @@ data class StagedParticipantGroup(
...
      * An optional name to represent the group of participants.
      */
-    val name: String? = null,
+    var name: String? = null,
 )
 {
-    private val _participantIds: MutableSet<UUID> = mutableSetOf()
+    /**
+     * The roles assigned to participants in this group.
+     */
```
Whathecode, line 28: You documented the private field instead of the public one. ;) I see you did the same in some later commits as well.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/versioning/RecruitmentServiceApiMigrator.kt
```diff
@@ -70,6 +70,9 @@ private val major1Minor2To3Migration =
                 }
             }
 
+            // Remove newly added 'roleAssignments' field from 'ParticipantGroupStatus'.
```
Whathecode: Comment doesn't match code. It's also pretty self-documenting so maybe can be removed altogether and grouped with "name" removal under a comment similar to:

> Remove newly added fields from `ParticipantGroupStatus`.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceTest.kt
```diff
@@ -185,6 +185,7 @@ interface RecruitmentServiceTest
             setOf( assignedParticipant ),
             "Group 1"
         )
+        // Deploy the same participants in a group with a different name.
```
Whathecode, line 189: Newline.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentTest.kt
```diff
@@ -123,20 +128,34 @@ class RecruitmentTest
...
     }
 
+    @Test
+    fun addParticipantGroup_fails_when_role_not_configured_in_protocol()
+    {
+        val recruitment = Recruitment( studyId )
+        val participant = recruitment.addParticipant( participantEmail )
+        val protocol = createEmptyProtocol()
+        recruitment.lockInStudy( protocol.getSnapshot(), StudyInvitation( "Some study" ) )
+        val role1 = ParticipantRole( "Role 1", isOptional = false )
```
Whathecode: It would help readability to call this `unknownRole`, or `nonExistingRole`. But, this field goes unused, so instead the very similar naming here can be cleaned up. The next one can just be `unknownRole`, with "Unknown role" inlined, and the one after that `unknownRoleAssignment`, or similar.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/StagedParticipantGroupTest.kt
```diff
@@ -17,34 +16,40 @@ class StagedParticipantGroupTest
     {
         val group = StagedParticipantGroup()
         val participantId = UUID.randomUUID()
-        group.addParticipants( setOf( participantId ) )
+        val roleAssignment = AssignedParticipantRoles( participantId, AssignedTo.All )
```
Whathecode: You could probably easily use a named parameter and inline `participantId` in several modified tests here, making things a bit more concise. :) (If it fits on one line.)

typescript-declarations/tests/carp-studies-test.ts
```diff
@@ -122,9 +122,11 @@ describe( "carp-studies-core", () => {
             const emptyDeploymentStatusList = KtList.fromJsArray( [] )
             const emptyParticipantStatusList = KtList.fromJsArray( [] )
             const deploymentStatus = new StudyDeploymentStatus.Running( now, deploymentId, emptyDeploymentStatusList, emptyParticipantStatusList, now )
-            const participantsSet = new Set( [ new Participant( new UsernameAccountIdentity( new Username( "Test" ) ) ) ] )
-            const participants = KtSet.fromJsSet( participantsSet )
-            const group = new ParticipantGroupStatus.Invited( deploymentId, participants, now, deploymentStatus, "Test group" )
+            const participantsSet = [ new Participant( new UsernameAccountIdentity( new Username( "Test" ) ) ) ]
+            const participants = KtSet.fromJsSet( new Set(participantsSet ) )
```
Whathecode: Spacing.

rpc/schemas/studies/users/ParticipantGroupStatus.json
```diff
@@ -15,10 +15,16 @@
         "name": { "type": [ "string", "null" ] },
         "participants": {
           "type": "array",
-          "items": { "$ref": "../users/Participant.json" }
+          "items": {
+            "$ref": "../users/Participant.json"
+          }
```
Whathecode: Unnecessary change, and all other schemas place it on single line when possible it seems.

Whathecode (changes requested):
Last comments on the first commit. Great work! A lot of minor stuff, and only one "big" comment about events to enable event sourcing (something I totally overlooked before myself).

I suggest we move the other two commits to another PR so that we can get this merged and we have less comments to scroll through for subsequent part of review. :)

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentTest.kt
yuanchen233: I also renamed the test case name `addParticipantGroup_fails_for_unknown_participant_roles` to match other similar tests.

rpc/schemas/studies/users/ParticipantGroupStatus.json
yuanchen233: When only focusing on the changes from amending a commit, but not reviewing the final commit, lessons learned.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
yuanchen233: This is an `AggregateRoot.Event`, I think we only testing against the types using `consumeEvents`, not the content, they are not serializable either. Our ApiMigrator are implemented on `JsonObject`, the `migrateEvent` works for `ApplicationService.Event`, where we can test using test-request json -> published events. Am I missing something here?

> I wonder whether we overlooked it in some of the previous PRs

I think the `StagedParticipantGroup.name` we added last commit should also be part of this event.

Whathecode: Nope; my bad. We don't have proper coverage here, and if it doesn't cross API boundaries, we consider it part of upgrading the persisted data which is out of scope for core. 🤷‍♂️

yuanchen233: Well, I think we should still upgrade this event to stay aligned with our design and documentation, but include a migration guide for the infrastructure and implementations to handle the transition.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -29,7 +30,10 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
     sealed class Event : DomainEvent
     {
         data class ParticipantAdded( val participant: Participant ) : Event()
-        data class ParticipantGroupAdded( val participantIds: Set<UUID> ) : Event()
+        data class ParticipantGroupAdded(
+            val participantIds: Set<AssignedParticipantRoles>,
```
Whathecode: Wrong member name.

Whathecode (changes requested):
Other than one erroneous name, we're good to go! Don't forget to update the PR title/description.

yuanchen233:
Great! I also updated the commit message to include the need of event migration we discussed.

Whathecode (approved):

PR #519, opened 2025-10-22
Add separate endpoint for create and invite participant group methods…

### LINKED ISSUE #319: Endpoints to edit staged participant groups

yuanchen233:
… to RecruitmentService

Also marks `InviteNewParticipantGroup` Endpoint and related method as deprecated, and replace their usage with new create and invite method respectively.

----------
- I did not include `connectedDevicePreregistration` here, as I think it should be a separate issue/commit.
- I did not include 'add `UpdateParticipantGroup` endpoint' in this PR, to keep thing clean here.

Qs:
- Should test cases related to `inviteNewParticipant` be removed? as its functionalities are now in create and invite, respectively and we have coverage for them. So some suppress can be removed. ( Kotlin compiler fails it even the deprecation level is warning )

---
TODO(not for this PR):
1. Update 'stopParticipantGroup' parameter in the order from the most relevant one to secondary ones.
2. Create an issue to keep all creation endpoint consistent: let user pass IDs.
3. Create new endpoint `UpdateParticipantGroup`
4. Rename `StagedParticipantGroup `

### DIFF
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
@@ -74,11 +74,49 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
      *  - not all necessary participant roles part of the study have been assigned a participant
      * @throws IllegalStateException when the study is not yet ready for deployment.
      */
+    @Deprecated(
+        "Use createParticipantGroup and inviteParticipantGroup instead",
+        ReplaceWith(
+            "inviteParticipantGroup( createParticipantGroup( UUID.randomUUID(), group, studyId, name ).id )",
+            "dk.cachet.carp.common.application.UUID"
+        ),
+        level = DeprecationLevel.WARNING
+    )
     suspend fun inviteNewParticipantGroup(
         studyId: UUID,
         group: Set<AssignedParticipantRoles>,
         name: String? = null
     ): ParticipantGroupStatus
+
+    /**
+     * Create a new participant [group] of previously added participants for the study with the given [studyId],
+     * and an optional [name] representing this group, but do not yet send out invitations.
+     * This is used to create a group of participants which can be deployed at a later time.
+     *
+     * As long as no final study protocol is locked in for the study, a participant group can't be created
+     * since participant roles to which participants need to be assigned are unknown.
+     * @throws IllegalArgumentException when:
+     *  - a study with [studyId] does not exist
+     *  - an existing participant group with [groupId] already exists
+     *  - any of the participant roles specified in [group] does not exist
+     * @throws IllegalStateException when the study is not yet ready for deployment.
+     */
+    suspend fun createParticipantGroup(
+        groupId: UUID,
+        group: Set<AssignedParticipantRoles>,
+        studyId: UUID,
+        name: String? = null
+    ): ParticipantGroupStatus
+
+    /**
+     * Invite the participant group with the specified [groupId] to start participating in its study.
+     *
+     * @throws IllegalArgumentException when:
+     *  - the participant group with [groupId] does not exist
+     *  - not all necessary participant roles part of the study have been assigned a participant
+     * @throws IllegalStateException when group has already been deployed.
+     */
+    suspend fun inviteParticipantGroup( groupId: UUID ): ParticipantGroupStatus
 
     /**
      * Get the status of all participant groups in the study with the specified [studyId].
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
@@ -81,7 +81,7 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
     ): ParticipantGroupStatus
 
     /**
-     * Get the status of all deployed participant groups in the study with the specified [studyId].
+     * Get the status of all participant groups in the study with the specified [studyId].
      *
      * @throws IllegalArgumentException when a study with [studyId] does not exist.
      */
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
@@ -125,6 +125,13 @@ class RecruitmentServiceHost(
      *  - not all necessary participant roles part of the study have been assigned a participant
      * @throws IllegalStateException when the study is not yet ready for deployment.
      */
+    @Deprecated(
+        "Use createParticipantGroup and inviteParticipantGroup instead",
+        replaceWith = ReplaceWith(
+            "inviteParticipantGroup( createParticipantGroup( UUID.randomUUID(), group, studyId, name ).id )",
+            "dk.cachet.carp.common.application.UUID"
+        )
+    )
     override suspend fun inviteNewParticipantGroup(
         studyId: UUID,
         group: Set<AssignedParticipantRoles>,
@@ -156,6 +163,64 @@ class RecruitmentServiceHost(
         //  This should use eventual consistency: https://github.com/imotions/carp.core-kotlin/issues/295
         //  And probably be separate requests: https://github.com/imotions/carp.core-kotlin/issues/319
         val deploymentStatus = deploymentService.createStudyDeployment( participantGroup.id, protocol, invitations )
+
+        return recruitment.getParticipantGroupStatus( participantGroup.id ) { deploymentStatus }
+    }
+
+    /**
+     * Create a new participant [group] of previously added participants for the study with the given [studyId],
+     * and an optional [name] representing this group, but do not yet send out invitations.
+     * This is used to create a group of participants which can be deployed at a later time.
+     *
+     * As long as no final study protocol is locked in for the study, a participant group can't be created
+     * since participant roles to which participants need to be assigned are unknown.
+     * @throws IllegalArgumentException when:
+     *  - a study with [studyId] does not exist
+     *  - an existing participant group with [groupId] already exists
+     *  - any of the participant roles specified in [group] does not exist
+     * @throws IllegalStateException when the study is not yet ready for deployment.
+     */
+    override suspend fun createParticipantGroup(
+        groupId: UUID,
+        group: Set<AssignedParticipantRoles>,
+        studyId: UUID,
+        name: String?
+    ): ParticipantGroupStatus
+    {
+        val recruitment = getRecruitmentOrThrow( studyId )
+        require( participantRepository.getRecruitmentWithParticipantGroup( groupId ) == null )
+            { "Participant group with ID \"$groupId\" already exists." }
+
+        val participantGroup = recruitment.addParticipantGroup( group, name, groupId )
+        participantRepository.updateRecruitment( recruitment )
+
+        return recruitment.getParticipantGroupStatus( participantGroup.id ) {
+            error( "Participant group with ID \"$groupId\" is not yet deployed." )
+        }
+    }
+
+    /**
+     * Invite the participant group with the specified [groupId] to start participating in its study.
+     *
+     * @throws IllegalArgumentException when:
+     *  - the participant group with [groupId] does not exist
+     *  - not all necessary participant roles part of the study have been assigned a participant
+     * @throws IllegalStateException when group has already been deployed.
+    */
+    override suspend fun inviteParticipantGroup( groupId: UUID ): ParticipantGroupStatus
+    {
+        val (recruitment, participantGroup) = getRecruitmentWithGroupOrThrow( groupId )
+
+        check( !participantGroup.isDeployed ) { "Participant group has already been deployed." }
+
+        val (protocol, invitations) = recruitment.createInvitations( participantGroup.roleAssignments )
+        val deploymentStatus = deploymentService.createStudyDeployment( participantGroup.id, protocol, invitations )
+
+        // Mark participant group as deployed.
+        // TODO: If `createStudyDeployment` succeeds, but 'updateRecruitment' fails, state will be inconsistent.
+        //  This should use eventual consistency: https://github.com/imotions/carp.core-kotlin/issues/295
+        participantGroup.markAsDeployed()
+        participantRepository.updateRecruitment( recruitment )
 
         return recruitment.getParticipantGroupStatus( participantGroup.id ) { deploymentStatus }
     }
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
@@ -16,6 +16,7 @@ import dk.cachet.carp.studies.application.users.Participant
 import dk.cachet.carp.studies.application.users.ParticipantGroupStatus
 import dk.cachet.carp.studies.domain.users.ParticipantRepository
 import dk.cachet.carp.studies.domain.users.Recruitment
+import dk.cachet.carp.studies.domain.users.StagedParticipantGroup
 import kotlinx.datetime.Clock
 
 
@@ -142,7 +143,7 @@ class RecruitmentServiceHost(
             ?.let { deploymentService.getStudyDeploymentStatus( it.key ) }
         if ( deployedStatus != null && deployedStatus !is StudyDeploymentStatus.Stopped )
         {
-            return recruitment.getParticipantGroupStatus( deployedStatus )
+            return recruitment.getParticipantGroupStatus( deployedStatus.studyDeploymentId ) { deployedStatus }
         }
 
         // Create participant group and mark as deployed.
@@ -156,11 +157,11 @@ class RecruitmentServiceHost(
         //  And probably be separate requests: https://github.com/imotions/carp.core-kotlin/issues/319
         val deploymentStatus = deploymentService.createStudyDeployment( participantGroup.id, protocol, invitations )
 
-        return recruitment.getParticipantGroupStatus( deploymentStatus )
+        return recruitment.getParticipantGroupStatus( participantGroup.id ) { deploymentStatus }
     }
 
     /**
-     * Get the status of all deployed participant groups in the study with the specified [studyId].
+     * Get the status of all participant groups in the study with the specified [studyId].
      *
      * @throws IllegalArgumentException when a study with [studyId] does not exist.
      */
@@ -168,13 +169,10 @@ class RecruitmentServiceHost(
     {
         val recruitment: Recruitment = getRecruitmentOrThrow( studyId )
 
-        // Get study deployment status list.
-        val studyDeploymentIds = recruitment.participantGroups.keys
-        val studyDeploymentStatusList: List<StudyDeploymentStatus> =
-            if ( studyDeploymentIds.isEmpty() ) emptyList()
-            else deploymentService.getStudyDeploymentStatusList( studyDeploymentIds )
-
-        return studyDeploymentStatusList.map { recruitment.getParticipantGroupStatus( it ) }
+        return recruitment.getParticipantGroupStatusList(
+            recruitment.participantGroups.keys,
+            deploymentService::getStudyDeploymentStatusList
+        )
     }
 
     /**
@@ -182,23 +180,29 @@ class RecruitmentServiceHost(
      * of the participant group with the specified [groupId] (equivalent to the studyDeploymentId).
      * No further changes to this deployment will be allowed and no more data will be collected.
      *
-     * @throws IllegalArgumentException when a study with [studyId] or participant group with [groupId] does not exist.
+     * @throws IllegalArgumentException when:
+     *  - a study with [studyId] or participant group with [groupId] does not exist
+     *  - the participant group with [groupId] does not belong to the study with [studyId].
      */
     override suspend fun stopParticipantGroup( studyId: UUID, groupId: UUID ): ParticipantGroupStatus
     {
-        val recruitment = getRecruitmentWithGroupOrThrow( studyId, groupId )
+        val (recruitment, participantGroup) = getRecruitmentWithGroupOrThrow( groupId )
+
+        require( recruitment.id == studyId ) {
+            "Participant group with ID \"$groupId\" does not belong to study with ID \"$studyId\"."
+        }
 
         val deploymentStatus = deploymentService.stop( groupId )
-        return recruitment.getParticipantGroupStatus( deploymentStatus )
+        return recruitment.getParticipantGroupStatus( participantGroup.id ) { deploymentStatus }
     }
 
-    private suspend fun getRecruitmentWithGroupOrThrow( studyId: UUID, groupId: UUID ): Recruitment
+    private suspend fun getRecruitmentWithGroupOrThrow( groupId: UUID ): Pair<Recruitment, StagedParticipantGroup>
     {
-        val recruitment: Recruitment = getRecruitmentOrThrow( studyId )
-        val participations = recruitment.participantGroups[ groupId ]
-        requireNotNull( participations ) { "Study deployment with the specified groupId not found." }
+        val recruitment = participantRepository.getRecruitmentWithParticipantGroup( groupId )
+            ?: throw IllegalArgumentException( "Study deployment with the specified groupId not found." )
 
-        return recruitment
+        val group = checkNotNull( recruitment.participantGroups[ groupId ] )
+        return Pair( recruitment, group )
     }
 
     private suspend fun getRecruitmentOrThrow( studyId: UUID ): Recruitment =
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/ParticipantRepository.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/ParticipantRepository.kt
@@ -21,6 +21,12 @@ interface ParticipantRepository
     suspend fun getRecruitment( studyId: UUID ): Recruitment?
 
     /**
+     * Returns the [Recruitment] which contains the participant group with the specified [groupId],
+     * or null when no recruitment containing the specified participant group exists.
+     */
+    suspend fun getRecruitmentWithParticipantGroup( groupId: UUID ): Recruitment?
+
+    /**
      * Update a [Recruitment] which is already stored in this repository.
      *
      * @throws IllegalArgumentException when no previous version of this recruitment is stored in the repository.
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
@@ -206,24 +206,78 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
     }
 
     /**
-     * Get the [ParticipantGroupStatus] of the study deployment identified by [studyDeploymentStatus].
+     * Get the [ParticipantGroupStatus] for participant groups with [groupIds] in this recruitment.
+     * For deployed participant groups, the matching [StudyDeploymentStatus] is retrieved using
+     * [getDeploymentStatusList].
      *
-     * @throws IllegalArgumentException when the study deployment identified by [studyDeploymentStatus] is not part of this recruitment.
+     * @throws IllegalArgumentException when:
+     * - [groupIds] contains an id which is not part of this recruitment
+     * - [getDeploymentStatusList] doesn't return a matching [StudyDeploymentStatus] for each of the requested IDs
      */
-    fun getParticipantGroupStatus( studyDeploymentStatus: StudyDeploymentStatus ): ParticipantGroupStatus
+    suspend fun getParticipantGroupStatusList(
+        groupIds: Set<UUID>,
+        getDeploymentStatusList: suspend (Set<UUID>) -> List<StudyDeploymentStatus>
+    ): List<ParticipantGroupStatus>
     {
-        val deploymentId = studyDeploymentStatus.studyDeploymentId
-        val group: StagedParticipantGroup = requireNotNull( _participantGroups[ deploymentId ] )
-            { "A study deployment with ID \"$deploymentId\" is not part of this recruitment." }
+        require( participantGroups.keys.containsAll( groupIds ) )
+            { "One of the group IDs a status is requested for isn't part of this recruitment." }
 
-        val participants = group.participantIds.map { id -> _participants.first { it.id == id } }
-        return ParticipantGroupStatus.InDeployment.fromDeploymentStatus(
-            participants.toSet(),
-            group.roleAssignments,
-            studyDeploymentStatus,
-            group.name
-        )
+        val (deployedGroups, stagedGroups) = groupIds
+            .map { participantGroups.getValue( it ) }
+            .partition { it.isDeployed }
+
+        // Get participant group status for staged groups.
+        val stagedGroupStatuses = stagedGroups.map { group ->
+            ParticipantGroupStatus.Staged(
+                id = group.id,
+                participants = getParticipantsFor( group ),
+                assignedParticipantRoles = group.roleAssignments,
+                name = group.name
+            )
+        }
+
+        // Get deployment status for deployed participant groups.
+        val deployedGroupIds = deployedGroups.map { it.id }.toSet()
+        val deploymentStatuses =
+            if ( deployedGroupIds.isEmpty() ) emptyMap()
+            else getDeploymentStatusList( deployedGroupIds ).associateBy { it.studyDeploymentId }
+
+        // Get participant group status for deployed groups.
+        val deployedGroupStatuses = deployedGroups.map { group ->
+            val deploymentStatus = requireNotNull( deploymentStatuses[ group.id ] )
+                { "No study deployment status returned for the requested ID: \"${group.id}\"." }
+            ParticipantGroupStatus.InDeployment.fromDeploymentStatus(
+                getParticipantsFor( group ),
+                group.roleAssignments,
+                deploymentStatus,
+                group.name
+            )
+        }
+
+        return stagedGroupStatuses + deployedGroupStatuses
     }
+
+    /**
+     * Get the [ParticipantGroupStatus] for the participant group with [groupId] in this recruitment.
+     * If the participant group is deployed, the matching [StudyDeploymentStatus] is retrieved using
+     * [getDeploymentStatus].
+     *
+     * @throws IllegalArgumentException when:
+     * - [groupId] is not part of this recruitment
+     * - [getDeploymentStatus] returns a [StudyDeploymentStatus] which doesn't match the requested ID
+     */
+    suspend fun getParticipantGroupStatus(
+        groupId: UUID,
+        getDeploymentStatus: (UUID) -> StudyDeploymentStatus
+    ): ParticipantGroupStatus =
+        getParticipantGroupStatusList( setOf( groupId ) )
+            { deploymentIds -> listOf( getDeploymentStatus( deploymentIds.single() ) ) }.single()
+
+    /**
+     * Get the participants for the participant [group].
+     */
+    private fun getParticipantsFor( group: StagedParticipantGroup ): Set<Participant> =
+        group.participantIds.map { id -> _participants.first { it.id == id } }.toSet()
 
     /**
      * Get an immutable snapshot of the current state of this [Recruitment] using the specified snapshot [version].
--- a/docs/carp-studies.md
+++ b/docs/carp-studies.md
@@ -56,5 +56,5 @@ Allows setting recruitment goals, adding participants to studies, and creating d
 | `getParticipant` | Returns the participant with a specified ID for a study. | manage study: `studyId` | |
 | `getParticipants` | Get all participants for a study. | manage study: `studyId` | |
 | `inviteNewParticipantGroup` | Create and instantly invite a group of previously added participants to a study. | manage study: `studyId` | |
-| `getParticipantGroupStatusList` | Get the status of all deployed participant groups in a study. | manage study: `studyId` | |
+| `getParticipantGroupStatusList` | Get the status of all participant groups in a study. | manage study: `studyId` | |
 | `stopParticipantGroup` | Stop the study deployment in a study of a participant group. No further changes to this deployment will be allowed and no more data will be collected. | manage study: `studyId` | |
--- a/docs/carp-studies.md
+++ b/docs/carp-studies.md
@@ -17,7 +17,7 @@ which reflects the underlying state machine:
 
 ![Participant group state machine](https://i.imgur.com/VIv3HKk.png)
 
-Note: `createParticipantGroup` and `inviteParticipantGroup` are envisioned new endpoints currently not yet available.
+Note: The legacy `inviteNewParticipantGroup` is still available to create and invite in one step, but planned to be removed in the future.
 
 Once a participant group is `InDeployment`, the state of the underlying `studyDeploymentStatus` determines the concrete `ParticipantGroupStatus`.
 Calling `RecruitmentService.stopParticipantGroup()` will stop the underlying deployment, but the deployment can also be stopped by participants in the study.
@@ -55,6 +55,8 @@ Allows setting recruitment goals, adding participants to studies, and creating d
 | `addParticipant` | Add a participant identified by a specified email address or username to a study. | manage study: `studyId` | |
 | `getParticipant` | Returns the participant with a specified ID for a study. | manage study: `studyId` | |
 | `getParticipants` | Get all participants for a study. | manage study: `studyId` | |
-| `inviteNewParticipantGroup` | Create and instantly invite a group of previously added participants to a study. | manage study: `studyId` | |
+| `inviteNewParticipantGroup` | (Deprecated) Create and instantly invite a group of previously added participants to a study. | manage study: `studyId` | |
+| `createParticipantGroup` | Prepare a group of previously added participants for deployment without inviting them yet. | manage study: `studyId` | |
+| `inviteParticipantGroup` | Invite a previously prepared participant group to start participating in the study. | manage study: `studyId` | |
 | `getParticipantGroupStatusList` | Get the status of all participant groups in a study. | manage study: `studyId` | |
 | `stopParticipantGroup` | Stop the study deployment in a study of a participant group. No further changes to this deployment will be allowed and no more data will be collected. | manage study: `studyId` | |
--- a/README.md
+++ b/README.md
@@ -245,8 +245,13 @@ if ( studyStatus.canDeployToParticipants )
     // Create a 'participant group' with a single participant; `AssignedTo.All` assigns the "Patient's phone".
     val participation = AssignedParticipantRoles( participant.id, AssignedTo.All )
     val participantGroup = setOf( participation )
-
-    val groupStatus: ParticipantGroupStatus = recruitmentService.inviteNewParticipantGroup( studyId, participantGroup )
+  
+    val groupId = UUID.randomUUID()
+    var groupStatus: ParticipantGroupStatus =
+        recruitmentService.createParticipantGroup( groupId, participantGroup, studyId )
+    val isStaged = groupStatus is ParticipantGroupStatus.Staged // True.
+  
+    groupStatus = recruitmentService.inviteParticipantGroup( groupId )
     val isInvited = groupStatus is ParticipantGroupStatus.Invited // True.
 }
 ```
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/InMemoryParticipantRepository.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/InMemoryParticipantRepository.kt
@@ -29,6 +29,14 @@ class InMemoryParticipantRepository : ParticipantRepository
         recruitments[ studyId ]?.let { Recruitment.fromSnapshot( it ) }
 
     /**
+     * Returns the [Recruitment] which contains the participant group with the specified [groupId],
+     * or null when no recruitment containing the specified participant group exists.
+     */
+    override suspend fun getRecruitmentWithParticipantGroup( groupId: UUID ): Recruitment? =
+        recruitments.values.firstOrNull { groupId in it.participantGroups.keys }
+            ?.let { Recruitment.fromSnapshot( it ) }
+
+    /**
      * Update a [Recruitment] which is already stored in this repository.
      *
      * @throws IllegalArgumentException when no previous version of this recruitment is stored in the repository.
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/RecruitmentServiceDecorator.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/RecruitmentServiceDecorator.kt
@@ -33,11 +33,29 @@ class RecruitmentServiceDecorator(
     override suspend fun getParticipants( studyId: UUID ) =
         invoke( RecruitmentServiceRequest.GetParticipants( studyId ) )
 
+    @Deprecated(
+        "Use CreateParticipantGroup and InviteParticipantGroup instead",
+        ReplaceWith(
+            "inviteParticipantGroup( createParticipantGroup( UUID.randomUUID(), group, studyId, name ).id )",
+            "dk.cachet.carp.common.application.UUID"
+        )
+    )
+    @Suppress( "DEPRECATION" )
     override suspend fun inviteNewParticipantGroup(
         studyId: UUID,
         group: Set<AssignedParticipantRoles>,
         name: String?
     ) = invoke( RecruitmentServiceRequest.InviteNewParticipantGroup( studyId, group, name ) )
+
+    override suspend fun createParticipantGroup(
+        groupId: UUID,
+        group: Set<AssignedParticipantRoles>,
+        studyId: UUID,
+        name: String?
+    ) = invoke( RecruitmentServiceRequest.CreateParticipantGroup( groupId, group, studyId, name ) )
+
+    override suspend fun inviteParticipantGroup( groupId: UUID ) =
+        invoke( RecruitmentServiceRequest.InviteParticipantGroup( groupId ) )
 
     override suspend fun getParticipantGroupStatusList( studyId: UUID ) =
         invoke( RecruitmentServiceRequest.GetParticipantGroupStatusList( studyId ) )
@@ -49,6 +67,7 @@ class RecruitmentServiceDecorator(
 
 object RecruitmentServiceInvoker : ApplicationServiceInvoker<RecruitmentService, RecruitmentServiceRequest<*>>
 {
+    @Suppress( "DEPRECATION" )
     override suspend fun RecruitmentServiceRequest<*>.invoke( service: RecruitmentService ): Any =
         when ( this )
         {
@@ -58,6 +77,9 @@ object RecruitmentServiceInvoker : ApplicationServiceInvoker<RecruitmentService,
             is RecruitmentServiceRequest.GetParticipants -> service.getParticipants( studyId )
             is RecruitmentServiceRequest.InviteNewParticipantGroup ->
                 service.inviteNewParticipantGroup( studyId, group, name )
+            is RecruitmentServiceRequest.CreateParticipantGroup ->
+                service.createParticipantGroup(groupId, group, studyId, name )
+            is RecruitmentServiceRequest.InviteParticipantGroup -> service.inviteParticipantGroup( groupId )
             is RecruitmentServiceRequest.GetParticipantGroupStatusList ->
                 service.getParticipantGroupStatusList( studyId )
             is RecruitmentServiceRequest.StopParticipantGroup -> service.stopParticipantGroup( studyId, groupId )
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/RecruitmentServiceRequest.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/infrastructure/RecruitmentServiceRequest.kt
@@ -65,12 +65,31 @@ sealed class RecruitmentServiceRequest<out TReturn> : ApplicationServiceRequest<
     }
 
     @Serializable
+    @Deprecated( "Use CreateParticipantGroup and InviteParticipantGroup instead." )
     data class InviteNewParticipantGroup(
         val studyId: UUID,
         val group: Set<AssignedParticipantRoles>,
         val name: String? = null
-    ) :
-        RecruitmentServiceRequest<ParticipantGroupStatus>()
+    ) : RecruitmentServiceRequest<ParticipantGroupStatus>()
+    {
+        override fun getResponseSerializer() = serializer<ParticipantGroupStatus>()
+    }
+
+    @Serializable
+    data class CreateParticipantGroup(
+        val groupId: UUID,
+        val group: Set<AssignedParticipantRoles>,
+        val studyId: UUID,
+        val name: String? = null
+    ) : RecruitmentServiceRequest<ParticipantGroupStatus>()
+    {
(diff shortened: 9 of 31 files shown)


carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
```diff
@@ -74,12 +74,51 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
...
         group: Set<AssignedParticipantRoles>,
         name: String? = null
     ): ParticipantGroupStatus
 
+    /**
+     * Create a new participant [group] of previously added participants for the study with the given [studyId],
+     * and an optional [name] representing this group, but do not yet send out invitations.
+     * This is used to create a group of participants which can be deployed at a later time.
+     *
+     * In case a group with the same participants has already been deployed and is still running (not stopped),
```
Whathecode: This feels weird. Now that create/invite is separated, do we still need this same behaviour? As long as the same parameters are provided, can't we always return the status? Since there is no "invite" expectation now, I think this original restriction no longer makes sense. 

If the caller passes the groupID, the contract is even clearer. 

And, I guess we could allow creating groups after the study has stopped, they just can't be deployed.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
```diff
@@ -74,12 +74,51 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
...
+     * @throws IllegalStateException when the study is not yet ready for deployment.
+     */
+    suspend fun createParticipantGroup(
+        studyId: UUID,
+        group: Set<AssignedParticipantRoles>,
+        name: String? = null
+    ): ParticipantGroupStatus
+
+    /**
+     * Invite the participant group with the specified [groupId] (equivalent to the studyDeploymentId)
```
Whathecode: Isn't the equivalence between deployment ID and groupID an implementation detail? I.e., not something to document/make part of the contract.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
```diff
@@ -74,12 +74,51 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
...
+     * in the study with the given [studyId] to start participating in the study.
+     *
+     * @throws IllegalArgumentException when:
+     *  - a study with [studyId] or participant group with [groupId] does not exist.
+     *  - participant role assignment is empty
```
Whathecode: Wrong documentation.

And, we need to document what happens when the group has already been invited (like the old behavior).

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -124,14 +125,53 @@ class RecruitmentServiceHost(
      *  - not all necessary participant roles part of the study have been assigned a participant
      * @throws IllegalStateException when the study is not yet ready for deployment.
      */
+    @Deprecated(
```
Whathecode, line 128: Are you certain this is needed on both interface and concrete class?

Whathecode (commented):
Only reviewed the interface for now.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
Whathecode: And, considering the idempotent implementation, do we actually really want that?

Take a look [at how this is done for](https://github.com/carp-dk/carp.core-kotlin/blob/develop/carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/application/DeploymentService.kt#L67) `createStudyDeployment`.

I'm leaning towards just being consistent instead of trying to be fancy, unless there is some clear recommendation to have idempotent "create" operations (we should do some pro/con reading). If it is preferred, we should likely create a task to make all create operations in CARP idempotent.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -124,14 +125,53 @@ class RecruitmentServiceHost(
...
+        val participantGroupStatus = createParticipantGroup( studyId, group, name )
+
+        if ( participantGroupStatus is ParticipantGroupStatus.InDeployment )
+        {
+            // Already deployed.
+            return participantGroupStatus
+        }
+
+        // Create study deployment, which sends out invitations.
+        // TODO: If the repository gets updated but `createStudyDeployment` fails, state will be inconsistent.
```
Whathecode, line 162: The repository update and `createStudyDeployment` are no longer visible in this method. Probably more legible to write "if `createParticipantGroup` succeeds, but `inviteParticipantGroup` fails, repository state will be inconsistent.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -145,17 +185,39 @@ class RecruitmentServiceHost(
...
+     *  - any of the participant roles specified in this participant group does not exist
+     *  - not all necessary participant roles part of the study have been assigned a participant
+     * @throws IllegalStateException when group has already been deployed.
+     */
+    override suspend fun inviteParticipantGroup( studyId: UUID, groupId: UUID ): ParticipantGroupStatus
+    {
+        val (recruitment, participantGroup) = getRecruitmentAndGroupOrThrow( studyId, groupId )
+        check( !participantGroup.isDeployed ) { "Participant group has already been deployed." }
+
+        val (protocol, invitations) = recruitment.createInvitations( participantGroup.roleAssignments )
```
Whathecode, line 216: The original code mentioned:
>         //  This should use eventual consistency: https://github.com/imotions/carp.core-kotlin/issues/295

I believe that is still relevant, since there is a call to `deploymentService` here. That call may succeed, but subsequent repo update may fail.

Not that this has to be fixed in this PR, but the comment/task should remain.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -169,12 +231,15 @@ class RecruitmentServiceHost(
...
+            getStagedParticipantGroupStatus( it, recruitment )
+        }
         val studyDeploymentStatusList: List<StudyDeploymentStatus> =
-            if ( studyDeploymentIds.isEmpty() ) emptyList()
-            else deploymentService.getStudyDeploymentStatusList( studyDeploymentIds )
+            if ( deployedStudyDeploymentIds.isEmpty() ) emptyList()
+            else deploymentService.getStudyDeploymentStatusList( deployedStudyDeploymentIds )
 
-        return studyDeploymentStatusList.map { recruitment.getParticipantGroupStatus( it ) }
+        return stagedParticipantGroups + studyDeploymentStatusList.map { recruitment.getParticipantGroupStatus( it ) }
```
Whathecode: I was wondering why you updated this. But, I see it is because you also return _staged_ participants now. Make sure to update the documentation (it mentiones "all deployed participant groups"). I guess, since before you couldn't have just "staged" participants, we can still consider this fully backwards compatible.

You can make this code more readable by not intermingling getting the status of staged groups with that of deployed groups. They could be two separate "code blocks" with a comment above.

It's also not great that the status of one is created _within_ `recruitment`, and the other as a private method here. Can't both be encapsulated in `recruitment`? Maybe even the same method but with a nullable deployment status (meaning, not deployed).

Whathecode (commented):
Reviewed `RecruitmentServiceHost`. Primarily, we should try to come to a decision on whether or not we want idempotency and then pursue something consistent.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
Whathecode: I think [the main reason for this being idempotent in the past](https://github.com/carp-dk/carp.core-kotlin/issues/125#issue-613229634) is resolved by both splitting these calls, and letting the caller pass the ID. 

Given the ID (if we add it), the create call will fail a second time around. And, the invite call simply shouldn't send out invites again. There is no need to verify parameters to assess whether it is a new or existing group.

It's both consistent and simpler.

yuanchen233: I agree that this behavior is no longer needed.

> I'm leaning towards just being consistent instead of trying to be fancy, unless there is some clear recommendation to have idempotent "create" operations 

In the current flow of creating a new study deployment (in CAWS at least):  `createStudyDeployment` is called by `RecruitmentService.inviteNewParticipantGroup`. Therefore a groupID is provided by the recruitment subsystem, which keeps all `UUID` generation in the backend.

We also have example of [createStudy](https://github.com/carp-dk/carp.core-kotlin/blob/50ef1471e20639837c4b1a1fdb95d79748abd0da/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyServiceHost.kt#L27) which generate a new ID for the study.

Side note: what happens if `CreateStudyDeployment` called directly, would It be a deployment not tracked by any study?

----
 Actually, I believe allowing the caller to pass the groupID helps maintain isolation between subsystems. This leaves the potential of creating study deployments first, and add them to a study later; or separate `Study` subsystem and `StudyDeployment` subsystem completely.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
yuanchen233: I'm not following this, should empty role assignment be `IllegalStateException`?

I'm throwing `IllegalStateException` when group has already been deployed. Should I keep the old begavior to return running group's status here?

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
yuanchen233: Gradle and our detekt are okay with it.

I added it because IntelliJ warns about overriding a deprecated member but not mark it as deprecated.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
yuanchen233: This comment were left untouched in old `inviteNewParticipantGroup` method, I will move it down to new endpoint.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
yuanchen233: This change could probably be a separate commit — my main goal was just to make it possible to get info about `StagedParticipantGroup` after creating them.

It was pulled out into a private method because:
1. The multi-layer nested `.map` calls didn’t look clean; and  
2. Building a lookup table seemed potentially more efficient.

I was trying to find a good place for the method that builds `ParticipantGroupStatus.Staged`.  
Since it depends on `recruitment`, it couldn’t live inside `StagedParticipantGroup`.  
Now that you mention it, `recruitment` does seem like the right place.

I’m thinking of adding it as a separate method in `recruitment`.  
`DeployedGroups` needs `deploymentStatus`, while `StagedParticipantGroup` only needs `groupID`;  
but since `getParticipantGroupStatus` returns a `ParticipantGroupStatus`, would method overloading make sense here?

> 💡 By `StagedParticipantGroups` I mean non-deployed groups — maybe we should consider renaming it for clarity. I think you also mentioned rename it in an earlier issue to `ParticipantGroup`, but it's taken by deployment subsystem already..

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
Whathecode: > We also have example of [createStudy](https://github.com/carp-dk/carp.core-kotlin/blob/50ef1471e20639837c4b1a1fdb95d79748abd0da/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/StudyServiceHost.kt#L27) which generate a new ID for the study.

If we ever need to refactor that, it would make sense to align and let the caller pass an ID. All in all, I seem to recall that is best practice. So, maybe something you can add as an issue to the backlog. :) (not urgent though, I bet)

> Side note: what happens if CreateStudyDeployment called directly, would It be a deployment not tracked by any study?

Yes; the whole point is you can have deployments without studies, or different study subsystems could decide how to manage their deployments. Or, you can use a hardcoded deployment for a singular app, etc ...

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
Whathecode: > any of the participant roles specified in [group] does not exist

Can this ever happen? This is specified when you create the group (or later when you update it). So, a group with invalid participant roles can't exist at this point in time. That's an invariant protected by this service.

> not all necessary participant roles part of the study have been assigned a participant

That one I guess can still happen. And, can only be resolved once you allow the user to update (but it's fine that isn't part of this commit).

I guess it also covers "participant role assignment is empty" which is redundant then?

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
Whathecode: Yes, but it is also relevant for the new endpoint. :)

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
yuanchen233: Got it, thanks!

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
yuanchen233: I remembered keeping that behaviour so the deprecated inviteNewParticipantGroup would keep working; that logic now lives inside the legacy method. 

I also added a guard in createParticipantGroup so if a group already exists under that groupId but its config doesn’t match, we let users know a new group wasn’t created.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
```diff
@@ -74,14 +74,53 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
...
+     *
+     * When [groupId] identifies an existing group, the latest status for this group is returned.
+     *
+     * @throws IllegalArgumentException when:
+     *  - a study with [studyId] does not exist
+     *  - an existing participant group with [groupId] has a different configuration
+     *  - any of the participant roles specified in [group] does not exist
+     * @throws IllegalStateException when the study is not yet ready for deployment.
+     */
+    suspend fun createParticipantGroup(
```
Whathecode, line 104: We also need to add this endpoint to the docs: https://github.com/carp-dk/carp.core-kotlin/blob/develop/docs/carp-studies.md
Maybe I'll take that on me, as the state diagram should probably also be updated. But, you could already update the text, though. :)

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
```diff
@@ -74,14 +74,53 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
...
         name: String? = null
     ): ParticipantGroupStatus
 
     /**
-     * Get the status of all deployed participant groups in the study with the specified [studyId].
+     * Create a new participant [group] of previously added participants for the study with the given [studyId],
+     * and an optional [name] representing this group, but do not yet send out invitations.
+     * This is used to create a group of participants which can be deployed at a later time.
+     *
+     * When [groupId] identifies an existing group, the latest status for this group is returned.
```
Whathecode: I was actually thinking of just throwing here, consistent with `DeploymentService`. https://github.com/carp-dk/carp.core-kotlin/blob/develop/carp.deployments.core/src/commonMain/kotlin/dk/cachet/carp/deployments/application/DeploymentService.kt#L60

Or, consistent with `inviteParticipantGroup` in this service which throws `IllegalStateException`.

Do you see a benefit in doing this instead?

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -124,57 +125,153 @@ class RecruitmentServiceHost(
...
+            return participantGroupStatus
         }
 
-        // Create participant group and mark as deployed.
-        val participantGroup = recruitment.addParticipantGroup( group, name, uuidFactory.randomUUID() )
-        participantGroup.markAsDeployed()
+        // Create study deployment, which sends out invitations.
+        // TODO: If createParticipantGroup succeeds, but inviteParticipantGroup fails, repository state will be inconsistent.
+        //  This should use eventual consistency: https://github.com/imotions/carp.core-kotlin/issues/295
+
```
Whathecode, line 166: Remove newline. This comment belongs to the `return` that follows immediately.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -124,57 +125,153 @@ class RecruitmentServiceHost(
...
+        val existingByConfiguration = recruitment.participantGroups.entries.firstOrNull { (_, existingGroup) ->
+            existingGroup.roleAssignments == group && existingGroup.name == name
+        }
+        if ( existingByConfiguration != null )
+        {
+            val (existingId, existingGroup) = existingByConfiguration
+            if ( !existingGroup.isDeployed )
+            {
+                // A staged group should still go through the deprecated flow and trigger invitations.
+                return inviteParticipantGroup( studyId, existingId )
```
Whathecode: I think I can see what you tried to do here, but, this wasn't part of the old `inviteNewParticipantGroup` behavior, and neither is this inviting a _new_ participant group as per the contract of this method.

Instead, it's supporting a call to the new `createParticipantGroup` followed by a call to `inviteNewParticipantGroup`, as far as I can follow.

It's also not addressing the main comment above:
>  Support the deprecated API by reusing groups that were created through repeated calls with the same configuration ...

On a more personal note, the multiple returns hamper readability. They are great for clear early outs (such as the previous "existing group" edge case), but now there are 4 returns, two of which to the same method, with space between them. The behavior is less clear at a glance.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
```diff
@@ -74,14 +74,53 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
...
+     *
+     * @throws IllegalArgumentException when:
+     *  - a study with [studyId] does not exist
+     *  - an existing participant group with [groupId] has a different configuration
+     *  - any of the participant roles specified in [group] does not exist
+     * @throws IllegalStateException when the study is not yet ready for deployment.
+     */
+    suspend fun createParticipantGroup(
+        studyId: UUID,
+        groupId: UUID,
```
Whathecode, line 105: Having this as the first parameter would be more consistent.

And, having `groupId` and `group` be followed by `studyId` could reduce the change of passing wrong IDs (`groupId` to `studyId` and vice-versa).

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -124,57 +125,153 @@ class RecruitmentServiceHost(
...
+     * @throws IllegalStateException when group has already been deployed.
+     */
+    override suspend fun inviteParticipantGroup( studyId: UUID, groupId: UUID ): ParticipantGroupStatus
+    {
+        val (recruitment, participantGroup) = getRecruitmentAndGroupOrThrow( studyId, groupId )
+        check( !participantGroup.isDeployed ) { "Participant group has already been deployed." }
 
-        // Create study deployment, which sends out invitations.
         // TODO: If the repository gets updated but `createStudyDeployment` fails, state will be inconsistent.
         //  This should use eventual consistency: https://github.com/imotions/carp.core-kotlin/issues/295
```
Whathecode: If `createStudyDeployment` fails, this method would fail. The problem is different now: `createStudyDeployment` may succeed, but `updateRecruitment` may fail.

This comment applies to the `"Mark participant group as deployed" code block now.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -124,57 +125,153 @@ class RecruitmentServiceHost(
...
+
+        // Get status list for deployed participant groups.
         val studyDeploymentStatusList: List<StudyDeploymentStatus> =
-            if ( studyDeploymentIds.isEmpty() ) emptyList()
-            else deploymentService.getStudyDeploymentStatusList( studyDeploymentIds )
+            if ( deployedStudyDeploymentIds.isEmpty() ) emptyList()
+            else deploymentService.getStudyDeploymentStatusList( deployedStudyDeploymentIds )
 
-        return studyDeploymentStatusList.map { recruitment.getParticipantGroupStatus( it ) }
+        return stagedParticipantGroups + studyDeploymentStatusList.map { recruitment.getParticipantGroupStatus( it ) }
```
Whathecode: For consistency, move the `.map { recruitment.getParticipantGroupStatus( it ) }` to the "Get status list for deployed participant groups" block.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -124,57 +125,153 @@ class RecruitmentServiceHost(
...
     {
         val recruitment: Recruitment = getRecruitmentOrThrow( studyId )
 
+        // Get status list for staged participant groups.
+        val stagedParticipantGroups = recruitment.participantGroups
+            .filter { !it.value.isDeployed }
+            .values
+            .map { recruitment.getParticipantGroupStatus( it ) }
+
         // Get study deployment status list.
```
Whathecode: This is getting the IDs, ... not the status list. And, this code block is no longer meaningful. Just inline it with "Get status list for deployed participant groups" below.

Whathecode (changes requested):
Still reviewing, but here are some further comments. :)

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
yuanchen233: Yep, I was trying to reuse some code, as the plan was to separate creation and invitation to participant groups, but the behaviour of those endpoints are changed now. I went to far on this route :(

I'll revert this endpoint/method and leave it as is.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
yuanchen233: The only potential benefit would be that the user can get the status of the group back using the same method call. However, since `GetParticipantGroupStatusList` now includes staged groups, it should be the recommended approach.

I’ll go with `IllegalStateException`, as the provided argument is valid, but "an object is in an invalid state for that operation.” We should agreed on which one to use and make it consistent across the system.

---
Update: I ended up with `IllegalArgumentException`, as "the caller supplied parameters that violate the method’s precondition".

I recall that I have seen people arguing about the usage between `IllegalArgumentException` vs `NullPointerException`, when the provided non-nullable argument is `null` a long time ago, now here I am :)

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
Whathecode: It's deprecated code; no harm in not maintaining it. ;)

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
yuanchen233: Should this be applied to `inviteParticipantGroup` and `stopParticipantGroup` as well? Since they also take `studyId` and `groupId`.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
Whathecode: It is indeed an illegal argument. 😉 The object which is supposed to safeguard state doesn't exist yet, so that'd be odd.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
Whathecode: Well spotted! it is indeed a good convention to have endpoints operating on an object(s) take the identifier of said object as the first parameter. Even more important, to be consistent. Go from the most important/relevant parameters to secondary ones; it helps intuition in the API.

yuanchen233: I updated `inviteParticipantGroup` regarding this change, I'll update `stopParticipantGroup` in a separate commit later on to keep this commit focused.

Whathecode: We could also consider removing `studyId` entirely. If groups have unique IDs, `studyID` shouldn't be needed after creation. 🤔 But yeah, separate change, maybe even other PR.

yuanchen233: Yep, we could. I kept `studyId` here since `getRecruitmentOrThrow` requires it, we currently don't keep track of `studyId` in participant groups or deployments, and loop through all recruitments are inefficient.

We could keep `studyId` inside each `StagedParticipantGroup` to support this.

Whathecode: For optimization it is maybe a bit premature. Unless you run massive studies you won't feel this, and then the first limitation will likely be persistence instead. 😉

But, from a design perspective I guess there is no more reason to do this differently. I think I just took some shortcuts back when I designed this.

yuanchen233: I added a new helper `getRecruitmentAndGroupByGroupIdOrThrow` to do this, and `stopParticipantGroup` can reuse it later.
(better name for this are very much welcome

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
```diff
@@ -74,14 +74,49 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
...
     /**
-     * Get the status of all deployed participant groups in the study with the specified [studyId].
+     * Create a new participant [group] of previously added participants for the study with the given [studyId],
+     * and an optional [name] representing this group, but do not yet send out invitations.
+     * This is used to create a group of participants which can be deployed at a later time.
+     * @throws IllegalArgumentException when:
+     *  - a study with [studyId] does not exist
+     *  - an existing participant group with [groupId] already exists
+     *  - any of the participant roles specified in [group] does not exist
+     * @throws IllegalStateException when the study is not yet ready for deployment.
```
Whathecode, line 102: Re-reading this again, I think this exception warrants a bit more explanation. I would add:

> As long as no final study protocol is locked in for the study, a participant group can't be created since participant roles to which participants need to be assigned are unknown.

Don't forget to also update `RecruitmentServiceHost`. :)

That's at least where the current limitation comes from. But, that does seem limiting and perhaps a useful area to improve functionality for later on.

In the main use case there is only a single participant and participant roles are irrelevant. That's why `AssignTo.All` exists. If group only contains `AssignTo.All`, that should be a valid configuration for any study protocol, even if it changes after.

And/or, maybe it is fine to only validate the assignments in context of the _current_ protocol, make the assignments nullable, and simply invalidate existing role assignments when a new protocol is set.

Does this make sense? If so, we should likely spin off a new feature/discussion issue for this.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
```diff
@@ -74,14 +74,49 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
...
+        group: Set<AssignedParticipantRoles>,
+        studyId: UUID,
+        name: String? = null
+    ): ParticipantGroupStatus
+
+    /**
+     * Invite the participant group with the specified [groupId] to start participating in its study.
+     *
+     * @throws IllegalArgumentException when:
+     *  - the participant group with [groupId] does not exist.
```
Whathecode: Minor: drop punctuation at the end. Seemingly it's not used in any of the other bullet lists.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -204,4 +277,17 @@ class RecruitmentServiceHost(
     private suspend fun getRecruitmentOrThrow( studyId: UUID ): Recruitment =
         participantRepository.getRecruitment( studyId )
             ?: throw IllegalArgumentException( "Study with ID \"$studyId\" not found." )
+
+    private suspend fun getRecruitmentAndGroupByGroupIdOrThrow(
```
Whathecode: I think you can safely drop "ByGroupId" as that is already documented by the parameter name.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/ParticipantRepository.kt
```diff
@@ -20,6 +20,11 @@ interface ParticipantRepository
      */
     suspend fun getRecruitment( studyId: UUID ): Recruitment?
 
+    /**
+     * Returns all stored [Recruitment]s.
+     */
+    suspend fun getRecruitments(): List<Recruitment>
```
Whathecode: This is a very bad idea. It's a full table dump and will inevitably get slower and slower.

I guess you added this to be able to drop the `studyId` parameter. But, the way the repository pattern in DDD works is that you are supposed to expose the _queries_ you need from the domain layer in the interface (the use cases), instead of the very typical simplistic CRUD operations and end up writing query code in the layer above.

So, instead what we likely need is a `getRecruitmentWithParticipantGroup( groupId )`. This can be optimized by whatever underlying DB technology and doesn't pull in unnecessary data.

Whathecode (changes requested):
Some minor comments, but then I ran into the `getRecruitments()` method on the repository, which is a bigger problem and I'll already post now so you can fix it while I continue reviewing later on. :)

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
yuanchen233: Yes, this feature would decouple study setup from inviting participants. So as a researcher, I can recruit participant before finalize technical details "protocol".

> In the main use case there is only a single participant and participant roles are irrelevant. That's why AssignTo.All exists. If group only contains AssignTo.All, that should be a valid configuration for any study protocol, even if it changes after.

Since protocol also defines the number of participant, `AssignTo.All` could be misused here. But yep, the discussion should be hold in a separate issue.

yuanchen233: > Re-reading this again, I think this exception warrants a bit more explanation. I would add:
> As long as no final study protocol is locked in for the study, a participant group can't be created since participant roles to which participants need to be assigned are unknown.

Should this also be added to `Recruitment.addParticipantGroup`'s documentation, or the error message?

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
Whathecode: I'm afraid you squashed fixups for this into the wrong commit (the second, instead of the first).

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
Whathecode: I'm afraid you squashed fixups for this into the wrong commit (the second, instead of the first).

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
Whathecode: I'm afraid you squashed fixups for this into the wrong commit (the second, instead of the first).

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/ParticipantRepository.kt
Whathecode: I'm afraid you squashed fixups for this into the wrong commit (the second, instead of the first).

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
```diff
@@ -74,14 +74,51 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
...
+     * and an optional [name] representing this group, but do not yet send out invitations.
+     * This is used to create a group of participants which can be deployed at a later time.
+     * @throws IllegalArgumentException when:
+     *  - a study with [studyId] does not exist
+     *  - an existing participant group with [groupId] already exists
+     *  - any of the participant roles specified in [group] does not exist
+     * @throws IllegalStateException when the study is not yet ready for deployment;
+     *  As long as no final study protocol is locked in for the study, a participant group can't be created
+     *  since participant roles to which participants need to be assigned are unknown.
+     */
```
Whathecode: If I'm not mistaken, you shouldn't have text underneath `@throws`. The documentation is first, and then additional stuff. So this should be part of the "main" documentation.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/ParticipantRepository.kt
```diff
@@ -20,6 +20,14 @@ interface ParticipantRepository
      */
     suspend fun getRecruitment( studyId: UUID ): Recruitment?
 
+    /**
+     * Returns the [Recruitment] and [StagedParticipantGroup] which contains the specified [groupId],
+     * or null when no participant group with that ID exists.
+     */
+    suspend fun getRecruitmentWithParticipantGroup(
+        groupId: UUID
+    ): Pair<Recruitment, StagedParticipantGroup>?
```
Whathecode: I'm not sure returning `StagedParticipantGroup` in addition to `Recruitment` instead of just `Recruitment` makes sense. It feels inconsistent with the rest of this repository which focuses on `Recruitment`. Also, with the `groupId` retrieving the `StagedParticipantGroup` is a single line of code (and a map lookup) for the caller. So the win vs. added complexity here is extremely minimal.

At the very least, the documentation is confusing. It now reads like it returns the `StagedParticipantGroup` chich contains the specified `groupId`. :)

I may have set you on the wrong track with my suggested naming "WithParticipantGroup". But, I what I meant by that was that the recruitment has that participant group. :) Maybe `getRecruitmentContainingParticipantGroup` or `getRecruitmentForParticipantGroup` is better? Whathever name you land on, it should be clearer from the documentation. :)

yuanchen233: The name is fine, we also have `getRecruitmentWithGroupOrThrow` and keep name aligned makes sense for consistency.

I kept the signature this way because repeating a `recruitment[groupId]` null check after repository already validated it felt redundant; while since repository implementations might be rewritten in implementations later, being a bit defensive here could pay off later. I couldn't make up my mind so I returned both 😶

I also removed the pair return for the helper method in `recruitmentServiceHost` to keep code in a similar style; and overloaded `getRecruitmentWithGroupOrThrow` name for this method since the parameter is different and that's the most suitable name.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -160,21 +167,91 @@ class RecruitmentServiceHost(
...
+        group: Set<AssignedParticipantRoles>,
+        studyId: UUID,
+        name: String?
+    ): ParticipantGroupStatus
+    {
+        val recruitment = getRecruitmentOrThrow( studyId )
+
+        require( recruitment.participantGroups[ groupId ] == null ) {
+            "Participant group with ID \"$groupId\" already exists."
+        }
```
Whathecode: This constraint should be enforced by `Recruitment.addParticipantGroup`, not the application service.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -160,21 +168,86 @@ class RecruitmentServiceHost(
...
+    override suspend fun createParticipantGroup(
+        groupId: UUID,
+        group: Set<AssignedParticipantRoles>,
+        studyId: UUID,
+        name: String?
+    ): ParticipantGroupStatus
+    {
+        val recruitment = getRecruitmentOrThrow( studyId )
+        require( participantRepository.getRecruitmentWithParticipantGroup( groupId ) == null )
+            { "Participant group with ID \"$groupId\" already exists." }
```
Whathecode, line 192: @yuanchen233 Note that I changed this to check for _any_ `groupId`, instead of an existing `groupId` within the `recruitment`.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -160,21 +168,86 @@ class RecruitmentServiceHost(
...
+     * Invite the participant group with the specified [groupId] to start participating in its study.
+     *
+     * @throws IllegalArgumentException when:
+     *  - the participant group with [groupId] does not exist
+     *  - not all necessary participant roles part of the study have been assigned a participant
+     * @throws IllegalStateException when group has already been deployed.
+    */
+    override suspend fun inviteParticipantGroup( groupId: UUID ): ParticipantGroupStatus
+    {
+        val (recruitment, participantGroup) = getRecruitmentWithGroupOrThrow( groupId )
```
Whathecode, line 212: @yuanchen233 I rolled this back to return the pair, because in this context it does serve a purpose. And, it's also private so less impactful. For the repository interface, I just didn't want to cause confusion as to the group being contained _in_ the recruitment.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -216,15 +216,34 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
...
             group.name
         )
     }
 
+    /**
+     * Get the [ParticipantGroupStatus] for the staged participant [group].
+     *
+     * @throws IllegalArgumentException when the specified [group] is not part of this recruitment.
+     */
+    fun getParticipantGroupStatus( group: StagedParticipantGroup ): ParticipantGroupStatus.Staged
```
Whathecode: There are some design issues here:
- The caller can pass _any_ group (an actual deployed group, or even from an entirely different recruitment) and that would go unnoticed.
- The logic of whether to call this method or the one with the `StudyDeploymentStatus` parameter really shouldn't live in the application service.

I think this calls for an external "domain service" (in DDD terminology) which aggregates the two together (study deployments and participant groups) and is called from the application service. So both your new `getParticipantGroupStatus`, but also the previous one I added, should live elsewhere (and probably be merged). Maybe a `ParticipantGroupStatusService`. That would make a logical preceding commit.

`participantGroups` also still documents "deployment IDs", but that must group IDs, now. Slightly related, the naming/documentation of `StagedParticipantGroup` doesn't feel spot on. It starts out staged, for sure, but it seems a bit of a misnomer once deployed. :) Since we plan to allow to update this even _after_ deployment (name change, for example), we should maybe start thinking about a better name 🤔. I can't think of any atm.

Whathecode (changes requested):
The small stuff I simply amended myself. But, I noticed a problem in `Recruitment` which I think warrants the introduction of a `ParticipantGroupStatusService` (or similar if you can come up with a better name :) ).

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
yuanchen233: > The caller can pass any group (an actual deployed group, or even from an entirely different recruitment) and that would go unnoticed.

We throw IllegalArgumentException for this, as we `requireNotNull( _participantGroups[ deploymentId ]`. For deployed groups, it could be a simple `isDeployed` check. I do get the point for a "domain service" here, after some reading on DDD :)

Hi @Whathecode, I am not sure if I'm on the right track here, so I pushed what I have to 'fail early'.

I had different approaches:

1. Passing the current `recruitment` to `ParticipantGroupStatusService.getParticipantGroupStatus()`, so it can check membership, and have access to `participants` by moving `getParticipantsFor` here; and also it can check if a group `isDeployed`.
2. Also making `ParticipantGroupStatusService` depend on `StudyDeploymentService`; then `getParticipantGroupStatus` only requires the `groupId`, and `StudyDeploymentStatus` can be fetched.
3. Current implementation, I am not passing current `recruitment` as it against the purpose of this new service: arguably it should be free of recruitment. I didn't go with approach 2 either because domain service should not depend on another application service, per my understanding so far.

Therefore, as current implementation it looks more like a `factory` rather than `service` as it only builds group status from provided arguments.

====
For names, I am thinking `recruitmentGroup`, `recruitmentEnrollment`, or `EnrollmentGroup` (or Enrolment for British English).

I don't want to get confused with `Deployment.ParticipantGroup`, I'll ask the team for more ideas :wink:

Whathecode: If we consider studies and deployments truly separate domains, DDD would argue there is no harm overloading the term, and the term having a different meaning depending on context: "deployment" participant groups vs. "study" participant groups. :)

yuanchen233: We could also do that, and referring them with 'studies' and `deployment` prefix, it's more natural as it corresponds to `ParticipantGroupStatus`; however, `Staged` state feel more like 'Configured' or 'Planned' tho.

Whathecode: Finally looked at this; sorry for the slow responses. :(

I'm thinking a static `ParticipantGroupStatusService` (class with companion object in Kotlin), and not a class with interface. That can live in the domain layer.

Then I am considering a: `ParticipantGroupStatusService.forGroup( group: StagedParticipantGroup, getDeploymentStatus: (UUID) -> StudyDeploymentStatus)` call. The service can conditionally call `getDeploymentStatus` based on `isDeployed`. If you already have the status, you can just inject a lambda returning that (on create). When you don't, you can inject a call to `deploymentService`.

 I see no need to DI this into `RecruitmentServiceHost`, as this opens up the possibility of injecting a different `DeploymentService` into `RecruitmentServiceHost` than that used by calls to `ParticipantGroupStatusService`, which always seems wrong to me. I think information hiding still wins out here. The only benefit of injecting it would be that you can test the implementation does calls to it where expected. But, that is testing the _implementation_ and not the desired _behavior_.

And, to support `getParticipantGroupStatusList` the method should generalize to `forGroups` instead, so that we can benefit from the single aggregate `getStudyDeploymentStatusList` call instead of multiple calls.

Lastly, for review purposes, I think it could make sense to have an initial commit introducing this service and replacing the existing implementation, before introducing the new `create/invite` endpoints.

docs/carp-studies.md
```diff
@@ -55,6 +55,8 @@ Allows setting recruitment goals, adding participants to studies, and creating d
 | `addParticipant` | Add a participant identified by a specified email address or username to a study. | manage study: `studyId` | |
 | `getParticipant` | Returns the participant with a specified ID for a study. | manage study: `studyId` | |
 | `getParticipants` | Get all participants for a study. | manage study: `studyId` | |
-| `inviteNewParticipantGroup` | Create and instantly invite a group of previously added participants to a study. | manage study: `studyId` | |
+| `inviteNewParticipantGroup` | (Deprecated) Create and instantly invite a group of previously added participants to a study. | manage study: `studyId` | |
+| `createParticipantGroup` | Prepare a group of previously added participants for deployment without inviting them yet. | manage study: `studyId` | |
+| `inviteParticipantGroup` | Invite a previously prepared participant group to start participating in the study. | manage study: `studyId` | |
```
yuanchen233, line 60: For `inviteParticipantGroup`, the recommended claim is manage study, but the endpoint now only requires `groupId`, is this still applicable here? since `getParticipant` and `stopParticipantGroup` will also require only `groupId` in the future.

They are still part of `RecruitmentService`, but not requiring `studyId` as an the argument, so maybe document them separately?

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
yuanchen233: End of years are just busy times..Thanks for the details!
I'm also passing the `participants` to make this service free from `recruitment`.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/ParticipantGroupStatusService.kt
```diff
@@ -0,0 +1,102 @@
...
+            val deployedIds = deployedGroups.map { it.id }.toSet()
+            val deploymentStatuses =
+                if ( deployedIds.isEmpty() ) emptyList()
+                else getDeploymentStatusList( deployedIds )
+
+            val deploymentStatusIds = deploymentStatuses.map { it.studyDeploymentId }.toSet()
+            require( deploymentStatusIds == deployedIds )
+            {
+                "Deployment status IDs \"$deploymentStatusIds\" don't match requested group IDs \"$deployedIds\"."
+            }
```
yuanchen233: There is a change in strictness, `Recruitment.getParticipantGroupStatus` only require each deployment status belongs to current recruitment, now we also check if there are missing/exceeding deployment status compare to participant groups.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
Whathecode: Looking at it closer now, by injecting the `getDeploymentStatusList` delegate, _maybe_ there is no real problem anymore keeping this functionality in `Recruitment`. 

I started adding commits on top of yours in that direction. If you don't see problems with this, feel free to squash these into your original commit introducing `ParticipantGroupStatusService` (and of course update the commit message), and continue on the same path.

I now see this breaks a detekt test since the max number of methods in `Recruitment` has been exceeded, which _could_ warrant maybe considering `ParticipantGroupStatusService`. :) It can easily be moved, or we could suppress the detekt warning. But, more importantly, I don't see the real problem you had with making this functionality free from `Recruitment` 🤔. And, rather than duplicating logic, I let the singular method call the plural method.

yuanchen233: My consideration was building `ParticipantGroupStatus` needs `Participants` which is persisted in the recruitments, therefore the caller needs an instance of recruitment to use `ParticipantGroupStatusService`. By using a `getParticipants` delegate like `getDeploymentStatusList`, I don't think it's a concern anymore.

I made `getParticipantsFor` an internal method instead private call it from the host, that adds one more method, changing it back should solve the detekt problem. However, we are still at maximum number of methods.

Looking at the list of methods, can the 3 overloaded `addParticipant` be optimised into one?
I think methods adding participant by email and username is not used currently, the logic is handled in recruitment host.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/ParticipantGroupStatusService.kt
Whathecode: When I refactored this I noticed this, and I thought it best to be accepting (so if there are "exceeding" deployment status returned, not fail, but of course we need the status for those requested, and we want to fail early if anything is missing.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
Whathecode: > the caller needs an instance of recruitment to use ParticipantGroupStatusService.

But why is that a problem? That seems fine to me. 🤔 The delegate for `getDeploymentStatusList` is to prevent coupling from the domain layer to the application service (`RecruitmentService`). The same doesn't apply to retrieving participants through `Recruitment` (which itself is domain layer).

Note: typically there should be _no_ reference from the `domain` to `application` (such as `Recruitment` depending on `StudyDeploymentStatus`). But, I seem to recall that I ended up placing a lot of immutable value types which really belong more to the domain in the application layer because I expose them "as is" through the application layer. That's because of how powerful `kotlinx.serialization` is; I really didn't feel like creating DTOs for each of them. They are serializable as is already within the domain. Then, to make clear what the external surface area is of the application service (and for which schemas need to be created, etc.), I prioritized placing these in `application`.

> I made getParticipantsFor an internal method instead private call it from the host, that adds one more method,

True! I think `ParticipantGroupStatusService` may have been premature. Let's just keep things in `Recruitment` for now. We can always refactor later.

> Looking at the list of methods, can the 3 overloaded addParticipant be optimised into one?

If anything, these warrant suppressing the warning as the overloads really don't add complexity. It's used in tests and removes some boilerplate. As long as `AccountIdentity` is a sealed class hierarchy (and thus not an extension point) I think the overloads make perfect sense to keep.

yuanchen233: Got it now, thanks, I was confused about where we should use a domain service in this project and the constrains, I found `ParticipantGroupService` and trying to follow it's structure and use case.

yuanchen233: Added a guard in `stopParticipantGroup` to check if the provided group belongs to the provided study, since we don't use `studyId` to get recruitment anymore.

Whathecode: > I found ParticipantGroupService and trying to follow it's structure and use case.

Interestingly, I see there are also application services injected into the domain service there, which never really seemed to have bothered me. 😅 It may not be 100% DDD, but at least its just interfaces, so maybe not too bad. Injecting the get deployment delegate like we did isn't too different from injecting an interface, except that it applies the interface segregation principle and injects the minimal subset of the interface it needs.🤷🏼‍♂️

I think I may need a refresher on DDD to fully understand how much we should care about this 'violation'.🙂 As long as we don't inject anything concrete it seems fine to me.

Whathecode: > I found ParticipantGroupService and trying to follow it's structure and use case.

Interestingly, I see there are also application services injected into the domain service there, which never really seemed to have bothered me. 😅 It may not be 100% DDD, but at least its just interfaces, so maybe not too bad. Injecting the get deployment delegate like we did isn't too different from injecting an interface, except that it applies the interface segregation principle and injects the minimal subset of the interface it needs.🤷🏼‍♂️

I think I may need a refresher on DDD to fully understand how much we should care about this 'violation'.🙂 As long as we don't inject anything concrete it seems fine to me.

At least [some people online](https://softwareengineering.stackexchange.com/a/456594/15464) seem to think the same and argue it could be fine to place said interfaces in the domain layer. (Not that I think we should start moving stuff around.) It makes sense; it's the request objects that are the 'true' application layer.

docs/carp-studies.md
Whathecode: Yes it is. Otherwise people who don't manage the study can invite participants. 🤷🏼‍♂️

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
yuanchen233: Sounds good, we have some thing to follow in this project on this matter now.

docs/carp-studies.md
yuanchen233: Yes the claim should be the same, just documentations here.

After removing the `studyId` argument, we can no longer enforce the provided group belongs to the desired study, it's up to the implementations to check this. Therefore, get the `studyId` of the group and check for the claim serves another purpose for the caller. 

It is not a carp.core concern here, just a matter if we want to point this out.

Whathecode: >  it's up to the implementations to check this.

Claims so far have always been an infrastructure concern. These columns are simply suggestions.

But, this does mean that an additional lookup, other than the information in the request (e.g., repo or backend cache), is needed to implement claim-based authorization here, since without that `studyId` is unknown.

Whathecode:
@yuanchen233 Full review finally done. :) I made minor changes, which you can see in [my force-pushed changes](https://github.com/carp-dk/carp.core-kotlin/compare/bd1d21ed1a9de42d2773e1ce8f54e6b9257d73f8..df75a5924bcfbd0eae34f8169bf8bb47337a2dd5). Primarily:
- Updated the README code sample based on the readme() code you updated. That is currently a copy/paste still. The actual code is there to ensure the code in the README code compiles. :)
- Spacing here and there.
- More focused tests of `createParticipantGroup` and `getParticipantGroupStatusList`, rather than calling the latter in the former.
- Removed `getParticipantGroupStatusList_returns_staged_groups_as_staged` and instead return/test both deployed and staged group in `getParticipantGroupStatusList_returns_multiple_deployments`.
- Removed TypeScript `can serialize DeployParticipantGroup` test. I think the previous serialization test was there for the more complex `InviteNewParticipantGroup` parameters. So, your update to `CreateParticipantGroup` and testing that was sufficient. 👍

Whathecode (approved):

PR #524, opened 2026-01-16
Add UpdateParticipantGroup endpoint in RecruitmentService

### LINKED ISSUE #319: Endpoints to edit staged participant groups

yuanchen233:
This PR adds support for updating participant groups in RecruitmentService.

### DIFF
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
@@ -109,6 +109,25 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
     ): ParticipantGroupStatus
 
     /**
+     * Update the participant group for the specified [groupId].
+     *
+     * Participant assignments can't be changed after the group has been invited; the group name can always be updated.
+     *
+     * @param group If set, role assignments are updated; unchanged otherwise.
+     * @param name If set, the group name is updated; unchanged otherwise.
+     * @throws IllegalArgumentException when:
+     *  - the participant group with [groupId] does not exist
+     *  - any of the participant roles specified in [group] does not exist
+     *  - any of the participants specified in [group] does not exist
+     * @throws IllegalStateException when the group has already been deployed and [group] changes participant assignments.
+     */
+    suspend fun updateParticipantGroup(
+        groupId: UUID,
+        group: Set<AssignedParticipantRoles>? = null,
+        name: String? = null
+    ): ParticipantGroupStatus
+
+    /**
      * Invite the participant group with the specified [groupId] to start participating in its study.
      *
      * @throws IllegalArgumentException when:
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
@@ -200,6 +200,39 @@ class RecruitmentServiceHost(
     }
 
     /**
+     * Update the participant group for the specified [groupId].
+     *
+     * Participant assignments can't be changed after the group has been invited; the group name can always be updated.
+     * If [group] is provided, role assignments are updated; unchanged otherwise.
+     * If [name] is provided, the group name is updated; unchanged otherwise.
+     *
+     * @throws IllegalArgumentException when:
+     *  - the participant group with [groupId] does not exist
+     *  - any of the participant roles specified in [group] does not exist
+     *  - any of the participants specified in [group] does not exist
+     * @throws IllegalStateException when the group has already been deployed and [group] changes participant assignments.
+     */
+    override suspend fun updateParticipantGroup(
+        groupId: UUID,
+        group: Set<AssignedParticipantRoles>?,
+        name: String?
+    ): ParticipantGroupStatus
+    {
+        val (recruitment, participantGroup) = getRecruitmentWithGroupOrThrow( groupId )
+
+        if ( group != null || name != null )
+        {
+            recruitment.updateParticipantGroup( groupId, group, name )
+            participantRepository.updateRecruitment( recruitment )
+        }
+
+        return recruitment.getParticipantGroupStatus(
+            participantGroup.id,
+            deploymentService::getStudyDeploymentStatus
+        )
+    }
+
+    /**
      * Invite the participant group with the specified [groupId] to start participating in its study.
      *
      * @throws IllegalArgumentException when:
--- a/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
+++ b/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
@@ -24,6 +24,7 @@ import kotlinx.datetime.Instant
 /**
  * Represents a set of [participants] recruited for a study identified by [studyId].
  */
+@Suppress( "TooManyFunctions" )
 class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: Instant = Clock.System.now() ) :
     AggregateRoot<Recruitment, RecruitmentSnapshot, Recruitment.Event>( id, createdOn )
 {
@@ -206,6 +207,42 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
     }
 
     /**
+     * Update the [participants] and/or [name] of an existing participant group with the specified [groupId].
+     *
+     * @throws IllegalArgumentException when:
+     *  - the participant group with [groupId] does not exist.
+     *  - one or more of the participants aren't in this recruitment.
+     *  - any of the participant roles specified in [participants] are not part of the configured study protocol.
+     * @throws IllegalStateException when:
+     *  - the recruitment is not ready for deployment
+     *  - the group has already been deployed and [participants] changes assignments
+     */
+    fun updateParticipantGroup(
+        groupId: UUID,
+        participants: Set<AssignedParticipantRoles>? = null,
+        name: String? = null
+    )
+    {
+        val group = requireNotNull( _participantGroups[ groupId ] )
+            { "Participant group with ID \"$groupId\" does not exist." }
+
+        val newRoleAssignments = participants != null && group.roleAssignments != participants
+        if ( newRoleAssignments )
+        {
+            check( !group.isDeployed )
+                { "Participant group has already been deployed; participant assignments can no longer be changed." }
+
+            val status = getStatus()
+            check( status is RecruitmentStatus.ReadyForDeployment ) { "The study is not yet ready for deployment." }
+
+            validateRoleAssignments( status.studyProtocol, participants )
+            group.replaceParticipants( participants )
+        }
+
+        if ( name != null ) group.name = name
+    }
+
+    /**
      * Get the [ParticipantGroupStatus] for participant groups with [groupIds] in this recruitment.
      * For deployed participant groups, the matching [StudyDeploymentStatus] is retrieved using
      * [getDeploymentStatusList].
@@ -231,7 +268,7 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
             ParticipantGroupStatus.Staged(
                 id = group.id,
                 participants = getParticipantsFor( group ),
-                assignedParticipantRoles = group.roleAssignments,
+                assignedParticipantRoles = group.roleAssignments.toSet(),
                 name = group.name
             )
         }
@@ -248,7 +285,7 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
                 { "No study deployment status returned for the requested ID: \"${group.id}\"." }
             ParticipantGroupStatus.InDeployment.fromDeploymentStatus(
                 getParticipantsFor( group ),
-                group.roleAssignments,
+                group.roleAssignments.toSet(),
                 deploymentStatus,
                 group.name
             )
@@ -268,7 +305,7 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
      */
     suspend fun getParticipantGroupStatus(
         groupId: UUID,
-        getDeploymentStatus: (UUID) -> StudyDeploymentStatus
+        getDeploymentStatus: suspend (UUID) -> StudyDeploymentStatus
     ): ParticipantGroupStatus =
         getParticipantGroupStatusList( setOf( groupId ) )
             { deploymentIds -> listOf( getDeploymentStatus( deploymentIds.single() ) ) }.single()
--- a/carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceTest.kt
+++ b/carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceTest.kt
@@ -375,6 +375,163 @@ interface RecruitmentServiceTest
     }
 
     @Test
+    fun updateParticipantGroup_succeeds_before_invite() = runTest {
+        val (recruitmentService, studyService) = createSUT()
+        val (studyId, _) = createLiveStudy( studyService )
+        val participant = recruitmentService.addParticipant( studyId, EmailAddress( "test@test.com" ) )
+        val groupId = UUID.randomUUID()
+        val initialAssignments = setOf( AssignedParticipantRoles( participant.id, AssignedTo.All ) )
+        recruitmentService.createParticipantGroup( groupId, initialAssignments, studyId, "Initial name" )
+
+        val updatedParticipant = recruitmentService.addParticipant( studyId, EmailAddress( "test2@test.com" ) )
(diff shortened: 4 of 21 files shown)


README.md
```diff
@@ -250,7 +250,11 @@ if ( studyStatus.canDeployToParticipants )
     var groupStatus: ParticipantGroupStatus =
         recruitmentService.createParticipantGroup( groupId, participantGroup, studyId )
     val isStaged = groupStatus is ParticipantGroupStatus.Staged // True.
-  
+
+    val secondParticipant = recruitmentService.addParticipant( studyId, EmailAddress( "participant2@email.com" ) )
```
Whathecode: This seems to be added to document the update endpoint.

I suggest keeping the README's simple, focused on the main "happy path". It doesn't need to cover all possible endpoints.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
```diff
@@ -108,6 +108,22 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
         name: String? = null
     ): ParticipantGroupStatus
 
+    /**
+     * Update the participant [group] with the specified [groupId], optionally updating assignments and [name].
```
Whathecode: The way this is currently formulated, it sounds like there is something more to be updated than assignments and name. I'd drop the "optionally" part, and keep that for argument documentation: "If set, ...; unchanged otherwise."

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
```diff
@@ -108,6 +108,22 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
...
+     * Participant assignments can only be updated before the group is invited; updating [name] is always allowed.
```
Whathecode: Grammar/clarity: "Participant assignments can't be changed after the group has been invited. Updating the group name is always possible."

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -191,14 +199,45 @@ class RecruitmentServiceHost(
...
+    {
+        val (recruitment, participantGroup) = getRecruitmentWithGroupOrThrow( groupId )
+
+        if ( group != null || name != null )
+        {
+            recruitment.participantGroupService.updateParticipantGroup( participantGroup, group, name )
+            participantRepository.updateRecruitment( recruitment )
+        }
+
+        // TODO: If status lookup fails after updateRecruitment succeeds, state will be inconsistent.
```
Whathecode: I don't think it would be inconsistent. The get status call doesn't change any state in the other service. At most, it would be odd this request failed (and the status isn't returned) but the requested action still happened.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -205,6 +206,44 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
...
+     *  - any of the participant roles specified in [participants] are not part of the configured study protocol.
+     * @throws IllegalStateException when:
+     *  - the recruitment is not ready for deployment
+     *  - the group has already been deployed and [participants] changes assignments
+     */
+    fun updateParticipantGroup(
+        group: StagedParticipantGroup,
+        participants: Set<AssignedParticipantRoles>? = null,
+        name: String? = null
+    )
```
Whathecode, line 224: This doesn't make sense. Either _only_ provide `group` as an argument and use it to carry the entire desired new state. Or, use `id`, `participants`, and `name` as arguments.

Given that `StagedParticipantGroup` is mutable (not a value object), the latter is the best option. Also, `getParticipantGroupStatusList` uses IDs.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -205,6 +206,44 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
...
+        if ( participants != null )
+        {
+            val status = getStatus()
+            check( status is RecruitmentStatus.ReadyForDeployment ) { "The study is not yet ready for deployment." }
+
+            validateRoleAssignments( status.studyProtocol, participants )
+            group.replaceParticipants( participants )
+        }
+        if ( name != null ) group.name = name
+    }
```
Whathecode: I'd restructure this. As is always recommended: first do early outs/argument checking. I.e.:

```
if ( participants != null )
{
  check( !group.isDeployed || group.roleAssignments == participants ) ...
  check( status is RecruitmentStatus.ReadyForDeployment ) ...
}
```

Then just set name if != null, and participants if != null.

Whathecode (changes requested):
I'm still reviewing. But, given I'm short of time, better you already get started on these comments. :)

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -231,7 +270,7 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
             ParticipantGroupStatus.Staged(
                 id = group.id,
                 participants = getParticipantsFor( group ),
-                assignedParticipantRoles = group.roleAssignments,
+                assignedParticipantRoles = group.roleAssignments.toSet(),
```
Whathecode, line 271: Good catch! I was wondering why this was here. This is almost a bug fix, but it only became a problem now that things get mutable. This warrants a bit more description in the commit message:
- That would have saved me time reviewing this/figuring out why this changed.
- For posterity, if this commit ever surfaces when investigating the commit history.

README.md
Whathecode: Same comment for `StudiesCodeSamples`

docs/carp-studies.md
```diff
@@ -57,6 +57,7 @@ Allows setting recruitment goals, adding participants to studies, and creating d
 | `getParticipants` | Get all participants for a study. | manage study: `studyId` | |
 | `inviteNewParticipantGroup` | (Deprecated) Create and instantly invite a group of previously added participants to a study. | manage study: `studyId` | |
 | `createParticipantGroup` | Prepare a group of previously added participants for deployment without inviting them yet. | manage study: `studyId` | |
+| `updateParticipantGroup` | Update a prepared participant group before it is invited; group names can be updated anytime. | manage study: `studyId` | |
```
Whathecode: You can drop details, like constraints on what can be updated/when, from this high-level documentation (otherwise you have to update that here each time that changes). The goal here is just to get a high-level overview of available functionality, and suggested access restrictions.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceTest.kt
```diff
@@ -374,6 +374,143 @@ interface RecruitmentServiceTest
...
     }
 
+    @Test
+    fun updateParticipantGroup_succeeds_before_invite() = runTest {
+        val (recruitmentService, studyService) = createSUT()
+        val (studyId, protocol) = createLiveStudy( studyService )
+        val participant1 = recruitmentService.addParticipant( studyId, EmailAddress( "test@test.com" ) )
+        val participant2 = recruitmentService.addParticipant( studyId, EmailAddress( "test2@test.com" ) )
+        val role1 = protocol.participantRoles.first().role
+        val role2 = protocol.participantRoles.last().role
```
Whathecode, line 406: For this test, there is no need to assign both roles (which relies on checking `createLiveStudy` further down to understand there are more roles.) I would simplify this and use `AssignedTo.All` (as also used in many other tests) and just swap out participant 1 with participant 2.

It decouples this test a bit more from implementation details in `createLiveStudy`.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceTest.kt
```diff
@@ -374,6 +374,143 @@ interface RecruitmentServiceTest
...
+        val updatedAssignments = setOf(
+            AssignedParticipantRoles( participant2.id, AssignedTo.Roles( setOf( role2 ) ) )
+        )
+        val updatedStatus =
+            recruitmentService.updateParticipantGroup( groupId, updatedAssignments, "Updated name" )
+
+        assertTrue( updatedStatus is ParticipantGroupStatus.Staged )
+        assertEquals( "Updated name", updatedStatus.name )
+        assertEquals( setOf( participant2 ), updatedStatus.participants )
+        assertEquals( updatedAssignments, updatedStatus.assignedParticipantRoles )
```
Whathecode, line 421: I suspect you can simply do a full object comparison of an `expectedStatus` instead, since all of these are `data` classes.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceTest.kt
```diff
@@ -374,6 +374,143 @@ interface RecruitmentServiceTest
...
+        assertEquals( "Renamed after invite", updatedStatus.name )
+        assertTrue( updatedStatus is ParticipantGroupStatus.Invited )
+    }
+
+    @Test
+    fun updateParticipantGroup_fails_when_updating_participants_after_invite() = runTest {
+        val (recruitmentService, studyService) = createSUT()
+        val (studyId, _) = createLiveStudy( studyService )
+        val participant1 = recruitmentService.addParticipant( studyId, EmailAddress( "test@test.com" ) )
+        val participant2 = recruitmentService.addParticipant( studyId, EmailAddress( "test2@test.com" ) )
```
Whathecode: Move closer to where it is used (the `updatedAssignments`). And, naming-wise you could call this `updatedParticipant` instead of `participant1` and `participant2`.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceTest.kt
```diff
@@ -374,6 +374,143 @@ interface RecruitmentServiceTest
...
+        val assignments = setOf( AssignedParticipantRoles( participant1.id, AssignedTo.All ) )
+        val groupStatus = recruitmentService.createParticipantGroup( UUID.randomUUID(), assignments, studyId )
+        recruitmentService.inviteParticipantGroup( groupStatus.id )
+
+        val updatedAssignments = setOf( AssignedParticipantRoles( participant2.id, AssignedTo.All ) )
+        assertFailsWith<IllegalStateException> {
+            recruitmentService.updateParticipantGroup( groupStatus.id, updatedAssignments, "New name" )
```
Whathecode: Don't pass a "New name" here. It's unrelated to this test, which focuses on updating participants.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceTest.kt
```diff
@@ -374,6 +374,143 @@ interface RecruitmentServiceTest
...
+        val updatedAssignments = setOf(
+            AssignedParticipantRoles( participant.id, AssignedTo.Roles( setOf( "Missing role" ) ) )
+        )
+        assertFailsWith<IllegalArgumentException> {
+            recruitmentService.updateParticipantGroup( groupStatus.id, updatedAssignments )
+        }
+    }
+
+    @Test
+    fun updateParticipantGroup_succeeds_with_assignments_only_before_invite() = runTest {
```
Whathecode: Isn't this the same as `updateParticipantGroup_succeeds_before_invite`? Remove if so.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceTest.kt
```diff
@@ -374,6 +374,143 @@ interface RecruitmentServiceTest
...
+        val (recruitmentService, studyService) = createSUT()
+        val (studyId, _) = createLiveStudy( studyService )
+        val participant = recruitmentService.addParticipant( studyId, EmailAddress( "test@test.com" ) )
+        val assignments = setOf( AssignedParticipantRoles( participant.id, AssignedTo.All ) )
+        val groupStatus =
+            recruitmentService.createParticipantGroup( UUID.randomUUID(), assignments, studyId, "Initial name" )
+        recruitmentService.inviteParticipantGroup( groupStatus.id )
+
+        val updatedStatus =
+            recruitmentService.updateParticipantGroup( groupStatus.id, assignments, "Renamed after invite" )
```
Whathecode: "Renamed after invite" is not part of this test. Probably better to pass `null`, with the added benefit you get coverage for that which you actually currently don't have!

Whathecode (changes requested):
I finished reviewing the first commit. I notice missing test coverage for the methods which got added to `Recruitment`. That _may_ be fine given your coverage in `RecruitmentService`. Normally, I would emphasize testing domain classes more. But, given that service tests in this repository add backwards compatibility tests for free, I guess they are actually preferred. :)

And, without having looked at the second commit, since you factor stuff out there, maybe you just add tests for the extracted part at that point, which would also be a good reason not to add it here to instantly move it in the next commit.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
yuanchen233: Since `Recruitment.updateParticipantGroup` now takes `groupId` instead of a group object, I ended up checking existence twice (service `getRecruitmentWithGroupOrThrow` + domain `requireNotNull`). Domain check feels right to preserve invariants if this method is called elsewhere, but the service-side check also makes sense — is this redundancy okay?

yuanchen233:
Thanks for the review and apologies for the delay.

I went back and forth on updateParticipantGroup design: 

Should we enforce group = null when no assignment change after invite, or allow passing the same assignments for idempotent calls? Allowing equality keeps the endpoint forgiving and avoids forcing callers to fetch/compare current assignments.

Also: should we allow a no-op call when `group` and `name` are null, or should we simply reject it?

---
Side note: I think I messed up with reset/rebase so the test cases were a bit off, and I was only focusing on the logic part when I reviewing my own code here. Do you have suggestions on how to do self review, or code reviews in general? I tend to focus on the part I spent most time on, and ignoring details..

yuanchen233:
After reconsideration, I removed the second commit from this PR:
- The extraction did not establish a meaningful domain-service boundary; it mostly functioned as a helper around Recruitment.
- It introduced extra indirection without reducing conceptual complexity.
- It also changed the public usage pattern, which could break consumers that use Recruitment directly in applications.
- With [#525](https://github.com/imotions/carp.core-kotlin/issues/525) now in place, we can continue from there with a clearer direction.
- 
That said, the original concern remains valid:
- Recruitment still has too many functions (as flagged by Detekt). 
- We have quite a few participant-group level operations now, they might deserve their own class. Currently, most of those operations mutate Recruitment, so extracting them cleanly requires a better boundary design than this iteration provided.

Whathecode:
> With https://github.com/imotions/carp.core-kotlin/issues/525 now in place, we can continue from there with a clearer direction.
> 
> We have quite a few participant-group level operations now, they might deserve their own class. 

I think we identified a design flaw, which is also a scalability flaw, at the same time for different reasons. :) It's good to see they are pointing in the same direction.

I'll review your changes/provide input later this week.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
Whathecode: I meant using `@param` docs. :) I updated it.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
Whathecode: In the service we no longer use the returned participant group (only to access the ID we already were given), so it feels redundant. Maybe by changing `getRecruitmentWithGroupOrThrow`. I'll have a try.

Whathecode: Nah ... I think the redundancy is fine for now. It's cheap either way, and protects the same invariant at a different level.

Whathecode: I updated the logic a bit for readability, though.

Whathecode (approved):
I made some minor stylistic changes. The main change was slightly throwing logic around in `Recruitment.updateParticipantGroup()`.

Great work!

yuanchen233:
I get sidetracked [Update branch with rebase repository setting](https://github.com/orgs/community/discussions/12032) before click on the update branch with rebase, was expecting Github to have that setting.

PR #526, opened 2026-02-15
Build: fix README status badges

Whathecode:
These have been broken for a while, ever since Sonatype switched their endpoints. This should make the readme look a bit prettier again. And, the links now point to the new location to find artifacts.

image


docs/carp-protocols.md
```diff
@@ -1,4 +1,4 @@
-# carp.protocols [![Maven Central](https://maven-badges.herokuapp.com/maven-central/dk.cachet.carp.protocols/carp.protocols.core/badge.svg)](https://mvnrepository.com/artifact/dk.cachet.carp.protocols) [![Sonatype Nexus (Snapshots)](https://img.shields.io/nexus/s/dk.cachet.carp.protocols/carp.protocols.core?server=https%3A%2F%2Foss.sonatype.org)](https://oss.sonatype.org/content/repositories/snapshots/dk/cachet/carp/protocols/) 
+# carp.protocols [![Maven Central](https://img.shields.io/maven-central/v/dk.cachet.carp.protocols/carp.protocols.core)](https://central.sonatype.com/artifact/dk.cachet.carp.protocols/carp.protocols.core) [![Snapshot)](https://img.shields.io/maven-metadata/v?metadataUrl=https%3A%2F%2Fcentral.sonatype.com%2Frepository%2Fmaven-snapshots%2Fdk%2Fcachet%2Fcarp%2Fprotocols%2Fcarp.protocols.core%2Fmaven-metadata.xml)](https://central.sonatype.com/repository/maven-snapshots/dk/cachet/carp/protocols/carp.protocols.core/maven-metadata.xml)
```
yuanchen233: This is the only subsystem image alt text is changed, typo?

Whathecode: Good catch! Oversight on all the other ones. :) Thanks!

yuanchen233 (dismissed):
I skimmed through this document:
1. Is the link for snapshots under [Usage section](https://github.com/carp-dk/carp.core-kotlin?tab=readme-ov-file#usage) still available? As we are only providing XML files for snapshots now.
2. `iMotions Mobile Research Platform` link broken
3. `Link to github orgnization` is broken due to rename
4. Recommanded IDEA version
...
Also the UML diagrams for new endpoints.

Just sketched some notes here as I'm not available at the moment, we should update docs before version 1.3 release (hopefully soon 😃 )./

Whathecode:
> Is the link for snapshots under Usage section still available?

No. That's the wrong link and should have been changed back when switching to new sonatype endpoints. I actually tried to look for it yesterday as I recalled something like that was documented. But, I couldn't find it immediately and it was late. 😅

That needs to be fixed in this commit. The other stuff we can do later.

Whathecode:
> That needs to be fixed in this commit. The other stuff we can do later.

@yuanchen233  It was easy to fix all. So all should be good now. Thank you for spotting it!

yuanchen233 (approved):
:shipit:

PR #527, opened 2026-02-19
Add GetDataStreamsStatus endpoints to DataStreamService

### LINKED ISSUE #480: Need an endpoint for getting the latest sequence ID
Right now sequence IDs are enforced - see #375 .

If this is the case, we need a way to get the latest sequence ID for a data stream, so that we can continue from there (and avoid the error).

yuanchen233:
Base on real-life use cases, study normally requires different type of data, therefore I also added `GetLastSequenceIds` which gives all 'lastSequenceId' for a deployment.
(or maybe `GetLastSequenceId` is not needed, only `GetLastSequenceIds` is sufficient?)

I did not add this to `StudyDeploymentService`:
- lastSequenceId is per DataStreamId (deployment + device + dataType), so payload can get large.
- DeploymentServiceHost would need cross-service aggregation from data on every status call, which can add latency and coupling.
- By the use case given in the original issue, the 'lastSequenceId' is needed when the client device loose track of data ( app re-install ), therefore it is not needed with every deployment status call. Though 'lastSequenceId' can represent data collection status for that deployment.

---
Updated:

This PR adds`GetDataStreamsStatus` endpoint for retrieving status information for a study deployment. Currently it contains all `lastSequenceId`s for each `dataStream`, and whether if the `dataSteam` is closed.

### DIFF
--- a/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/InMemoryDataStreamService.kt
+++ b/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/InMemoryDataStreamService.kt
@@ -7,6 +7,7 @@ import dk.cachet.carp.common.domain.ExtractUniqueKeyMap
 import dk.cachet.carp.data.application.DataStreamBatch
 import dk.cachet.carp.data.application.DataStreamId
 import dk.cachet.carp.data.application.DataStreamService
+import dk.cachet.carp.data.application.DataStreamStatus
 import dk.cachet.carp.data.application.DataStreamsConfiguration
 import dk.cachet.carp.data.application.MutableDataStreamBatch
 import dk.cachet.carp.data.application.MutableDataStreamSequence
@@ -27,6 +28,7 @@ class InMemoryDataStreamService : DataStreamService
         }
     private val stoppedStudyDeploymentIds: MutableSet<UUID> = mutableSetOf()
     private val dataStreams: MutableDataStreamBatch = MutableDataStreamBatch()
+    private val lastSequenceIdByDataStream: MutableMap<DataStreamId, Long?> = mutableMapOf()
 
 
     /**
@@ -37,6 +39,9 @@ class InMemoryDataStreamService : DataStreamService
     override suspend fun openDataStreams( configuration: DataStreamsConfiguration )
     {
         configuredDataStreams.tryAddIfKeyIsNew( configuration )
+        configuration.expectedDataStreamIds.forEach { dataStreamId ->
+            lastSequenceIdByDataStream[ dataStreamId ] = null
+        }
     }
 
     /**
@@ -63,6 +68,13 @@ class InMemoryDataStreamService : DataStreamService
             { "Data streams for this study deployment have been closed." }
 
         dataStreams.appendBatch( batch )
+
+        // Update last sequence ID for all streams with new data.
+        batch.sequences
+            .groupBy { it.dataStream }
+            .forEach { (dataStreamId, dataStreamSequences) ->
+                lastSequenceIdByDataStream[ dataStreamId ] = dataStreamSequences.last().range.last
+            }
     }
 
     /**
@@ -112,6 +124,23 @@ class InMemoryDataStreamService : DataStreamService
     }
 
     /**
+     * Retrieve status for each configured data stream in [studyDeploymentId].
+     *
+     * @throws IllegalArgumentException when no data streams were ever opened for [studyDeploymentId].
+     */
+    override suspend fun getDataStreamsStatus( studyDeploymentId: UUID ): List<DataStreamStatus>
+    {
+        val configuration: DataStreamsConfiguration? = configuredDataStreams[ studyDeploymentId ]
+        requireNotNull( configuration ) { "No data streams configured for this study deployment." }
+
+        val isOpen = studyDeploymentId !in stoppedStudyDeploymentIds
+        return configuration.expectedDataStreamIds.map { dataStream ->
+            val lastSequenceId: Long? = lastSequenceIdByDataStream[ dataStream ]
+            DataStreamStatus( dataStream, lastSequenceId, isOpen )
+        }
+    }
+
+    /**
      * Stop accepting incoming data for all data streams for each of the [studyDeploymentIds].
      *
      * @throws IllegalArgumentException when no data streams were ever opened for any of the [studyDeploymentIds].
@@ -134,9 +163,15 @@ class InMemoryDataStreamService : DataStreamService
     {
         stoppedStudyDeploymentIds.removeAll( studyDeploymentIds )
 
-        return studyDeploymentIds.mapNotNull { toRemove ->
+        val removedStudyDeploymentIds = studyDeploymentIds.mapNotNull { toRemove ->
             if ( configuredDataStreams.removeKey( toRemove ) ) toRemove
             else null
         }.toSet()
+
+        lastSequenceIdByDataStream.keys.removeAll { dataStreamId ->
+            dataStreamId.studyDeploymentId in removedStudyDeploymentIds
+        }
+
+        return removedStudyDeploymentIds
     }
 }
--- a/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamBatch.kt
+++ b/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamBatch.kt
@@ -102,11 +102,11 @@ class MutableDataStreamBatch : DataStreamBatch
                 // This might seem like premature optimization, but currently it is the only concrete class.
                 // We expect many sequences for one data type to be common, e.g., RR intervals have many sync points.
                 batch.sequenceMap
-                    .mapValues { it.value.last() }
-                    .all { (dataStream, lastSequence) ->
+                    .mapValues { it.value.first() }
+                    .all { (dataStream, firstNewSequence) ->
                         val lastStoredSequence = sequenceMap[ dataStream ]?.last()
                         if ( lastStoredSequence == null ) true
-                        else lastStoredSequence.range.last < lastSequence.range.first
+                        else lastStoredSequence.range.last < firstNewSequence.range.first
                     }
             else ->
                 batch.sequences.all { sequence ->
--- a/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamService.kt
+++ b/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamService.kt
@@ -61,6 +61,13 @@ interface DataStreamService : ApplicationService<DataStreamService, DataStreamSe
     ): DataStreamBatch
 
     /**
+     * Retrieve status for each configured data stream in [studyDeploymentId].
+     *
+     * @throws IllegalArgumentException when no data streams were ever opened for [studyDeploymentId].
+     */
+    suspend fun getDataStreamsStatus( studyDeploymentId: UUID ): List<DataStreamStatus>
+
+    /**
      * Stop accepting incoming data for all data streams for each of the [studyDeploymentIds].
      *
      * @throws IllegalArgumentException when no data streams were ever opened for any of the [studyDeploymentIds].
--- a/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamStatus.kt
+++ b/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamStatus.kt
@@ -1,0 +1,25 @@
+@file:Suppress( "NON_EXPORTABLE_TYPE" )
+
+package dk.cachet.carp.data.application
+
+import kotlinx.serialization.Serializable
+import kotlin.js.JsExport
+
+
+/**
+ * Status of a [dataStream] in a study deployment.
+ */
+@Serializable
+@JsExport
+data class DataStreamStatus(
+    val dataStream: DataStreamId,
+    /**
+     * The sequence number of the last received measurement for this [dataStream],
+     * or `null` when no data has been received yet.
+     */
+    val lastSequenceId: Long?,
+    /**
+     * Determines whether this [dataStream] currently accepts incoming data.
+     */
+    val isOpen: Boolean
+)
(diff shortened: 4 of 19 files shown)


carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/InMemoryDataStreamService.kt
```diff
@@ -139,4 +172,9 @@ class InMemoryDataStreamService : DataStreamService
             else null
         }.toSet()
     }
+
+    private fun getLastSequenceIdOrNull( dataStream: DataStreamId ): Long? =
+        dataStreams.sequences
+            .filter { it.dataStream == dataStream }
+            .maxOfOrNull { it.range.last }
```
Whathecode: This is something which clearly should be cached/indexed by the data subsystem, even for the in memory implementation.

Whathecode:
I think the long-term idea was that the deployment subsystem provides data subsystem endpoint information as part of device deployment. This would enable horizontal scaling of the data subsystem which can be expected to need a lot more throughput than the deployment subsystem. Whether or not that needs CARP Core involvement is not immediately clear to me; it does feel like an infrastructure concern.

But, this general notion of the deployment subsystem as a mediator is why my intuition was [this information should live in the deployment status](https://github.com/carp-dk/carp.core-kotlin/issues/480#issuecomment-2184034996). Of course, once the data endpoint is known, that could be queried for `lastSequenceId`, so I think your points still hold. The client subsystem should contact the data subsystem directly either way.

> By the use case given in the original issue, the 'lastSequenceId' is needed when the client device loose track of data ( app re-install ), therefore it is not needed with every deployment status call.

It could be that it would only be needed for the `NeedsRedeployment` status. But, I agree with your assessment here. Let's keep it simple and let the client subsystem contact the data subsystem directly for this information.

> (or maybe GetLastSequenceId is not needed, only GetLastSequenceIds is sufficient?)

Yes. There is no use case to get it for a single sequence. But, I would go further. The general use case would be to receive the status of data stream sequences for a given `studyDeploymentId`. This is more future proof. We may also want to add total number of data points, size, etc., over time as the need arises. There will be no benefit to having each of these as separate endpoints as far as I can tell.

Whathecode (changes requested):
Very good observation on why you are adding this to the data subsystem only.👍

However, see my comment on the PR. I think we should create a more general `getDataStreamsStatus( studyDeploymentId: UUID )` endpoint which can include more information over time. Right now, it can already also include whether the streams are open or closed.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/InMemoryDataStreamService.kt
yuanchen233: I added `lastSequenceIdByDataStream` in `InMemoryDataStreamService` to caches `lastSequenceId` per `DataStreamId`.

My concern is, batch append is owned by `MutableDataStreamBatch` and not guaranteed atomic. So I choose to recomputes affected index entries from stored sequences after append. This keeps the index correct even on partial append failures. 

I didn't do much verifying as this is in memory implementation, but I think this can be a note for people implementing the infrastructures ( likely future me, at least )

Whathecode: > My concern is, batch append is owned by `MutableDataStreamBatch` and not guaranteed atomic. 

When would we have a partial success here? It seems like there is at least an attempt made to make this atomic. [The precondition check is done in its entirety before appending anything](https://github.com/carp-dk/carp.core-kotlin/blob/develop/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamBatch.kt#L98).

If there is anything missing, I would fix that to make it atomic (but thread safe is not needed).

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/InMemoryDataStreamService.kt
```diff
@@ -62,7 +66,19 @@ class InMemoryDataStreamService : DataStreamService
...
+        {
+            dataStreams.appendBatch( batch )
+        }
+        finally
+        {
+            affectedDataStreamIds.forEach { dataStreamId ->
+                lastSequenceIdByDataStream[ dataStreamId ] = dataStreams.sequences
+                    .filter { sequence -> sequence.dataStream == dataStreamId }
+                    .maxOfOrNull { sequence -> sequence.range.last }
+            }
```
Whathecode: This complexity seems unnecessary.

Looking at `MutableDataStreamBatch`, it seems like the contract for `appendBatch` is [that it should be atomic](https://github.com/carp-dk/carp.core-kotlin/pull/527#discussion_r2907188765). And, the contract for `DataStreamBatch` is for it to be a "collection of non-overlapping, ordered, data stream [sequences]". So, you should just be able to take the last sequence per ID. That ordering is e.g. also relied upon by `DataStreamBatch.getDataStreamPoints`.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/InMemoryDataStreamService.kt
```diff
@@ -134,9 +150,15 @@ class InMemoryDataStreamService : DataStreamService
...
 
-        return studyDeploymentIds.mapNotNull { toRemove ->
+        val removedStudyDeploymentIds = studyDeploymentIds.mapNotNull { toRemove ->
             if ( configuredDataStreams.removeKey( toRemove ) ) toRemove
             else null
         }.toSet()
+
+        lastSequenceIdByDataStream.keys.removeAll { dataStreamId ->
+            dataStreamId.studyDeploymentId in removedStudyDeploymentIds
+        }
```
Whathecode, line 173: Can't you just use `removeAll()` and [pass a collection](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/remove-all.html) here? 🤔

Whathecode (changes requested):
Only reviewed first commit so far. As is, that only introduces private behavior. I suggest you include `getDataStreamsStatus` on `DataStreamService` already as part of the first commit (and that's also the focus of the commit message).

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/InMemoryDataStreamService.kt
yuanchen233: Taking a closer look, I think there are two cases:

1. `syncPoint` is validated during per-sequence append, so an older `syncPoint` can fail after earlier sequences were already appended.

2. For `MutableDataStreamBatch` pre-check, we only compare the stored tail against the [_last_](https://github.com/carp-dk/carp.core-kotlin/blob/98638ca1070474764f23295a7bc80eb3cdf72190/carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamBatch.kt#L105) incoming sequence for each stream.  
   That means earlier incoming sequences for the same `dataStreamId` can still violate range ordering and fail later in per-sequence append.  
   If my understanding is correct, this is a bug, and we should compare against the _first_ incoming sequence instead.

3. In `appendToDataStreams` endpoint, the batch comes from the caller, so we should treat it as untrusted input (sequence order, sync-time order, etc.)

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/InMemoryDataStreamService.kt
yuanchen233: `lastSequenceIdByDataStream` is `<MutableMap<DataStreamId, Long?>`, I can build the collection from configuration first like:

```
val dataStreamIdsToRemove = studyDeploymentIds
    .mapNotNull { studyDeploymentId -> configuredDataStreams[ studyDeploymentId ] }
    .flatMap { configuration -> configuration.expectedDataStreamIds }
    .toSet()

lastSequenceIdByDataStream.keys.removeAll( dataStreamIdsToRemove )
```
This is cleaner, is this what you are referring to?>

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/InMemoryDataStreamService.kt
yuanchen233: Then yes, if `appendBatch` is atomic, that's easier and makes more sense.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/InMemoryDataStreamService.kt
Whathecode: Point 2 relies on the fact that `MutableDataStreamBatch` already protects that invariant in `appendSequence`. Unless I'm overlooking something, you can't instantiate a `MutableDataStreamBatch` which violates that constraint. Hence, you can assume checking just the last sequence is fine.

Point 1: the `syncPoint` invariant is indeed less clear, but should arguably also be part of the `DataStreamBatch` contract.

But, to point 3: maybe there is no point in `DataStreamBatch` being open and we should just make it `sealed`. It would also get rid of the fallback `else` in `appendBatch` that currently exists there. I can't immediately see why we would have a requirement to leave `DataStreamBatch` open. I think it had more to do with having a read-only interface than anything else.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/InMemoryDataStreamService.kt
Whathecode: Nah ... I think the original is better. I didn't immediately see the impact of `DataStreamId` vs just study deployment IDs here.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/InMemoryDataStreamService.kt
Whathecode: So, [if this makes sense](https://github.com/carp-dk/carp.core-kotlin/pull/527#discussion_r2936003032), let's simplify this.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/InMemoryDataStreamService.kt
yuanchen233: Yes, I agree that things should be more strict for point 1 and 3.

> Point 2 relies on the fact that `MutableDataStreamBatch` already protects that invariant in `appendSequence`. Unless I'm overlooking something, you can't instantiate a `MutableDataStreamBatch` which violates that constraint. Hence, you can assume checking just the last sequence is fine.

Thanks, this helped clarify things for me.  
From what I can see:

1. `appendSequence` only protects the incoming batch while building it, not incoming vs already stored sequences.

2. If the error happens during per-sequence append in `appendBatch`, then it’s outside pre-check, so partial success can still happen.

3. This case shows up when there are multiple sequences for the same `dataStreamId` in the incoming batch and they are not `isImmediatelyFollowedBy` each other.

I added a [test case](https://github.com/carp-dk/carp.core-kotlin/pull/527/changes/df0346518251fcd59abc9c37e45fcdc193129a6b) to show this. Let me know if this looks right, or if I’m still off track.

Whathecode: It seems like you found a bug, yes. The intended logic should still work, but is implemented incorrectly. Looks like it should be `mapValues { it.value.first() }` instead of `last`:
```
                batch.sequenceMap
                    .mapValues { it.value.first() }
                    .all { (dataStream, firstNewSequence) ->
                        val lastStoredSequence = sequenceMap[ dataStream ]?.last()
                        if ( lastStoredSequence == null ) true
                        else lastStoredSequence.range.last < firstNewSequence.range.first
                    }
```

Can you add a unit test providing coverage for this and add a fix commit?

yuanchen233: Yep, that's where I was going but did not manage to clarify, code speaks better.
I'll add the fix.

Whathecode: I don't think this warranted a new test. I just added an additional sequence to the existing `appendBatch_fails_for_overlapping_sequence_range` test which makes it more representative and made the test fail without the fix. (An alternative would be to add multiple test _cases_/input, instead of adding a new test. But, this is fine.)

I also updated the commit message to more clearly reflect this is a bug fix (useful when later writing release notes).

Whathecode (changes requested):
I removed the extra test in your first original commit and instead just updated an existing test to provide the same coverage. (force pushed).

For your original second commit I did some rework which I added as a third commit which you can review. If you agree, just squash it into your second commit, and I can approve after.

Have a look whether the JSON output needs to be regenerated, though (~~I believe not, but I didn't think hard about it 😅~~ The test fails, so yes, apparently you should).

Lastly, this is a new service endpoint, so the README for that subsystem needs to be updated.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/infrastructure/InMemoryDataStreamService.kt
yuanchen233: > Point 1: the `syncPoint` invariant is indeed less clear, but should arguably also be part of the `DataStreamBatch` contract.
> 
> But, to point 3: maybe there is no point in `DataStreamBatch` being open and we should just make it `sealed`. It would also get rid of the fallback `else` in `appendBatch` that currently exists there. I can't immediately see why we would have a requirement to leave `DataStreamBatch` open. I think it had more to do with having a read-only interface than anything else.

Do we want separate issues to keep those in the backlog? otherwise this can marked as resolved.

yuanchen233:
> Have a look whether the JSON output needs to be regenerated, though (~I believe not, but I didn't think hard about it 😅~ The test fails, so yes, apparently you should).
> 
It was the `isClosed` to `isOpen` change, I regenerated the output and also updated the JSON schema, 

> Lastly, this is a new service endpoint, so the README for that subsystem needs to be updated.

The endpoint documentation was already there, I also added a line in README's example code.

Whathecode (dismissed):

Whathecode (approved):

Issue #525, opened 2026-02-10
Lift out persistence of participants and groups from Recruitment for scalability


Whathecode:
Currently, participants and participant groups are loaded into a `Recruitment` aggregate root per study. This means that for _very large studies_, all participant data needs to be loaded into memory as part of each `Recruitment` operation. The intent behind this was that `Recruitment` should protect invariants, such as recruitment targets (number of participants, or distribution of male/female participants).

However, that of course doesn't scale, and there is currently a need to potentially recruit a million participants in a single study.

Looking a bit into invariants _currently_ protected by `Recruitment`:

- Prevent duplicate participants. That can just as well be implemented on the repository level, e.g. a `participantRepository.addParticipant()` call.
- Another is upon creating invitations, that the participants are actually part of the recruitment. That would have to be lifted out into `RecruitmentServiceHost`, or a supporting domain service.

So, I think there is something there. And, it may just be internal refactoring of the data model (and DB update, of course), but doesn't necessarily require changes to the service API.

Overall, it looks like persistence of participants in `Recruitment` should be lifted out. It could still protect invariants such as valid participant assignments, etc., and eventually update recruitment goals.

PR #529, opened 2026-02-23
Use `ParticipantGroupName` for name updates and add missing domain events

### LINKED ISSUE #319: Endpoints to edit staged participant groups

yuanchen233:
> As an architectural design rule, every side effect on an aggregate root needs to be emitted as an event. Then, a persistence layer can consume that and choose how to persist changes.

This PR adds the missing domain event for `updateParticipantGroup`, with the introduction to the generic `StagedParticipantGroup.representation` to enrich the current `name` filed.

---

TODO (?): Add tests to enforce event emission coverage per aggregate command (for persistence-relevant mutations). This works only if such mutations consistently go through aggregate-root methods; if ServiceHost can mutate internals directly, event gaps are still easy to miss.

---
This PR introduces breaking data-shape changes and requires a data migration:

1. `StagedParticipantGroup.name` was renamed to `StagedParticipantGroup.representation`.
2. The `ParticipantGroupAdded` domain event payload was changed from `name` to `representation`.

For version upgrades from 1.2 to 1.3:

1. `StagedParticipantGroup.representation` is a newly added field.
2. The `ParticipantGroupAdded` event payload has a new field`representation`.


carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -35,6 +35,11 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
             val participants: Set<AssignedParticipantRoles>,
             val name: String? = null
         ) : Event()
+        data class ParticipantGroupUpdated(
+            val groupId: UUID,
+            val participants: Set<AssignedParticipantRoles>? = null,
+            val name: String? = null
```
Whathecode: Only now I realize that `name` was nullable. 🤦‍♂️  That means that `updateParticipantGroup()` is ambiguous when `null` is passed for `name`. And, that currently it can't be set to `null` after having assigned something else.

We need to decide what makes sense.

The reason for `null` making sense is that the UI could list the names of participants by default. I still like the single endpoint for updates, as that is typically one use case. We are just lacking type information to tell `null` apart from "don't set".

Maybe a sealed class hierarchy of `ParticipantGroupName` with inheriting `Keep` and `Set`. `Keep` can be just an `object`, and it simply denotes the current `null`. `Set` can carry a `String` for the custom name. And, as a parameter to `updateParticipantGroup()`, `ParticipantGroupName` is then nullable, which means "don't change". That type would only be used as a parameter then. And, probably should be generic as it sounds like we can run into this in the future.

If you think that makes sense, I think this warrants a separate preceding "Fix: ..." commit.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentTest.kt
```diff
@@ -162,6 +162,10 @@ class RecruitmentTest
 
         assertEquals( "Updated name", group.name )
         assertEquals( updatedAssignments, group.roleAssignments )
+        assertEquals(
+            Recruitment.Event.ParticipantGroupUpdated( group.id, updatedAssignments, "Updated name" ),
```
Whathecode: Maybe time to assign `updatedName` to a `val` higher up. ;)

Whathecode: We need a similar assertion in the next test: `updateParticipantGroup_allows_name_change_after_deployed()`

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -248,6 +249,22 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
         event( Event.ParticipantGroupUpdated( groupId, participants, name ) )
     }
 
+    /**
+     * Mark the participant group with [groupId] as deployed.
+     *
+     * @throws IllegalArgumentException when the participant group with [groupId] does not exist.
+     * @throws IllegalStateException when the participant group has already been deployed.
+     */
+    fun inviteParticipantGroup( groupId: UUID )
```
Whathecode: If we go that route we should also reconsider whether `markAsDeployed()` should remain a mutable operation. But, this goes again towards the refactoring of staged participant groups we want to do (#525), so maybe for the time being we can just consider it a bug (missing event). Nobody is using SQL normalization with multiple tables which would benefit from this event data. And, even if they are, they wouldn't be blocked entirely as they can read and verify for changes.

Could you write it up and link #525?

Whathecode (changes requested):

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
Whathecode: I wrote this late last night. If we have a generic `OptionalSet<string?>` type with an inheriting `Set` and `Keep`, we have what we need I think. So I was mistaken to write that the whole argument would be `null`. :)

Setting up a serializer for generics is a bit more tricky, though. Have a look at what kotlinx.serialization supports. If that gets too messy we can reconsider.

The inconsistency with other "optional" arguments is already a bit messy, in fact. :( I can't immediately think of a better solution. It would be interesting to look at the tradeoff with a simple `setName` boolean parameter. Or even a `clearName()` endpoint. Or, separate `updateName` and `updateAssignment` endpoints. 🤔

yuanchen233: Yep, I would argue that the whole argument being nullable is still ambiguous, at least on readability level. 

We could give `updateParticipantGroup` a default value to `name = Keep`, which allows the caller to omit `name`.

I'm calling it `OptionalFieldUpdate` now, do you have a preference?

Whathecode: Sounds good.

Whathecode: What I dislike about that, though, is that the other parameter is still `null` == leave unchanged. The inconsistency is icky. 

A dedicated `ParticipantGroupName` with a nullable string in it would actually be sufficient. Then we can have the same semantics: if the full argument is `null` then leave unchanged, else set whatever the name inside is set to. 

This is also more in line with DDD's approach to adding value types for domain concepts. If we then ever add constraints (max length etc) we also have a place to enforce it. 

Sorry for going back and forth on this. 😅 What do you think? It's the cleanest approach I think, consistent with the rest of the endpoints, it can be applied consistently to future endpoints, and it fixes the issue.

yuanchen233: Agreed, I think consistency with other arguments is important. Otherwise it's just confusing.

I'll give it a try and let's see how it looks.

yuanchen233:
How strict are we regarding domain events? Do we want events to mean “something actually changed,” or just “an update was called”?

`Study` emits `InternalDescriptionChanged` every time we set name/description, even if the value is the same. I am following this behavior.

Then what about no-op updates? I think this is the first time we introduce it, as both argument are nullable. I added an check for this, and early return to skip the event emission.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/users/ParticipantGroupName.kt
```diff
@@ -0,0 +1,15 @@
+package dk.cachet.carp.studies.application.users
+
+import kotlinx.serialization.Serializable
+import kotlin.js.JsExport
+
+
+/**
+ * Value type for updating a participant group's name.
```
Whathecode: This shouldn't be purpose-specific to updating a group's name. It should simply represent a participant's group name, which is used when creating a group, updating it, packing it along with other data, etc. (the object doesn't care why it is used; as usual document the concept, not the callers, it keeps things decoupled).

As I wrote earlier:
> The reason for null making sense is that the UI could list the names of participants by default. ... We are just lacking type information to tell null apart from "don't set".

So `null` is a valid "name" for a participant group, which means "no specific name is set"; if you want to represent it, you will have to derive a name from the participants in the participant group. That's probably what the documentation of `ParticipantGroupName` should focus on. Maybe, this can lead to reconsider the type name, e.g., `ParticipantGroupRepresentation` with a nullable `name` field. It also indicates it could evolve to contain an icon identifier, etc.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -209,6 +214,8 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
 
     /**
      * Update the [participants] and/or [name] of an existing participant group with the specified [groupId].
+     * Only omitting both [participants] and [name] is treated as a no-op; supplying either argument records
+     * an update event, even when the supplied value matches the current state.
```
Whathecode: Regarding your comment in relation to this:
> How strict are we regarding domain events? Do we want events to mean “something actually changed,” or just “an update was called”?
>
> Study emits InternalDescriptionChanged every time we set name/description, even if the value is the same. I am following this behavior.
> 
> Then what about no-op updates? I think this is the first time we introduce it, as both argument are nullable. I added an check for this, and early return to skip the event emission.

Good observation. I would argue only firing the event if something actually changes is probably the "better" pattern to adopt, even if that currently doesn't happen in `Study`. These events are consumed by relational databases to trigger DB updates. So, they won't do harm (the data remains unchanged), but it seems to go against the "contract" of these events, which are designed to indicate changes to data (as opposed to let's say, logging _actions_ on data).

In the current code it is also counterintuitive that we have a `newRoleAssignments` boolean, so we actually go through the effort of comparing already.

I appreciate you are considering consistency here, but arguably the other place is wrong and should do a better job (not necessarily in this PR). :) Also, with data classes in Kotlin deep comparisons are trivial.

Whathecode (changes requested):

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/users/ParticipantGroupName.kt
yuanchen233: > Maybe, this can lead to reconsider the type name, e.g., ParticipantGroupRepresentation with a nullable name field. It also indicates it could evolve to contain an icon identifier, etc.

I’d prefer to keep it narrow for now. The current wrapper works because it represents a single nullable field, so we can distinguish between no change and explicitly clearing it. If we broaden it later, that distinction becomes unclear.

yuanchen233: > This shouldn't be purpose-specific to updating a group's name. It should simply represent a participant's group name, which is used when creating a group, updating it, packing it along with other data, etc. (the object doesn't care why it is used; as usual document the concept, not the callers, it keeps things decoupled).

I tried to implement this and realized when updating the schemas:

`ParticipantGroupName?` makes sense for `updateParticipantGroup`: outer null means "don't change the name", while ParticipantGroupName(null) means "remove the current explicit name".

For `createParticipantGroup` and `inviteNewParticipantGroup`, that difference does not matter, because there is no existing name yet. In those methods, both outer null and inner null mean "create the group without an explicit name", which makes the API less clear.

We could make the outer wrapper non-nullable, but to me it just adds complexity: callers are always forced to pass a wrapper object. Or, accepting both forms with the same meaning makes the API feel less strict, since two different inputs effectively represent the same state.

Whathecode: > Or, accepting both forms with the same meaning makes the API feel less strict, since two different inputs effectively represent the same state.

If that were the case, I would agree. But, `nullable` behavior of a parameter is defined by the place within which it is made nullable. Not by the _type_ which is made nullable.

In other words, the meaning of nullability for update methods doesn't have to be the same as that of create methods. For update methods, we want to be consistent and make it mean "do not modify". For create methods, nullability typically means "use defaults". The fact that a `null` name is default upon create can have a different meaning.

This is especially so if you consider it a `ParticipantGroupRepresentation` instead of `ParticipantGroupName`. If you add an icon, the default icon could be specific, or randomly generated (such as gravatar icons), instead of null.

yuanchen233: Yes, on a boarder level `null` semantics are defined in the use case, and I do agree nullable parameter can have method-specific meanings.

My point here is more narrower, what I meant is within `createParticipantGroup`, those two inputs:

`createParticipantGroup( uuid, groupAssignment, studyId, name = null)`
`createParticipantGroup( uuid, groupAssignment, studyId, name = ParticipantGroupName( value = null ) )`

appear to express the same outcome: create a group without an explicit name.

However, if we are going for the direction of making this `ParticipantGroupRepresentation`, it will make more sense:
`representation = null` means use defaults for the whole representation, and
`representation = ParticipantGroupRepresentation( name = null, icon = null )` can partiality set some fields.

---
I think the key point is not nullability itself, but that each level of nullability should have a defined meaning. If both outer `null` and inner `null` are valid, the API should make clear what each one represents.

Whathecode: > However, if we are going for the direction of making this ParticipantGroupRepresentation, it will make more sense:

I think this is a very likely future as requirements evolve. let's do the rename/scoping now as part of this change?

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
yuanchen233: I'll create a separate issue to keep track of other event emissions on this.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
```diff
@@ -77,20 +78,21 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
...
+            "dk.cachet.carp.common.application.UUID",
+            "dk.cachet.carp.studies.application.users.ParticipantGroupRepresentation"
         ),
         level = DeprecationLevel.WARNING
     )
     suspend fun inviteNewParticipantGroup(
         studyId: UUID,
         group: Set<AssignedParticipantRoles>,
-        name: String? = null
+        representation: ParticipantGroupRepresentation? = null
```
Whathecode: Since this endpoint is deprecated, I don't think you need to update the interface. And, if you update this, shouldn't that require version migration? 🤔

It's odd tests pass. I looked at it quickly, and, it seems we don't have coverage for name. The parameter isn't passed so the default is used.

I recommend to not update it to save work. However, a similar problem must exist for `createParticipantGroup`, and here we do want to update it, so version migration will be needed here.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -202,9 +208,14 @@ class RecruitmentServiceHost(
...
-     * If [group] is provided, role assignments are updated; unchanged otherwise.
-     * If [name] is provided, the group name is updated; unchanged otherwise.
+     * Participant assignments can't be changed after the group has been invited; the group representation can always
+     * be updated.
+     *
+     * @param group If set, role assignments are updated; unchanged otherwise.
+     * @param representation Optional group representation update:
+     * Pass `null` to keep the current name.
+     * Pass [ParticipantGroupRepresentation] with a non-null [ParticipantGroupRepresentation.name] to set an explicit name.
+     * Pass [ParticipantGroupRepresentation] with `null` [ParticipantGroupRepresentation.name] to clear the explicit name.
```
Whathecode: A bit convoluted, and, would require updating this documentation each time new fields get added to representation. Defer documentation to fields of `ParticipantGroupPresentation` instead.

Here, I think this suffices: "If set, group representation is updated; unchanged otherwise."

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -33,7 +34,7 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
         data class ParticipantAdded( val participant: Participant ) : Event()
         data class ParticipantGroupAdded(
             val participants: Set<AssignedParticipantRoles>,
-            val name: String? = null
+            val representation: ParticipantGroupRepresentation? = null
```
Whathecode, line 47: Make note in the PR description this will require database migration.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -197,17 +199,17 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
 
         validateRoleAssignments( status.studyProtocol, participants )
 
-        val group = StagedParticipantGroup( id, name )
+        val group = StagedParticipantGroup( id, representation?.name )
```
Whathecode: Shouldn't `StagedParticipantGroup` carry the full representation as well? 🤔 That could be a follow-up commit, though, if you want to untangle changesets a bit.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -197,17 +199,17 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
...
         group.addParticipants( participants )
 
         _participantGroups[ group.id ] = group
-        event( Event.ParticipantGroupAdded( participants, null ) )
+        event( Event.ParticipantGroupAdded( participants, representation ) )
```
Whathecode, line 217: Huh? Former bug, I suppose?

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentTest.kt
```diff
@@ -111,8 +112,7 @@ class RecruitmentTest
         assertTrue( recruitment.getStatus() is RecruitmentStatus.ReadyForDeployment )
 
         val roleAssignment = setOf( AssignedParticipantRoles( participant.id, AssignedTo.All ) )
-        val group = recruitment.addParticipantGroup( roleAssignment, groupName )
-        assertEquals( Recruitment.Event.ParticipantGroupAdded( roleAssignment ), recruitment.consumeEvents().last() )
```
Whathecode, line 115: I assume this assertion shouldn't be removed, but be updated.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentTest.kt
Whathecode: This test (and other "name" tests should be updated to focus on full representation updates instead of just the name. And, since `ParticipantGroupRepresentation` is a data class, you can do an equal assertion on the full class (instead of just the name).

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/users/ParticipantGroupRepresentation.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import kotlinx.serialization.Serializable
+import kotlin.js.JsExport
+
+
+/**
+ * The explicit representation of a participant group.
+ *
+ * When [name] is `null`, no explicit name is set for the participant group and a display name may instead be derived
+ * from the participants in the group.
```
Whathecode: Document this on `name` field instead.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentTest.kt
```diff
@@ -171,13 +172,45 @@ class RecruitmentTest
...
+        val recruitment = createReadyRecruitment()
+        val participant = recruitment.addParticipant( participantEmail )
+        val group = recruitment.addParticipantGroup(
+            setOf( AssignedParticipantRoles( participant.id, AssignedTo.All ) ),
+            ParticipantGroupRepresentation( "Initial name" )
+        )
+
+        recruitment.updateParticipantGroup( group.id, representation = ParticipantGroupRepresentation( null ) )
+
+        assertEquals( null, group.name )
```
Whathecode: I find this test a bit superfluous, as it overlaps highly with `updateParticipantGroup_updates_assignments_and_name_for_staged_group`. I understand you wanted to add an extra test for the "clearing" behavior, but one probably suffices.

This is in line with updating the scope of the tests to cover the full representation, not just focus on the "name".

Whathecode (changes requested):

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
yuanchen233: I also changed some of the test cases so they pass, and I did not add version migration, both because we are still in 1.3.

Both `inviteNewParticipantGroup.name` and `createParticipantGroup` were introduced in the same minor version, which is not released yet.

Do we want to bump version number again to add migration?

Whathecode: If there are unreleased endpoints, no, you don't need migration. If they have been released in a previous minor version, you likely do. 

The deprecated endpoint definitely was released before. 🙂 But as I indicated, probably shouldn't be changed. I'm on my phone, and can't easily check the specific endpoints now.

yuanchen233: Yes, but only the `name` field were be introduced in this minor version, and that feels a bit messy in the broader release context: we would be updating the endpoint to accept an extra `name` field on an endpoint that is effectively no longer meant to be used. I think it would be better to revert the addition of the `name` field for the `InviteNewParticipantGroup` endpoint altogether, to keep this minor release clean.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
yuanchen233: Yep, I now fix this in a separate fix commit.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
Whathecode: > Yes, but only the name field were be introduced in this minor version

I was just continuing the review again. So, indeed, by the looks of it, `name` was never released before. [Here you have](https://github.com/carp-dk/carp.core-kotlin/blob/0a465e9e08a28fbc6a1103ee5bdd2a64f739cd03/carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt) the `Recruitment` at the 1.2.0 release.

In that case, I agree.

> I think it would be better to revert the addition of the name field for the InviteNewParticipantGroup endpoint altogether, to keep this minor release clean.

I can see in your second commit that you still make updates to retain `name` instead of preceding it with a commit to simply remove `name`.

Sorry for the late response (other than the 👍) in case you were waiting for an answer from me.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
```diff
@@ -184,14 +191,14 @@ class RecruitmentServiceHost(
         groupId: UUID,
         group: Set<AssignedParticipantRoles>,
         studyId: UUID,
-        name: String?
+        representation: ParticipantGroupRepresentation?
```
Whathecode: We should either clearly document which representation is used in case none is passed (e.g., a `ParticipantGroupRepresentation.Default`), or, simply don't make it optional.

The former is better for API discoverability (it forces clients to make a choice). The latter is probably more tailored to the majority use case when no names are needed. I have no strong opinion one way or the other, but it shouldn't stay as it is now. :)

And, whatever you choose, that should probably trickle down to `Recruitment.addParticipantGroup()`.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceTest.kt
```diff
@@ -274,19 +271,33 @@ interface RecruitmentServiceTest
...
+        val assignParticipant = AssignedParticipantRoles( participant.id, AssignedTo.All )
+        val groupStatus = recruitmentService.createParticipantGroup( UUID.randomUUID(), setOf( assignParticipant ), studyId, groupRepresentation )
+
+        assertTrue( groupStatus is ParticipantGroupStatus.Staged )
+        assertEquals( participant, groupStatus.participants.single() )
+        assertEquals( groupRepresentation.name, groupStatus.name )
+    }
+
+    @Test
+    fun createParticipantGroup_succeeds_with_null_representation_name() = runTest {
```
Whathecode: Does this really need to be tested? It feels a bit like writing a test for the type system, which is otherwise just forwarding the value. If we'd want to test this, we'd ideally be using a test framework that supports data rows to remove duplication. As an alternative, I sometimes use a simple helper function with data defined within the test you iterate over. _But_, I think I'd keep it easy here given the lack of added coverage.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceTest.kt
```diff
@@ -519,31 +528,53 @@ interface RecruitmentServiceTest
...
+        val participant = recruitmentService.addParticipant( studyId, EmailAddress( "test@test.com" ) )
+        val assignments = setOf( AssignedParticipantRoles( participant.id, AssignedTo.All ) )
+        val groupStatus = recruitmentService.createParticipantGroup( UUID.randomUUID(), assignments, studyId, ParticipantGroupRepresentation( "Initial name" ) )
+        recruitmentService.inviteParticipantGroup( groupStatus.id )
+
+        val updatedStatus = recruitmentService.updateParticipantGroup( groupStatus.id, representation = ParticipantGroupRepresentation( null ) )
+
+        assertTrue( updatedStatus is ParticipantGroupStatus.Invited )
+        assertEquals( null, updatedStatus.name )
+    }
```
Whathecode: To me this really provides the same coverage as `updateParticipantGroup_allows_representation_update_after_invite`. The only difference is to check whether `null` is set instead of "Initial name". I would remove this test.

Same goes for the next test, which is really a duplicate of `updateParticipantGroup_succeeds_before_invite()`.

If you want to provide _some_ coverage for nullability, you could within the same test call `updateParticipantGroup` twice with different value. It'd still be "in scope" of that test without this massive duplication. :)

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentTest.kt
```diff
@@ -147,37 +148,56 @@ class RecruitmentTest
...
+        val updatedGroupRepresentation = ParticipantGroupRepresentation( "Renamed after deploy" )
+        recruitment.updateParticipantGroup( group.id, representation = updatedGroupRepresentation )
 
-        assertEquals( "Renamed after deploy", group.name )
+        assertEquals( updatedGroupRepresentation.name, group.name )
+        assertEquals( updatedGroupRepresentation, ParticipantGroupRepresentation( group.name ) )
+    }
+
+    @Test
+    fun updateParticipantGroup_allows_representation_clear_after_deployed()
```
Whathecode: Essential duplicate of `updateParticipantGroup_allows_representation_change_after_deployed`. Could be removed if you ask me.

Whathecode (changes requested):
I reviewed the first two commits and responded to outstanding comments/resolved those which were addressed.

This is looking pretty good! It seems like postponing the introduction of `representation` to `StagedParticipantGroup` will be good to streamline the rest of the review 🙏. Sorry for the long intervals between reviews. :( I think I have time tomorrow to look at the last two commits (in fact I think I already reviewed/approved the last commit before).

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
Whathecode: Bump! Creating an issue is a good idea. :) Link it here if/when done.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
yuanchen233: Cool, I simply missed it, will proceed with this direction.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHost.kt
yuanchen233: I agree that a `Default` value suits our use cases better. I am adding it as a companion object like `SyncPoint.UnixEpoch` ( could not finish it today so I pushed other changes first, have not review them myself yet )

And, thanks and no worries, this PR evolved quite a bit since where we started. I needed to spend more time on it as well anyways.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceTest.kt
```diff
@@ -372,7 +372,7 @@ interface RecruitmentServiceTest
         val updatedStatus = recruitmentService.updateParticipantGroup( groupId, updatedAssignments )
 
         assertTrue( updatedStatus is ParticipantGroupStatus.Staged )
-        assertEquals( "Initial name", updatedStatus.name )
+        assertEquals( ParticipantGroupRepresentation( "Initial name" ), updatedStatus.representation )
```
Whathecode: Minor: a bit better to store the originally set representation in a variable instead of initializing the same twice. Also in previous, next test, and some others.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentTest.kt
```diff
@@ -180,11 +180,10 @@ class RecruitmentTest
         val updatedGroupRepresentation = ParticipantGroupRepresentation( "Renamed after deploy" )
         recruitment.updateParticipantGroup( group.id, representation = updatedGroupRepresentation )
 
-        assertEquals( updatedGroupRepresentation.name, group.name )
-        assertEquals( updatedGroupRepresentation, ParticipantGroupRepresentation( group.name ) )
+        assertEquals( updatedGroupRepresentation, group.representation )
         recruitment.updateParticipantGroup( group.id, representation = ParticipantGroupRepresentation( null ) )
 
-        assertNull( group.name )
+        assertEquals( ParticipantGroupRepresentation( null ), group.representation )
```
Whathecode: Minor: maybe store the set presentation in a variable and reuse here as `expected`.

rpc/schemas/studies/users/ParticipantGroupStatus.json
```diff
@@ -7,12 +7,25 @@
...
     "ParticipantGroupStatus": {
       "type": "object",
       "properties": {
         "__type": true,
         "id": { "type": "string", "format": "uuid" },
-        "name": { "type": [ "string", "null" ] },
+        "representation": {
+          "anyOf": [
+            { "$ref": "#/$defs/ParticipantGroupRepresentation" },
+            { "type": "null" }
```
Whathecode: In the same vein as the plan is to make the endpoint non-nullable, the status would also become non-nullable.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentTest.kt
Whathecode: Especially since it's used again in the last commit for asserting the expected events.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/domain/users/Recruitment.kt
```diff
@@ -36,6 +36,11 @@ class Recruitment( val studyId: UUID, id: UUID = UUID.randomUUID(), createdOn: I
             val participants: Set<AssignedParticipantRoles>,
             val representation: ParticipantGroupRepresentation? = null
         ) : Event()
+        data class ParticipantGroupUpdated(
+            val groupId: UUID,
+            val participants: Set<AssignedParticipantRoles>? = null,
+            val representation: ParticipantGroupRepresentation? = null
```
Whathecode, line 47: It seems like it's `null` when it didn't change, and otherwise set. That makes sense, but that also warrants some documentation.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentTest.kt
```diff
@@ -184,6 +192,133 @@ class RecruitmentTest
...
+            setOf( AssignedParticipantRoles( participant.id, AssignedTo.All ) ),
+            groupRepresentation
+        )
+
+        recruitment.updateParticipantGroup( group.id, representation = groupRepresentation )
+
+        assertEquals( groupRepresentation, group.representation )
+        val updateEvents = recruitment.consumeEvents().filterIsInstance<Recruitment.Event.ParticipantGroupUpdated>()
+        assertEquals( 0, updateEvents.count() )
     }
```
Whathecode, line 278: I think the added test coverage is warranted here. Nice coverage! But, the resulting code is a bit bloaty. It should be possible to set up a test helper (at a minimum I believe the test setup can be reused across tests) which is called from each test, so you have something more akin to a parameterized test.

In essence, all of these test "the same" type of thing/abstraction (partial updates and resulting event emissions). I think it is fine this is tested by one method with a foreach within it. For example, [I make heavy use of that in this library](https://github.com/Whathecode/kotlinx.interval/blob/develop/kotlinx.interval.testcases/src/commonMain/kotlin/IntervalTest.kt#L253).

If the testing framework would support parameterized tests (the default multiplatform setup in Kotlin still doesn't) it would be an added bonus that each set of parameters shows up as a separate test, but I would classify that as a nice to have. I think maintenance/clarity should win out here, so I wouldn't roll x number of methods per set of parameters separately.

Whathecode (changes requested):

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentTest.kt
yuanchen233: I tried the `forEach` style “parameterized test”, but it ended up a bit messier than I expected. The setup here is more complicated than the example you provided, so I couldn’t quite get the same clean look.

I ended up using a combination of `SUT` and helper function style, while keeping the test cases separated. I think this still reduces the repeated setup, and fits this repo’s test style better. Let me know what you think about this approach.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
```diff
@@ -102,7 +104,7 @@ interface RecruitmentService : ApplicationService<RecruitmentService, Recruitmen
         groupId: UUID,
         group: Set<AssignedParticipantRoles>,
         studyId: UUID,
-        representation: ParticipantGroupRepresentation? = null
+        representation: ParticipantGroupRepresentation? = ParticipantGroupRepresentation.Default
```
Whathecode: This is still nullable. The point of introducing `Default` was that it can be made non-nullable. And, the documentation should be along the lines of: "default is used when no representation is passed".

And, once the PR is finished, I think the "default" commit can be squashed with the introduction of `ParticipantGroupRepresentation`.

Whathecode (changes requested):

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentTest.kt
```diff
@@ -185,6 +193,88 @@ class RecruitmentTest
...
+        assertEquals( sut.updatedAssignments, sut.group.roleAssignments )
+        assertEquals( sut.initialRepresentation, sut.group.representation )
+        assertEquals(
+            Recruitment.Event.ParticipantGroupUpdated( sut.group.id, sut.updatedAssignments, null ),
+            sut.recruitment.consumeEvents().last()
+        )
+    }
+
+    @Test
+    fun updateParticipantGroup_unchanged_assignments_with_new_name_emits_name_only()
```
Whathecode: `_with_new_representation_emits_representation_only`()

Similarly for next test name.

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/domain/users/RecruitmentTest.kt
Whathecode: Much better. Seems like some overlap in coverage here and there, but it's clear what's happening with less code bloat than before. 👍 Good enough for now.

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
yuanchen233: My bad, I added it to the wrong commit when merging conflicts when rebasing, fixing it now.

And yes, this should be squashed, I kept it separate for easier to review changes.

rpc/schemas/studies/users/ParticipantGroupStatus.json
```diff
@@ -7,12 +7,20 @@
...
   ],
   "$defs": {
+    "ParticipantGroupRepresentation": {
+      "type": "object",
+      "properties": {
+        "name": { "type": [ "string", "null" ] }
+      },
+      "required": [ "name" ],
+      "additionalProperties": false
+    },
```
Whathecode: Centralize this in a separate schema so that you don't have to update it for the request and status object separately.

Whathecode (changes requested):
Last comment! :)

carp.studies.core/src/commonMain/kotlin/dk/cachet/carp/studies/application/RecruitmentService.kt
Whathecode: Looks good. Go ahead with squashing and you can resolve this.

Whathecode (approved):

PR #536, opened 2026-03-27
 Enforce registrationCreatedOn on device registrations

### LINKED ISSUE #535: `CustomDeviceRegistration` drops `registrationCreatedOn` for unknown device registrations
`CustomDeviceRegistration` does not preserve `registrationCreatedOn` when deserializing unknown device registrations.

`DeviceRegistration` defines `registrationCreatedOn` as part of its serialized state, but `CustomDeviceRegistration.BaseMembers` only reads:

- deviceId
- deviceDisplayName
- additionalSpecifications (1.3)

Because of that, the wrapper gets a new registrationCreatedOn value from the base class initializer instead of the timestamp from the serialized JSON.

This is causing problem when calling endpoint [deviceDeployed](https://github.com/carp-dk/carp.sensing-flutter/issues/561), since `PrimaryDeviceDeployment.lastUpdatedOn` is calculated using `registrationCreatedOn`

yuanchen233:
Fix #535 

We are tightening the model so `registrationCreatedOn` is always required on a device registration. The downside is that older serialized registrations which were saved without this field may no longer deserialize.

The tricky part is that making this backward compatible would mean filling in a timestamp that was never actually there. That would keep old payloads working, but it would also bring back the original problem by silently inventing required data instead of enforcing it.

I think this is acceptable, since `deviceDeployed` would already fail for these registrations because of the mismatch caused by the original bug, so this is not functionality we expect to be used in practice. However, it should still be called out in the release notes as a backward-incompatible change.

Note: 
- This fix requires a database migration for existing`CustomDeviceRegistration` records to include `registrationCreatedOn`.
- `DefaultDeviceRegistration` constructor args were reordered to keep `additionalSpecifications` last. Positional external callers may need to update to named args. This does not affect migrating from 1.2

### DIFF
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/DeviceRegistration.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/DeviceRegistration.kt
@@ -1,4 +1,4 @@
-@file:Suppress( "NON_EXPORTABLE_TYPE" )
+@file:Suppress( "NON_EXPORTABLE_TYPE", "UnnecessaryAbstractClass" )
 
 package dk.cachet.carp.common.application.devices
 
@@ -6,7 +6,6 @@ import dk.cachet.carp.common.application.ApplicationData
 import dk.cachet.carp.common.application.Immutable
 import dk.cachet.carp.common.application.ImplementAsDataClass
 import dk.cachet.carp.common.infrastructure.serialization.NotSerializable
-import kotlinx.datetime.Clock
 import kotlinx.datetime.Instant
 import kotlinx.serialization.*
 import kotlin.js.JsExport
@@ -41,7 +40,7 @@ abstract class DeviceRegistration
     abstract val deviceDisplayName: String?
 
     @Required
-    val registrationCreatedOn: Instant = Clock.System.now()
+    abstract val registrationCreatedOn: Instant
 
     /**
      * Additional device specifications which may be relevant to the researcher when interpreting collected data.
--- a/carp.deployments.core/src/commonTest/kotlin/dk/cachet/carp/deployments/infrastructure/StudyDeploymentSnapshotTest.kt
+++ b/carp.deployments.core/src/commonTest/kotlin/dk/cachet/carp/deployments/infrastructure/StudyDeploymentSnapshotTest.kt
@@ -12,7 +12,6 @@ import dk.cachet.carp.deployments.domain.StudyDeploymentSnapshot
 import dk.cachet.carp.deployments.domain.createComplexDeployment
 import dk.cachet.carp.deployments.domain.studyDeploymentFor
 import dk.cachet.carp.protocols.infrastructure.test.createEmptyProtocol
-import kotlinx.serialization.*
 import kotlin.test.*
 
 
--- a/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/AltBeacon.kt
+++ b/carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/AltBeacon.kt
@@ -12,6 +12,8 @@ import dk.cachet.carp.common.application.sampling.NoOptionsSamplingScheme
 import dk.cachet.carp.common.application.sampling.SamplingConfiguration
 import dk.cachet.carp.common.application.tasks.TaskConfigurationList
 import dk.cachet.carp.common.infrastructure.serialization.NotSerializable
+import kotlinx.datetime.Clock
+import kotlinx.datetime.Instant
 import kotlinx.serialization.*
 import kotlin.js.JsExport
 import kotlin.reflect.KClass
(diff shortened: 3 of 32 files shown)


yuanchen233:
Questions:
Should we keep `applicationData` as the last argument/field? There is no API change here, since `registrationCreatedOn` was already required before. So this mostly comes down to how we want to structure it internally.

Both fields have default values. `applicationData` is nullable; `registrationCreatedOn` is not something we expect callers to set explicitly in most cases.

Whathecode:
This sounds like a bug fix. The commit should be prefixed with "Fix: ". ;)

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/devices/DeviceRegistration.kt
```diff
@@ -41,7 +40,7 @@ abstract class DeviceRegistration
     abstract val deviceDisplayName: String?
 
     @Required
-    val registrationCreatedOn: Instant = Clock.System.now()
+    abstract val registrationCreatedOn: Instant
```
Whathecode, line 43: Why does this have to be abstract? Maybe because you had to override it for serialization with `BaseMembers`? In that case, just making it `open` should work as well, and, then all the other existing classes don't need to change.

yuanchen233: Thanks, I was trying to figure out a way make the serialization work without change existing classes. this is exactly what I needed and will give it a try, soon.

yuanchen233: I encountered `duplicate serial name` issue when serialize the overriding body property, and I see a few ways to deal with this:
1. Make `registrationCreatedOn` abstract, this is current implementation.
2. Make `BaseMembers` a plain DTO by not inheriting `DeviceRegistration`, but this adds more maintain work for future `DeviceRegistration` changes, and also different from other `BaseMember` use cases.
3. I am pretty sure there are way to work around, but I don't think it is worth the effort (and potentially unpleasing code)

Whathecode: I see. Seemingly a known limitation: https://github.com/Kotlin/kotlinx.serialization/issues/427#issuecomment-939982702

However, it's not all too bad, because as you have found out, for testability you should be able to pass this either way. :)

carp.deployments.core/src/commonTest/kotlin/dk/cachet/carp/deployments/infrastructure/StudyDeploymentSnapshotTest.kt
```diff
@@ -58,6 +58,17 @@ class StudyDeploymentSnapshotTest :
...
+    fun unknown_registration_preserves_registrationCreatedOn()
+    {
+        val expectedRegistrationCreatedOn = TestClock.now()
+        val serialized = serializeDeploymentSnapshotIncludingUnknownRegistration()
+        val parsed: StudyDeploymentSnapshot = JSON.decodeFromString( serialized )
+
+        val registration = parsed.deviceRegistrationHistory.values.flatten().single() as CustomDeviceRegistration
+        assertEquals( expectedRegistrationCreatedOn, registration.registrationCreatedOn )
+    }
+
```
Whathecode: Do we really need this test? It feels equivalent to adding a "preserves_deviceDisplayName" etc. test here, which feels out of scope.

You have the coverage you needed already by updating `UnknownDeviceSerializersTest`. Unfortunately, that test wasn't strong enough to catch it _missed_ coverage. But, that's a different challenge this test here doesn't solve either. If you'd really want to provide coverage for that you probably need to use reflection, but it's fine forgetting about that for now. :)

Whathecode (commented):
Sorry for the long delay again. :( I looked into the "duplicate serialization key" issue, and I seem to recall I even ran into the same in the past. No big deal.

> Should we keep applicationData as the last argument/field?

To me, it feels it belongs at the very end. All the "core" stuff, and then the "additional" non-core stuff.

Whathecode:
> To me, it feels it belongs at the very end. All the "core" stuff, and then the "additional" non-core stuff.

Oops. Now looking at it. It seems I got that wrong. The reasoning for keeping it at the end would be that you typically _don't_ provide it since the defaults suffice (as you wrote: "not something we expect callers to set explicitly in most cases"). Same reason `additionalSpecifications` came before `deviceId` for `DefaultDeviceRegistration`.

The ordering I mentioned still makes sense, but I guess this takes precedence. And, it's minor either way, since indeed, clients will typically use the builders either way. It does clean up the call sites within our codebase.

I'll roll the change back for you. I also bundled all unused import removals in a separate commit so that the commit history is clearer (otherwise it looked like you modified the file to fix `registrationCreatedOn`). And, I suppressed a deprecation warning that was there since you deprecrated an endpoint way back. This way there are no warnings in the project. ;)

Whathecode (dismissed):

Whathecode (dismissed):

Whathecode (dismissed):

Whathecode (approved):

PR #538, opened 2026-05-22
Update dependencies

yuanchen233:
This PR has been reworked into #539, #542, and #543.


yuanchen233:
Kotlin 2.1.20 -> 2.2.20: the main issue was Kotlin/JS [compiling Long values to JavaScript BigInt](https://kotlinlang.org/docs/whatsnew2220.html#usage-of-the-bigint-type-to-represent-kotlin-s-long-type). The build also fails without enabling `-Xes-long-as-bigint`, so I enabled it, but initially kept our existing `kotlin.Long` facade support.

In a separate commit, I removed `kotlin.Long` from our facade and relied on Kotlin/JS BigInt instead. For TypeScript, I kept deprecated compatibility aliases.

Kotlin 2.2.x -> 2.3.x: [exposing non-exportable API from interfaces](https://youtrack.jetbrains.com/issue/KT-83009) became stricter in 2.3.20. Since this project treats warnings as errors, I could not find a clean way to fix this without changing the public API in the data stream subsystem. For that reason, I kept the upgrade at 2.3.10.

Whathecode:
> Since this project treats warnings as errors, I could not find a clean way to fix this

Suppressing them is not possible/doesn't work?

publish-npm-packages/build.gradle
```diff
@@ -4,8 +4,11 @@ apply {
 
 kotlin {
     js(IR) {
-        moduleName = project.name.replaceAll("\\.", "-") + "-generated"
         binaries.executable() // Export JS/TypeScript files.
+        compilerOptions {
+            moduleName.set(project.name.replaceAll("\\.", "-") + "-generated")
```
yuanchen233: This is not working, the package name is now 'carp.core-kotlin-carp.common' instead of 'carp-common-generated'.

yuanchen233:
> Suppressing them is not possible/doesn't work?

I tried `@Suppress( "EXPOSED_NOT_EXPORTED_SUPER_INTERFACE" )`, but didn't get it to work yesterday, I might be missing something , I'll give it another try after fix the package naming issue.

yuanchen233:
It turns out to be a lot messier than we expected, Kotlin/JS 2.3.X introduced several stdlib/runtime declarations moved or became exposed differently, especially collections, Pair, Long/BigInt, Instant/Clock, generated/mangled names, and companion object access patterns.

As I do the explorations/learning on this topic, @Whathecode your old comment [here](https://youtrack.jetbrains.com/issue/KT-34995) took me to [here](https://youtrack.jetbrains.com/issue/KT-52641/KJS-IR-more-granular-control-over-JsExport-scope-and-behavior-and-fallback-mechanisms), where I noticed __Collections are exportable__ now. I tried to adopt to it but kept public API unchanged in a commit, fully migrate to it will lead to backward-incompatible ( Map -> KtMap, etc., ) and tests need to be updated where I cannot have controlled verification. 

`Kotlinx-datetime` 0.8.0 also moved `Instant` and `Clock` to Kotlin `stdlib` (`kotlin.time`). A lot of imports update and adjusted the TypeScript facade so the real runtime types come from `@cachet/carp-kotlin`, while `@cachet/carp-kotlinx-datetime` is kept for compatibility.

I am not confident on some changes, especially the facade structure changes. I don't know the topic/TypeScript good enough, and ideas were suggested by AI, where I don't know if they are preferred ways of handling those. However, given that it compiles and passes all the tests, I pushed them here, at least they can be used for reference. 😃

Whathecode:
I think I haven't followed semantic versioning for previous JS/Typescript releases. 🤔 The versioning is primarily based on the serialized endpoints, I believe. 

That would make sense, if it can help avoid taking on tech debt in the TypeScript wrappers. I can have a look what changed maybe tonight.

Whathecode (commented):
I started looking at this. But, we'll benefit from splitting this up in separate PRs. I'm creating one for the first commit now.

typescript-declarations/carp-kotlin/index.ts
```diff
@@ -53,17 +50,13 @@ export namespace kotlinExport.collections
 }
 export namespace kotlinExport.time
 {
-    export interface Duration
-    {
-        get inWholeMilliseconds(): number
-        get inWholeMicroseconds(): number
-    }
```
Whathecode, line 60: These shouldn't have been removed.

typescript-declarations/tests/carp-kotlin-test.ts
```diff
@@ -40,11 +40,10 @@ describe( "kotlin", () => {
     } )
 
     describe( "Long", () => {
-        it( "toLong and back toNumber equals", () => {
+        it( "toLong returns bigint", () => {
```
Whathecode, line 43: To have proper backwards compatibility with existing sources, assignment from bigint to Long is needed. The generated sources output bigint now.

```
declare global
{
    // Backwards compatibility for codebases which still use Long.
    interface BigInt extends kotlinExport.Long {}
}
```

Whathecode (commented):
Reviewed and reworked the first 3 commits (upgrade to 2.2.20): https://github.com/carp-dk/carp.core-kotlin/pull/539

typescript-declarations/carp-kotlin/kotlin-kotlin-stdlib.d.ts
```diff
@@ -62,9 +46,12 @@ declare module "@cachet/kotlin-kotlin-stdlib"
 
         namespace collections
         {
-            const KtList: any
-            const KtSet: any
-            const KtMap: any
+            type KtList<T> = { readonly __doNotUseOrImplementIt: any, asJsReadonlyArrayView(): Array<T> }
```
Whathecode, line 49: We should transition fully to `KtList` and other collection types, which are included in TypeScript declarations since Kotlin 2.0.0. This will require redeclaring the declarations here and exposing them in the wrapper. E.g., I started experimenting and am making progress with:
```
            namespace KtList
            {
                function fromJsArray<T>( elements: T[] ): KtList<T>
            }
            interface KtList<T>
            {
                asJsReadonlyArrayView(): readonly T[]
            }
```

We should erase our own `List` TS types, or at least, map them, and any other methods, to deprecated endpoints.

Whathecode:
@yuanchen233 Anything left to merge here, or this can be closed now?

yuanchen233:
> @yuanchen233 Anything left to merge here, or this can be closed now?

No, that's it, I'm closing it.

This PR has been reworked into #539, #542, and #543.

PR #539, opened 2026-06-03
Update kotlin 2.2.20

Whathecode:
Split off from #538 to simplify review process and untangling conflicting changes in subsequent commits due to changes which should actually happen earlier as part of this commit.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -93,7 +93,7 @@ configure( subprojects - devOpsModules ) {
                 useJUnitPlatform()
             }
         }
-        js(IR) {
+        js {
             moduleName = project.name.replaceAll("\\.", "-") + "-generated"
             binaries.executable() // Export JS/TypeScript files.
             browser()
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,7 @@ buildscript {
 
         versions = [
             // Kotlin multiplatform versions.
-            kotlin:'2.1.20',
+            kotlin:'2.2.20',
             serialization:'1.8.1',
             coroutines:'1.10.2',
             datetime:'0.6.2',
@@ -96,6 +96,13 @@ configure( subprojects - devOpsModules ) {
         js {
             moduleName = project.name.replaceAll("\\.", "-") + "-generated"
             binaries.executable() // Export JS/TypeScript files.
+            compilerOptions {
+                // Technically, since bigint is enabled, the target is at least "es2020", but this is not supported:
+                // https://youtrack.jetbrains.com/issue/KT-83011/Xes-long-as-bigint-documentation-improvements#focus=Comments-27-13867805.0-0
+                target = "es5"
+                freeCompilerArgs.add('-Xes-long-as-bigint')
+                freeCompilerArgs.add('-XXLanguage:+JsAllowLongInExportedDeclarations')
+            }
             browser()
             generateTypeScriptDefinitions()
         }
--- a/publish-npm-packages/build.gradle
+++ b/publish-npm-packages/build.gradle
@@ -3,7 +3,7 @@ apply {
 }
 
 kotlin {
-    js(IR) {
+    js {
         moduleName = project.name.replaceAll("\\.", "-") + "-generated"
         binaries.executable() // Export JS/TypeScript files.
         browser()
--- a/publish-npm-packages/build.gradle
+++ b/publish-npm-packages/build.gradle
@@ -6,6 +6,12 @@ kotlin {
     js {
         moduleName = project.name.replaceAll("\\.", "-") + "-generated"
         binaries.executable() // Export JS/TypeScript files.
+        compilerOptions {
+            // Technically, since bigint is enabled, the target is at least "es2020", but this is not supported:
+            // https://youtrack.jetbrains.com/issue/KT-83011/Xes-long-as-bigint-documentation-improvements#focus=Comments-27-13867805.0-0
+            target = "es5"
+            freeCompilerArgs.add('-Xes-long-as-bigint')
+        }
         browser()
     }
 
--- a/publish-npm-packages/src/forced-exports/kotlin-kotlin-stdlib
+++ b/publish-npm-packages/src/forced-exports/kotlin-kotlin-stdlib
@@ -13,3 +13,4 @@ mapOf_0
 Companion_getInstance_20
 _Duration___get_inWholeMilliseconds__impl__msfiry
 _Duration___get_inWholeMicroseconds__impl__8oe8vv
+Duration__toString_impl_8d916b
--- a/publish-npm-packages/src/forced-exports/kotlin-kotlin-stdlib
+++ b/publish-npm-packages/src/forced-exports/kotlin-kotlin-stdlib
@@ -1,5 +1,3 @@
-Long
-toLong
 EmptyList
 AbstractMutableList
 HashSet
--- a/rpc/src/main/kotlin/dk/cachet/carp/rpc/GenerateExampleRequests.kt
+++ b/rpc/src/main/kotlin/dk/cachet/carp/rpc/GenerateExampleRequests.kt
@@ -37,8 +37,9 @@ import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.JsonObject
 import kotlin.reflect.KFunction
 import kotlin.reflect.KSuspendFunction3
+import kotlin.reflect.full.declaredFunctions
 import kotlin.reflect.jvm.javaMethod
-import kotlin.reflect.jvm.kotlinFunction
+import kotlin.reflect.jvm.jvmErasure
 import kotlin.time.Duration.Companion.days
 import kotlin.time.Duration.Companion.seconds
 
@@ -48,7 +49,7 @@ import kotlin.time.Duration.Companion.seconds
  */
 fun generateExampleRequests( serviceInfo: ApplicationServiceInfo ): List<ExampleRequest>
 {
-    val requestMethods = serviceInfo.serviceKlass.methods
+    val requestMethods = serviceInfo.serviceKlass.kotlin.declaredFunctions
     val exampleRequests = exampleRequests.filter { it.key.javaMethod?.declaringClass == serviceInfo.serviceKlass }
 
     val json = Json( createDefaultJSON() ) { prettyPrint = true }
@@ -56,12 +57,11 @@ fun generateExampleRequests( serviceInfo: ApplicationServiceInfo ): List<Example
     return requestMethods.map { request ->
         // Retrieve example and verify whether it is valid.
         val requestName = serviceInfo.serviceName + "." + request.name
-        val kotlinRequest = checkNotNull( request.kotlinFunction )
-        val example = checkNotNull( exampleRequests[ kotlinRequest ] )
+        val example = checkNotNull( exampleRequests[ request ] )
             { "No example request and response instances provided for $requestName." }
-        check( example.request.matchesServiceRequest( kotlinRequest ) )
+        check( example.request.matchesServiceRequest( request ) )
             { "Incorrect request instance provided for $requestName." }
-        check( request.returnType.isInstance( example.response ) )
+        check( request.returnType.jvmErasure.isInstance( example.response ) )
             { "Incorrect response instance provided for $requestName." }
 
         // Create example JSON.
@@ -70,9 +70,9 @@ fun generateExampleRequests( serviceInfo: ApplicationServiceInfo ): List<Example
         val responseJson = json.encodeToString( checkNotNull( exampleJson[ "response" ] ) )
 
         ExampleRequest(
-            request,
+            checkNotNull( request.javaMethod ),
             ExampleRequest.JsonExample( example.request::class.java, requestObjectJson ),
-            ExampleRequest.JsonExample( request.returnType, responseJson )
+            ExampleRequest.JsonExample( request.returnType.javaClass, responseJson )
         )
     }
 }
--- a/typescript-declarations/carp-kotlin/index.ts
+++ b/typescript-declarations/carp-kotlin/index.ts
@@ -6,21 +6,32 @@ import extend from "@cachet/kotlin-kotlin-stdlib"
 export namespace kotlinExport
 {
     export type Nullable<T> = T | null | undefined
-    export interface Long
-    {
-        toNumber(): number
-    }
-    export const toLong: (number: number) => Long = extend.$_$.toLong
+    /**
+     * @deprecated Use bigint directly.
+     */
+    export type Long = bigint
+    /**
+     * @deprecated Use BigInt( number ) or bigint literal.
+     */
+    export const toLong = (number: number): Long => BigInt( number )
     export class Pair<K, V>
     {
         constructor( first: K, second: V ) {
             let kotlinPair = new extend.$_$.Pair( first, second );
-            kotlinPair.first = kotlinPair.me_1;
-            kotlinPair.second = kotlinPair.ne_1;
+            kotlinPair.first = kotlinPair.od_1;
+            kotlinPair.second = kotlinPair.pd_1;
             return kotlinPair;
         }
         get first(): K { return this.first; }
         get second(): V { return this.second; }
+    }
+}
+declare global
+{
+    // Backwards compatibility for codebases which still use Long.
+    interface BigInt
+    {
+        toNumber(): number
     }
 }
 export namespace kotlinExport.collections
@@ -55,16 +66,16 @@ export namespace kotlinExport.time
 {
     export interface Duration
     {
-        get inWholeMilliseconds(): number
-        get inWholeMicroseconds(): number
+        get inWholeMilliseconds(): bigint
+        get inWholeMicroseconds(): bigint
         toDurationString(): string
     }
     export namespace Duration
     {
         export const Companion: any = extend.$_$.Companion_getInstance_20()
-        export const parseIsoString: (isoDuration: string) => Duration = Companion.ii
-        export const ZERO: Duration = Companion.fi_1
-        export const INFINITE: Duration = Companion.gi_1
+        export const parseIsoString: (isoDuration: string) => Duration = Companion.kh
+        export const ZERO: Duration = Companion.hh_1
+        export const INFINITE: Duration = Companion.ih_1
     }
 }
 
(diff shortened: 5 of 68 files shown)


publish-npm-packages/src/forced-exports/kotlin-kotlin-stdlib
```diff
@@ -1,4 +1,3 @@
-Long
 toLong
```
Whathecode: This can be removed already as well. I've added a fixup commit: c608e3fac1506de5cf9f0ea6a808df3b0d675793

rpc/src/main/kotlin/dk/cachet/carp/rpc/GenerateExampleRequests.kt
```diff
@@ -48,7 +48,7 @@ import kotlin.time.Duration.Companion.seconds
  */
 fun generateExampleRequests( serviceInfo: ApplicationServiceInfo ): List<ExampleRequest>
 {
-    val requestMethods = serviceInfo.serviceKlass.methods
+    val requestMethods = serviceInfo.serviceKlass.methods.filter { it.kotlinFunction != null }
```
Whathecode: This feels like it should be fixed in a preceding refactoring commit. It seems like `serviceInfo.serviceKlass.kotlin.declaredFunctions` could be used here instead to sidestep the issue altogether. That is more closely aligned with the intent.

It also begs the question why `ApplicationServiceInfo` operates on JVM reflection classes instead of Kotlin; something worthwhile to see whether further refactoring can be done.

typescript-declarations/carp-kotlin/index.ts
```diff
@@ -61,9 +61,9 @@ export namespace kotlinExport.time
     export namespace Duration
     {
         export const Companion: any = extend.$_$.Companion_getInstance_20()
-        export const parseIsoString: (isoDuration: string) => Duration = Companion.ii
-        export const ZERO: Duration = Companion.fi_1
-        export const INFINITE: Duration = Companion.gi_1
+        export const parseIsoString = (isoDuration: string): Duration => Companion.kh( isoDuration ) as Duration
```
Whathecode: Odd that would be needed. Before an existing function was simply assigned. Now it is wrapped in another call.

typescript-declarations/carp-kotlin/index.ts
```diff
@@ -61,9 +61,9 @@ export namespace kotlinExport.time
...
+        export const ZERO: Duration = Companion.hh_1 as Duration
+        export const INFINITE: Duration = Companion.ih_1 as Duration
```
Whathecode: Why are casts needed all of a sudden? 😕

publish-npm-packages/build.gradle
```diff
@@ -3,9 +3,12 @@ apply {
 }
 
 kotlin {
-    js(IR) {
+    js {
```
Whathecode, line 6: From the [2.2.20 release notes](https://kotlinlang.org/docs/whatsnew2220.html#usage-of-the-bigint-type-to-represent-kotlin-s-long-type):

> Starting with Kotlin 2.2.20, Kotlin/JS now uses JavaScript's built-in BigInt type to represent Kotlin's Long values when compiling to modern JavaScript (ES2020).

Do we need to do anything extra here to compile to ES2020, or that's the default? I got confused, because the `tsconfig.json` is only updated to ES2020 [in the second commit on the original PR](https://github.com/carp-dk/carp.core-kotlin/pull/538/changes/e6db1a647c8f6922c906c905ee09a0f16d15b37f#diff-69792931d0f35c151be9b3f62ca807c8a5e909a5213a45a3032c5b57028a5530L3).

We likely want to do that already here if that is related. At least, that seems to work.

Whathecode: [Confusing](https://kotlinlang.org/docs/js-overview.html) ... 
> The current implementation of Kotlin/JS targets the [ES5](https://www.ecma-international.org/ecma-262/5.1/) and [ES2015](https://262.ecma-international.org/6.0/) standards.

We should also verify what the downside is of targeting ES2020 instead of ES5 or ES2015. I suppose ES2020 is supported widely enough, but worth checking.

publish-npm-packages/src/forced-exports/kotlin-kotlin-stdlib
```diff
@@ -1,4 +1,3 @@
-Long
```
Whathecode, line 1: If we no longer force this export because there is support for this by default, that means we can remove the NON_EXPORTABLE_TYPE warnings: https://github.com/carp-dk/carp.core-kotlin/issues/528

publish-npm-packages/build.gradle
```diff
@@ -6,6 +6,9 @@ kotlin {
     js(IR) {
         moduleName = project.name.replaceAll("\\.", "-") + "-generated"
         binaries.executable() // Export JS/TypeScript files.
+        compilerOptions {
+            freeCompilerArgs.add('-Xes-long-as-bigint')
+        }
```
Whathecode, line 14: Apparently, the reason for already doing this now, as opposed to in a separate commit, is that not doing so gives the following compilation error:
> Long can't be exported without using of the bigint type. Add -Xes-long-as-bigint compiler argument or set target to 'es2015'

That would be useful information in the commit message. :)

This doesn't seem to have a diagnostics key we can suppress. I tried to [enable](https://kotlinlang.org/docs/compiler-reference.html#xrender-internal-diagnostic-names) `-Xrender-internal-diagnostic-names` and nothing showed up.

typescript-declarations/carp-kotlin/index.ts
```diff
@@ -102,35 +96,40 @@ declare module "@cachet/kotlin-kotlin-stdlib"
...
 // Implement base interfaces in internal types.
-extend.$_$.Long.prototype.toNumber = function(): number { return this.h1(); };
-Object.defineProperty( extend.$_$.Long.prototype, "inWholeMilliseconds", {
+(BigInt.prototype as any).toNumber = function(): number { return Number( this.valueOf() ); };
+const infiniteDuration = kotlinExport.time.Duration.INFINITE
+Object.defineProperty( BigInt.prototype, "inWholeMilliseconds", {
     get: function inWholeMilliseconds()
     {
-        return extend.$_$._Duration___get_inWholeMilliseconds__impl__msfiry(this).toNumber();
+        if ( this.valueOf() === infiniteDuration ) return -1
```
Whathecode: Why was this added? 😕

rpc/src/main/kotlin/dk/cachet/carp/rpc/GenerateExampleRequests.kt
Whathecode: I've added a preceding commit that fixes it: b57fd53a879ae25cc4ba7a95809dcdda826f5d52

I haven't looked deeper into whether we can do something better with `ApplicationServiceInfo` here.

Whathecode: Probably `ApplicationServiceInfo` can be cleaned up by using `KClass`.

But, `ApplicationServiceInfo` currently lives in `carp.common` on JVM targets. I suppose the goal was for this to be useful to back-end infrastructures to simplify wiring up application services (e.g., it provides a dynamic way of getting the right serializers). By not using `KClass` (and its members), there is no need to pull in a `kotlin.reflect` dependency.

It seems like an architecturally sound decision not to have `carp.common` start depending on `kotlin.reflect`. On the other hand, for _backend_ infrastructures, there is really no harm in doing so. This probably is an indication that a separate output package tailored to this purpose may be warranted at some point in the future. A lot of what currently lives in `carp.common.infrastructure` (such as `ApplicationServiceApiMigrator` which is exposed by `ApplicationServiceInfo`) really only make sense to backends.

But, that means that for now we can consider this out of scope for this PR. It's a larger topic for which I will create a separate issue.

Whathecode: >  It's a larger topic for which I will create a separate issue.

New issue: https://github.com/carp-dk/carp.core-kotlin/issues/540

typescript-declarations/carp-kotlin/kotlin-kotlin-stdlib.d.ts
```diff
@@ -44,28 +40,28 @@ declare module "@cachet/kotlin-kotlin-stdlib"
...
         interface HashMap<K, V> extends Map<K, V> {}
         function mapOf_0<K, V>( pairs: Pair<K, V>[] ): Map<K, V>
 
-        interface Duration extends Long {}
+        type Duration = Long
         interface DurationCompanion
         {
             // parseIsoString
-            ii(): Duration
+            kh( isoDuration: string ): Duration
```
Whathecode, line 55: While the previous worked, this is more correct, indeed. 👍 I don't think the declaration verifier currently verifies parameters. If it would, it would fail without adding the parameter here.

typescript-declarations/carp-kotlin/index.ts
Whathecode: It wasn't needed. I removed this in a fixup commit: 790ceaf659fb2e42538c3f81b5de062d6bcebdc3

How was this code added? Was this done by AI and it actually managed to find the mangled names? 😮

typescript-declarations/carp-kotlin/index.ts
Whathecode: It wasn't needed. I removed this in a fixup commit: 790ceaf659fb2e42538c3f81b5de062d6bcebdc3

How was this code added? Was this done by AI and it actually managed to find the mangled names? 😮

publish-npm-packages/build.gradle
Whathecode: And indeed. [This is the property](https://kotlinlang.org/api/kotlin-gradle-plugin/older/2.2.0/kotlin-gradle-plugin-api/org.jetbrains.kotlin.gradle.dsl/-kotlin-js-compiler-options/target.html) to change the target. It also documents the only available options are "es5" and "es2015", with the former being default. When I try setting it to "es2020" (to try whether the documentation is outdated), I get:
> Unsupported ECMA version: ES2020

build.gradle
```diff
@@ -15,7 +15,7 @@ buildscript {
 
         versions = [
             // Kotlin multiplatform versions.
-            kotlin:'2.1.20',
+            kotlin:'2.2.20',
```
Whathecode, line 18: It seems this version also fixes: https://youtrack.jetbrains.com/issue/KT-77320/KJS-Big.js-times-is-compiled-to-multiply-operator

That should allow us to get rid of [the hack in `SyncPoint`](https://github.com/carp-dk/carp.core-kotlin/blob/develop/carp.data.core/src/jsMain/kotlin/dk/cachet/carp/data/application/SyncPoint.kt#L21).

More generally, I also wonder whether we still need Big.js in the data subsystem (as used in `SyncPoint`) if we have support for bigint now. 🤔 They may be unrelated; will have to check.

publish-npm-packages/build.gradle
Whathecode: I [think I understand now](https://youtrack.jetbrains.com/issue/KT-79644/BigInt-enabled-for-ES-2015-despite-being-an-ES-2020-feature#focus=Comments-27-13867793.0-0).

When enabling `-Xes-long-as-bigint`, you just enable `bigint` functionality, regardless of the current target set in `compileOptions`. That means that the output is no longer compliant with `es5`. But, as I write, I would then just consider the JS we output as `es2020`:

> I'm not that familiar with JS targets, but, isn't es2020 fully compatible with output of earlier versions? Would it then not make more sense to include a es2020 target and require anyone opting into this experimental feature to make that the target?
>
> Even if the generated JS code doesn't make use of other es2020 and intermediate version features for now (other than bigint the code remains the same), at least you won't run into surprises.

So, yes, it would make sense to also target es2020 in `tsconfig.json` for the wrappers. And, in addition, to try to update the Kotlin JS target to "es2015" which may have some additional advantages.

It seems like [es2020 is supported by at least 95% of browser versions currently out there](https://caniuse.com/?feats=mdn-javascript_operators_optional_chaining%2Cmdn-javascript_operators_nullish_coalescing%2Cmdn-javascript_builtins_globalthis%2Ces6-module-dynamic-import%2Cbigint%2Cmdn-javascript_builtins_promise_allsettled%2Cmdn-javascript_builtins_string_matchall%2Cmdn-javascript_statements_export_namespace%2Cmdn-javascript_operators_import_meta&list=1), so I see no problems there.

I added a commit which clarifies things and bumps `tsconfig` already: fad97d7827971cfac27a9d6f014737daf7d85d44
But, trying to upgrade Kotlin JS target to es2015 [can wait](https://github.com/carp-dk/carp.core-kotlin/issues/541).

typescript-declarations/carp-kotlin/index.ts
```diff
@@ -10,13 +10,13 @@ export namespace kotlinExport
     {
         toNumber(): number
     }
-    export const toLong: (number: number) => Long = extend.$_$.toLong
+    export const toLong = (number: number): Long => BigInt( number ) as any as Long
```
Whathecode: I don't understand this. The `tsconfig.json` was set to `es6`, which doesn't support `bigint`. Yet, somehow this compiles/doesn't complain.

I tried to read up a bit, and maybe this is loading some "polyfill", but I couldn't quite figure it out.

Regardless, it may be irrelevant [since I think it is best](https://github.com/carp-dk/carp.core-kotlin/pull/539/changes/5fbe4fedc09c04bb51983aa04705b3d3151e8fc6#r3364849040) to target `es2020` either way not to cause any confusion.

typescript-declarations/carp-kotlin/index.ts
Whathecode: This is probably done to satisfy the unit test:
```
expect( infinite.inWholeMilliseconds ).equals( -1 )
expect( infinite.inWholeMicroseconds ).equals( -1 )
```

Before, `INFINITY` was defined in JS compiled sources as:
```
    this.xf_1 = durationOfMillis(new Long(-1, 1073741823));
```

Now, because `bigint` is enabled, it is:
```
    this.ih_1 = durationOfMillis(4611686018427387903n);
```

So, I think the unit test was simply wrong here. The conversion from `Long` to JS `number` never could be successful, so if that overflowed (most likely the case for their chosen value for `INFINITY`), what the `inWholeMilliseconds` and `inWholeMicroseconds` value for that would be is arbitrary, i.e., not something to test for, or that needs to be maintained going forward.

Instead, the test should be updated, maybe by using `toString()`.

publish-npm-packages/build.gradle
Whathecode: Updated the commit message and squashed the various fixed into one commit.

typescript-declarations/carp-kotlin/index.ts
Whathecode: Fixed in a preceding commit: a97ad42a393928711c1ac8bcacc41eeee13ef358

typescript-declarations/carp-kotlinx-datetime/index.ts
```diff
@@ -34,8 +34,8 @@ declare module "@cachet/Kotlin-DateTime-library-kotlinx-datetime"
 
 
 // Implement base interfaces in internal types.
-extend.$_$.System.prototype.now = function(): kotlinx.datetime.Instant { return this.d16(); };
-extend.$_$.Instant_0.prototype.toEpochMilliseconds = function(): number { return this.h1m(); };
+extend.$_$.System.prototype.now = function(): kotlinx.datetime.Instant { return this.f15(); };
+extend.$_$.Instant_0.prototype.toEpochMilliseconds = function(): number { return Number( this.j1l() ); };
```
Whathecode: This was probably wrong before, as it returned a `Long`. So, this should be a `bigint` now and then the `Number()` constructor isn't needed.

The single unit test only tested `expect( now.toEpochMilliseconds() ).not.undefined` so it is probably a bug that didn't get caught before.

Whathecode: Fixed in a preceding commit now: 56b3905b56e0c496008477e788eed1003f271ff0

publish-npm-packages/src/forced-exports/kotlin-kotlin-stdlib
Whathecode: Done in the same commit which upgrades to 2.2.20. In addition, I added another more general cleanup commit reducing the need for `@Suppress` by using features which were introduced in Kotlin 1.8 and 2.0: 979d66849c580e114071c09c669956208e1ebec7

build.gradle
Whathecode: Fixed: e3a3d945999bdf54ed36a9800301c802a22bf4c1

In terms of whether we can abandon Big.js: similar to how we still need BigDecimal on JVM, Big.js is still relevant on JS.

rpc/src/main/kotlin/dk/cachet/carp/rpc/GenerateExampleRequests.kt
```diff
@@ -70,9 +70,9 @@ fun generateExampleRequests( serviceInfo: ApplicationServiceInfo ): List<Example
         val responseJson = json.encodeToString( checkNotNull( exampleJson[ "response" ] ) )
 
         ExampleRequest(
-            request,
+            checkNotNull( request.javaMethod ),
             ExampleRequest.JsonExample( example.request::class.java, requestObjectJson ),
-            ExampleRequest.JsonExample( request.returnType, responseJson )
+            ExampleRequest.JsonExample( request.returnType.javaClass, responseJson )
```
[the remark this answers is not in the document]
Whathecode: Works fine.

typescript-declarations/carp-kotlin/index.ts
```diff
@@ -6,23 +6,34 @@ import extend from "@cachet/kotlin-kotlin-stdlib"
...
         constructor( first: K, second: V ) {
             let kotlinPair = new extend.$_$.Pair( first, second );
-            kotlinPair.first = kotlinPair.me_1;
-            kotlinPair.second = kotlinPair.ne_1;
+            kotlinPair.first = kotlinPair.od_1;
+            kotlinPair.second = kotlinPair.pd_1;
             return kotlinPair;
         }
         get first(): K { return this.first; }
         get second(): V { return this.second; }
```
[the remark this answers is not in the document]
Whathecode: Apparently true. Will fix it as part of later commit in which these declarations get cleaned up for collection exports using KtList, etc.

yuanchen233 (approved):
:shipit:

PR #542, opened 2026-06-08
Replace custom TypeScript collection exports with Kotlin generated ones

Whathecode:
Since Kotlin 2.0.0, collection types are exported to TypeScript by default. Re-export these instead of custom exports relying on mangled JS names.

Progress:
- [x] List
- [x] Set
- [x] Map (+ Pair?)

### DIFF
--- a/typescript-declarations/carp-kotlin/kotlin-kotlin-stdlib.d.ts
+++ b/typescript-declarations/carp-kotlin/kotlin-kotlin-stdlib.d.ts
@@ -15,38 +15,19 @@ declare module "@cachet/kotlin-kotlin-stdlib"
 
         interface Collection<T>
         {
-            // contains
-            e1( value: T ): boolean
-
-            // size
-            o(): number
-
             toArray(): Array<T>
         }
 
-        interface List<T> extends Collection<T> {}
+        interface List<T> extends Collection<T>, kotlin.collections.KtList<T> {}
         interface EmptyList<T> extends List<T> {}
         interface AbstractMutableList<T> extends List<T> {}
-        function listOf_0<T>( elements: T[] ): List<T>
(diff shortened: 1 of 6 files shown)


typescript-declarations/tests/carp-kotlin-test.ts
```diff
@@ -88,12 +92,45 @@ describe( "kotlin", () => {
             expect( three.size() ).equals( 3 )
         } )
 
-        it( "empty list succeeds", () => {
```
Whathecode, line 91: I actually think this should be re-introduced. 🤔 But maybe as a KtList test instead.

typescript-declarations/carp-kotlin/kotlin-kotlin-stdlib.d.ts
```diff
@@ -15,24 +15,16 @@ declare module "@cachet/kotlin-kotlin-stdlib"
...
             toArray(): Array<T>
         }
 
-        interface List<T> extends Collection<T> {}
+        interface List<T> extends Collection<T>, kotlin.collections.KtList<T> {}
         interface EmptyList<T> extends List<T> {}
         interface AbstractMutableList<T> extends List<T> {}
-        function listOf_0<T>( elements: T[] ): List<T>
 
         interface Set<T> extends Collection<T> {}
```
Whathecode: How come an extends of KtSet, equivalent to why it seems to have been need for KtList, is not present here?

Whathecode: I still don't understand. I just added the extends for consistency.

yuanchen233 (approved):
:shipit:

PR #543, opened 2026-06-11
Update kotlin 2.4.0

yuanchen233:
Update Kotlin to 2.4.0, and also other dependencies.
Upgrade Gradle wrapper 9.5.0.
`Instant` and `Clock` are part of `kotlin.time`.
Mangled name needs to be updated.


carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/users/AccountIdentity.kt
```diff
@@ -10,11 +10,8 @@ import kotlin.js.JsName
  * Identifies an account.
  */
 @Polymorphic
-@Serializable
 @JsExport
-// HACK: defined as abstract class so that `JsExport` works: https://youtrack.jetbrains.com/issue/KT-51292
-@Suppress( "UnnecessaryAbstractClass", "UtilityClassWithPublicConstructor" )
-abstract class AccountIdentity
+interface AccountIdentity
```
yuanchen233, line 14: Kotlin 2.4.0, interface export are improved, and _Non-sealed interfaces are polymorphically serializable by default._, The hack here is therefore removed.

typescript-declarations/carp-kotlinx-serialization/index.ts
```diff
@@ -7,7 +7,11 @@ import extendJson from "@cachet/kotlinx-serialization-kotlinx-serialization-json
 // Facade with better method names and type conversions for internal types.
 export namespace kotlinx.serialization
 {
-    export function getSerializer( type: any ) { return type.Companion.c1o() }
+    export function getSerializer( type: any )
+    {
+        const serializer = type.$metadata$.associatedObjects[ 0 ]()
+        return serializer.m1o?.() ?? serializer
+    }
```
yuanchen233, line 22: Kotlin [unified companion object access across JS module systems﻿](https://kotlinlang.org/docs/whatsnew23.html#unified-companion-object-access-across-js-module-systems), the generated serializers are not exposed via companion anymore.

build.gradle
```diff
@@ -94,7 +94,7 @@ configure( subprojects - devOpsModules ) {
             }
         }
         js {
-            moduleName = project.name.replaceAll("\\.", "-") + "-generated"
```
yuanchen233, line 97: `moduleName` is removed

carp.common/src/commonMain/kotlin/dk/cachet/carp/common/application/users/AccountIdentity.kt
yuanchen233: [Kotlin/JS 2.4.0 Release notes](https://kotlinlang.org/docs/whatsnew24.html#kotlin-js), there are more supports on `JsExport` , e.g. `value class` which we are using for `Duration`, however, `Duration` itself in not exportable currently.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamBatch.kt
```diff
@@ -1,4 +1,4 @@
-@file:Suppress( "NON_EXPORTABLE_TYPE" )
+@file:Suppress( "EXPOSED_NOT_EXPORTED_SUPER_INTERFACE_WARNING", "NON_EXPORTABLE_TYPE" )
```
yuanchen233: [Exposing non-exportable API from interfaces](https://youtrack.jetbrains.com/issue/KT-83009) became stricter.

build.gradle
yuanchen233, line 54: Maybe look into this one, a bit old and uses `node` 18, or define another project `node` level.

yuanchen233: It defaults to _not_ download, so it's fine? I had a miss configuration so looked at this.

typescript-declarations/carp-kotlinx-serialization/index.ts
Whathecode: Apparently this version also introduces the ability to export suspend functions. And, exporting interfaces I think was supported before. That maybe means we should try to export our application service interfaces as part of upgrading to 2.3.10?

Of course, we can also postpone that and do it in a later commit.

Whathecode: As is, this is not very robust. It assumes the serializer is always the first metadata object. Is that the case? And, will that remain so? We could probably filter by name.

However, I wonder whether there isn't an easier way to retrieve it. I'm trying to generate some JS code from Kotlin to see whether I can find anything. :)

yuanchen233: I thought about, If each class has an companion object, which is the serializer of that class, then it become accessible due to this update. That will be a lot of companion objects added tho.

Or have an a new class that have a 'static function' (companion object) which can get the serializer of a given class.

Whathecode: > As is, this is not very robust. It assumes the serializer is always the first metadata object. Is that the case?

**Definitely not.** What `type.$metadata$.associatedObjects[ 0 ]()` actually seems to do is _either_ get the `Companion` object, _or_, get the `serializer`. If it is a `Companion` object, the serializer can be retrieved using `m1n` (just like `c10` was used before).

For some types (e.g., `ExpectedParticipantData`), there is no `Companion` property anymore, and only an associated `$serializer_getInstance_<xx>()` object. For others (e.g., `StudyProtocolSnapshot`), there is a `Companion` property, and `m1n` could just be called directly on that.

Whathecode:
```
    export function getSerializer( type: any )
    {
        type AssociatedObject = { name: string; (): any; };
        const associatedObjects = Object.values( type.$metadata$.associatedObjects ) as Array<AssociatedObject>
        
        const companion = type.Companion !== undefined
            ? type.Companion
            : associatedObjects.find( (o) => o.name.startsWith( "Companion" ) )?.()
        const serializer = companion !== undefined
            ? companion.m1n()
            : associatedObjects.find( (o) => o.name.startsWith( "$serializer" ) )?.()
        return serializer
    }
```

Better. ;) You can squash the fix into the first commit. I'll continue reviewing the rest later.

carp.data.core/src/commonMain/kotlin/dk/cachet/carp/data/application/DataStreamBatch.kt
Whathecode: This can be placed on `DataStreamBatch` to reduce the scope. Unlike "NON_EXPORTABLE_TYPE", it doesn't have to be added in multiple places.

Whathecode (commented):
The first three commits have been reviewed/approved (up to kotlin 2.4.0 upgrade). 👍 You could pull them into master on a separate PR already if you want.

I'll try to look at the rest soon. At home, when I tried to pull in the full branch, I was getting various IDE errors. Maybe due to Gradle updates and a need to clear caches.

yuanchen233:
> I'll try to look at the rest soon. At home, when I tried to pull in the full branch, I was getting various IDE errors. Maybe due to Gradle updates and a need to clear caches.

Yes, I might had that issue as well, had to manually clear the caches and stop daemons. Also I needed to upgrade Intellij IDEA to 26.01 so the IDE-plugin is compatible. This is how I noticed we have are using the old node plug in, I thought that was the problem.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.14-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.4-bin.zip
```
Whathecode: Just upgrade to [the latest supported version](https://kotlinlang.org/docs/whatsnew24.html#gradle) while at it (9.5.0). I did so locally and that works just fine.

Also, since it seems like this was done since a deprecation warning was triggered in the Kotlin 2.4.0 update, I think it is better to squash this in that commit and make mention of the upgrade in the description. That also provides a reason on why to update to 9.5.0 and not 9.6.0 yet (9.5.0 is documented as latest supported version on Kotlin 2.4.0).

I do still get tons of "unresolved references" even though everything builds fine, but I also got that with 8.14.4, so that should be unrelated. I invalidated IntelliJ caches, and cleared my Gradle .cache folder, but still no cigar. What did you end up doing to get rid of these?

build.gradle
Whathecode: Can't follow here. But generally, try to update everything, yes. :)

build.gradle
```diff
@@ -33,7 +33,7 @@ buildscript {
             detektPlugin:'1.23.8',
             detektVerifyImplementation:'1.2.6',
             nexusPublishPlugin:'2.0.0',
-            apacheCommons:'2.19.0'
+            apacheCommons:'2.22.0'
```
Whathecode, line 36: I think these dependency updates can be squashed together with the Kotlin 2.4.0 update. Nothing special happening here.

You could even add the serialization update to the Kotlin 2.3.21 upgrade (serialization 1.11.0 was released for Kotlin 2.3.20). However, as I suspect you may need to fix up more intermediate mangled names then, probably not worth it.

Whathecode (changes requested):
Looks good! I would just squash some commits and upgrade to Grade 9.5.0.

And, if you recall how you fixed your IDE, any input would be welcome. ;) I give up for today.

build.gradle
yuanchen233: It's the [plugin we are using ](https://github.com/node-gradle/gradle-node-plugin) haven't been updated for  two years, but looks like it's active again.

gradle/wrapper/gradle-wrapper.properties
yuanchen233: I was seeing that on IntelliJ IDEA 2025.3. Upgrading to 2026.1.3 via JetBrains Toolbox fixed it for me.
I’m not sure what side effects the upgrade might have had, though. For context, I had cleared the Gradle caches before upgrading, and I didn’t run anything again until the IntelliJ upgrade had fully finished. Could related to Plugins caches, one of the plugin was not compatible for me, that's why I did the upgrade.

Whathecode: I'm on 2025.1, and my update button in JetBrains Toolbox  is grayed out. I figured maybe because it was only another EAP release that was available, but I can see now that's not true. So, something must be wrong with my editor.

Whathecode: > Upgrading to 2026.1.3 via JetBrains Toolbox fixed it for me.

Same for me. Good reason to update the README to reference IntelliJ 2026 as part of the Gradle commit. :)

Whathecode (approved):

Issue #521, opened 2025-12-19
Refactor creation endpoints


yuanchen233:
Align all creation endpoints and let the caller pass and ID for the created object. As discussed [here](https://github.com/carp-dk/carp.core-kotlin/pull/519#discussion_r2472239365), endpoints like 'createStudy' should let the caller pass and ID.

PR #547, opened 2026-07-19
Fix: `stopParticipantGroup` verifies study membership using the wrong ID

### LINKED ISSUE #548: Deterministic test UUIDs collide across services
Separate `TestUUIDFactory` instances emit identical ID sequences, so in `RecruitmentServiceHostTest.createSUT()` study IDs coincidentally always equal recruitment-generated IDs. An example bug check #547 .

Possibly add a parameter to `TestUUIDFactory` to differentiate services: a `serviceHex` like the `idHex`.

yuanchen233:
`stopParticipantGroup` checked `require( recruitment.id == studyId )`, but `recruitment.id` is the aggregate's own random ID — so the check always threw `IllegalArgumentException`, even for the correct study. 

Existing tests pass unchanged, they couldn't catch this because deterministic test UUIDs made `recruitment.id` coincidentally equal the study ID. This is tracked #548.

---
Update: Also bumped snapshot versions, and updated test infrastructure so this bug can be caught.

### DIFF
--- a/carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHostTest.kt
+++ b/carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHostTest.kt
@@ -22,13 +22,14 @@ class RecruitmentServiceHostTest : RecruitmentServiceTest
         fun createSUT(): RecruitmentServiceTest.SUT
         {
             val eventBus = SingleThreadedEventBus()
+            val uuidFactory = TestUUIDFactory()
 
             // Create dependent study service.
             val studyRepo = InMemoryStudyRepository()
             val studyService = StudyServiceHost(
                 studyRepo,
                 eventBus.createApplicationServiceAdapter( StudyService::class ),
-                TestUUIDFactory(),
+                uuidFactory,
                 TestClock
             )
 
@@ -44,7 +45,7 @@ class RecruitmentServiceHostTest : RecruitmentServiceTest
                 InMemoryParticipantRepository(),
                 deploymentService,
                 eventBus.createApplicationServiceAdapter( RecruitmentService::class ),
-                TestUUIDFactory(),
+                uuidFactory,
                 TestClock
             )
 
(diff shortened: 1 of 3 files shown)


yuanchen233:
In `RecruitmentServiceHostTest`, I kept the recruitment service on the default `TestUUIDFactory()` and only changed the study service's.

In the backwards-compatibility tests, `participant.id` and `group.id` are generated by the recruitment service's UUID factory, so unless it stays the default they won't match the recorded IDs and the replay fails, which would force me to regenerate the test request Json files.

So I only changed the study service. That's enough to make `recruitment.studyId` differ from `recruitment.id` and expose the bug.

carp.common.test/src/commonMain/kotlin/dk/cachet/carp/common/test/infrastructure/TestUUIDFactory.kt
```diff
@@ -5,15 +5,17 @@ import dk.cachet.carp.common.application.UUIDFactory
 
 
 /**
- * Creates a UUIDs counting up from 1.
+ * Creates UUIDs counting up from 1.
+ * [service] differentiates IDs generated for different services, defaults to 0.
  */
-class TestUUIDFactory : UUIDFactory
+class TestUUIDFactory( service: Int = 0 ) : UUIDFactory
```
Whathecode: I wouldn't reference "service" here. Instead maybe just "prefix"?

carp.studies.core/src/commonTest/kotlin/dk/cachet/carp/studies/application/RecruitmentServiceHostTest.kt
```diff
@@ -28,7 +28,7 @@ class RecruitmentServiceHostTest : RecruitmentServiceTest
             val studyService = StudyServiceHost(
                 studyRepo,
                 eventBus.createApplicationServiceAdapter( StudyService::class ),
-                TestUUIDFactory(),
+                TestUUIDFactory( 1 ),
```
Whathecode: Aren't there more places this is needed? And, maybe rather than hardcoding a "1", it could be text from which a hash is created? E.g. "study" here. Another option (if trivial) is to pass the same uuid generator to all services.

yuanchen233: > Another option (if trivial) is to pass the same uuid generator to all services.

This works, and cleaner, and it's in line with #521.

This is the only host test has two ID-generating services, so it's not needed elsewhere.

Whathecode (approved):

